PR #7, opened 2016-03-03
Add default_prefix support for Windows

bheklilr:
I modified winexe.py and main.nsi.tmpl slightly to allow the construct.yaml file to specify the `default_prefix` key to be used to customize the installation location.  I believe it should have the same behavior as in shar.py for unix based systems, but I don't have access to one right now to confirm.  This change calls `os.path.expandvars` on the value of `default_prefix` instead of relying on the NSIS `$LocalAppData` variable to expand the path.  This should also make the feature as powerful as for unix installers since variables other than `$HOME`/`%LOCALAPPDATA%` could be used in the prefix.

An example usage is:

```yaml
name: test
version: "1.0.0.0"

channels:
  - http://repo.continuum.io/pkgs/free/

specs:
  - python

default_prefix: "%LOCALAPPDATA%\Company ABC\test"
```

This would build a minimal installer which when installed with "Just for me" on Windows would point at `C:\Users\username\AppData\Local\Company ABC\test`.  If you see any problems with it let me know and I'll get them fixed.

### DIFF
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -28,6 +28,7 @@
 !define PY_VER __PY_VER__
 !define PYVERSION_JUSTDIGITS __PYVERSION_JUSTDIGITS__
 !define PYVERSION __PYVERSION__
+!define DEFAULT_PREFIX __DEFAULT_PREFIX__
 !define PRODUCT_NAME "${NAME} ${VERSION} (${ARCH})"
 !define UNINSTALL_NAME "Python ${PYVERSION} (${NAME} ${VERSION} ${ARCH})"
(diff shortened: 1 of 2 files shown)


ilanschnell:
Thanks.  This looks like a useful addition.

bheklilr:
@ilanschnell I just noticed that there is a bug in this implementation, when using `os.path.expandvars` it will likely expand to include the username of person building the package, not the person installing it.  I'm working on a fix now and will submit a new pull request when I have that implemented.  My apologies for introducing a bug.

I'm going to guess that what needs to happen is that the variable expansion needs to happen entirely within the `nsi` file rather than in Python, I'll figure out how to make that happen.

PR #20, opened 2016-04-12
rework PATH setting, add path tests

msarahan:
- Adds Library/bin to paths
- sets ANACONDA_PATH with PATH modifications, then sets PATH to include %ANACONDA_PATH% - this helps avoid length limitations
- moves registry get/set to more isolated functions


msarahan:
@mingwandroid does this PR add anything useful, or should I just close it?

PR #22, opened 2016-04-29
Don't install menus if menuinst raises an OSError on import.

insertinterestingnamehere:
This can happen when, on Windows, menuinst can't find all the system
directories it needs in order to install the menus.
When running in an environment unable to displa a GUI interface,
an uncaught exception like this can result in a frozen install with
no visible debug info available on the command line.


ilanschnell:
Thanks

PR #25, opened 2016-06-01
force all users if installing as SYSTEM

msarahan:
(port of similar change to Anaconda installer from https://github.com/ContinuumIO/anaconda/pull/511)

CC @mingwandroid @ilanschnell


ilanschnell:
Thanks

PR #26, opened 2016-06-13
Use get_index, handle conda 4.1 channels

mcg1969:
I'm using a single call to `api.get_index` in order to get all the repodata. This has one consequence that you may wish me to reverse: it includes any channels explicitly listed in the `packages` list in the channel index that is used to resolve `specs`.


mcg1969:
@ilanschnell hold on, I forgot this code uses a function `url2dist` that I haven't pushed anywhere. This won't work until it's added

ilanschnell:
Thanks.

PR #43, opened 2016-09-08
Rdonnelly/allow noarch

mingwandroid:
Hi @ilanschnell, let me know if this is OK or if it needs more work. I think noarch is going to become increasingly important as we start to work on js and java packages.


ilanschnell:
Ok, that should be fine.  However, it will only for the generic `noarch` packages.  Not for the so-called "Python noarch" ones.

mingwandroid:
Thanks. I guess Python noarch ones have their own complications due to the (pointless) folder layout differences?

ilanschnell:
Yes, Python noarch packages use a (generated) post link script to make the mapping of the site-packages work on all Platforms and for all Python versions.

joelhullcio:
@ilanschnell @mingwandroid when will this change be released so I can use it?

mingwandroid:
@joelhullcio, was this rolled out yet?

PR #44, opened 2016-09-14
Add -s flag to shell installer to run without executing user-defined scripts

dsludwig:
If you don't want to execute the package-defined scripts, the `-s` flag installs all the packages by simply extracting and linking them into place.


ilanschnell:
Implemented similar approach which is going to be part of the 1.3.4 release.

Issue #46, opened 2016-09-29
resulting Installer needs an --update-existing option


srossross:
We need a robust way to update existing installs with a new installer file.

ilanschnell:
This feature already exists.  This is what the `-f` option of the resulting installer is for.

csoja:
I don't think -f is the right approach for an upgrade - even if it does what is intended, I think a -f might make someone think something is wrong that they need to force the install.  What about adding a --upgrade option?  @srossross are there other abilities needed for an upgrade feature that the -f feature does not provide?

dsludwig:
As far as I can tell, `-f` does not support unlinking the old packages that are installed in a prefix. Ideally this feature would act more like a `conda upgrade` where the installer provides a `file://` channel to pull from - allowing a compatible set of packages to be pulled from the new installer, along with all the correct upgrade semantics.

ilanschnell:
It is not exactly clear to me what the update (or upgrade) option should do.  Obviously, it is the responsibility of the creator of an installer that all packages contained in the installer are compatible with each other.  Therefore, updating an existing installation is basically replacing an old installation with the new one.  This is different from the exsiting `-f` option, which merely ignores an existing install prefix and "force" installs all packages on top of what is already there.
I've implemented the new `-u` option in https://github.com/conda/constructor/pull/48

csoja:
what is the next step here? @dsludwig  can you test the open PR and see if it meets your needs?

ilanschnell:
Exactly, this is why a created the branch (with PR), so that @dsludwig can try it out.

ilanschnell:
I've update #48 to not use `rm -rf` on the entire install prefix anymore.  Instead, only the excess conda-meta entries are removed upon update, which is also more consistent with what the Anaconda (and Miniconda) installers are currently doing.

csoja:
@dsludwig can you test the latest?

dsludwig:
Thanks Ilan, I will try this out.

dsludwig:
@ilanschnell, I got this error when testing on OS X (using the Maxiconda example):

```
installing: python.app-1.2-py35_4 ...
ln: ./lib: File exists
Error: post-link failed for: python.app-1.2-py35_4
```

ilanschnell:
@dsludwig I was able to reproduce the problem, and also able to fix it (by adding a missing to call to the pre-unlink script).  The Maxiconda example is working now.

dsludwig:
[...]

PR #48, opened 2016-10-21
add -u option to shell installer

ilanschnell:
See also https://github.com/conda/constructor/issues/46

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -173,7 +173,8 @@ case "$PREFIX" in
 esac
 
 if [[ ($FORCE == 0) && (-e $PREFIX) ]]; then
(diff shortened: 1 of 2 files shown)


constructor/header.sh
```diff
@@ -173,7 +173,8 @@ case "$PREFIX" in
 esac
 
 if [[ ($FORCE == 0) && (-e $PREFIX) ]]; then
-    echo "ERROR: File or directory already exists: $PREFIX" >&2
+    echo "ERROR: File or directory already exists: $PREFIX
+If you want to update and existing installation, use the -u option." >&2
```
dsludwig: `and` should be `an` here

ilanschnell: Thanks, just fixed the typo.

PR #53, opened 2016-12-14
add note on Windows channels /free and /msys2

electronwill:
add note on Windows channels /free and /msys2, based on discussion today in Flowdock channel "Anaconda Distribution" with @trentoliphant and @mingwandroid .

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,13 @@ in <a href="./examples/maxiconda">examples/maxiconda</a>.
 Notes:
 ------
 
-  * Constructor does not work with `noarch`-Python packages, all conda
+  * Constructor does not work with `noarch`-Python packages. All conda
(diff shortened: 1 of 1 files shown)


electronwill:
Good point. Would this be better?

"It is best to add `/msys2` as `http://repo.continuum.io/pkgs/msys2`. Be sure no firewall is set to block this URL."

mingwandroid:
Well I don't think we mention firewalls for repo.continuum.io in general, so we shouldn't here either.. that anaconda.org also has the packages is just because the are rsynced from there to repo.

Though it is am alternative route to the same packages ...

electronwill:
Thank you-- that makes sense!

ilanschnell:
I've manually made the changes to master, because there was a conflict.  Thanks @electronwill

PR #58, opened 2017-01-12
Unlink files prior to writing with a new prefix

### LINKED ISSUE #57: Bus Error (core dumped) with anaconda.org/anaconda and anaconda.org/conda-forge
Hi constructor devs,

I am getting a bus error when using `conda.anaconda.org/anaconda` and `conda.anaconda.org/conda-forge` as my channels in a construct.yaml.  I have pared down the construct.yaml as far as it can go

```yaml
name: python
version: 3.5 
channels:
    - https://conda.anaconda.org/conda-forge
    - https://conda.anaconda.org/anaconda
specs:
    - python 3.5*
```

Running the constructor that is produced by this construct.yaml produces the following error:

```
$ bash ./python-3.5-Linux-x86_64.sh -b -p /tmp/foo
PREFIX=/tmp/foo
installing: python-3.5.2-4 ...
Python 3.5.2
./python-3.5-Linux-x86_64.sh: line 161: 20122 Bus error               (core dumped) $PYTHON -E -s $PREFIX/pkgs/.install.py $INST_OPT
```

I tried to do a diff of the installer scripts produced when I remove conda-forge as a channel, but, as expected, that diff is pretty much inscrutable and filled with binary diffs :disappointed: 

cc @msarahan 

[edit] Running constructor without conda-forge as a channel does work

jjhelmus:
When performing prefix replacement during an install it may be possible that
the file being updated is memory mapped, for example binary files associated
with the Python package.  Performing an in-place update on a memory mapped file
can result in a bus error.  To prevent this the file is unlinked prior to
writing the file with the new prefix.

closes #57

### DIFF
--- a/constructor/install.py
+++ b/constructor/install.py
@@ -188,6 +188,7 @@ def update_prefix(path, new_prefix, placeholder, mode):
     if new_data == data:
         return
     st = os.lstat(path)
+    os.unlink(path)     # unlink in case the file is memory mapped
     with exp_backoff_fn(open, path, 'wb') as fo:
         fo.write(new_data)
     os.chmod(path, stat.S_IMODE(st.st_mode))


ilanschnell:
Thanks!

Issue #49, opened 2016-11-04
Uninstall information in HKCU when running silent installer as admin


wulmer:
Hi, I tried running an installer (created by constructor) with the `/S` option from an admin console. While the installation went well, I noticed that the uninstaller registry entry is located under `HKCU\Software\Microsoft\Windows\CurrentVersion\Uninstall`. 

When I run the installer without `/S`, the uninstall information is written under `HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall`.

In my opinion, the location of the uninstaller entry should not change depending on the mode, so I consider this a bug.
I have checked the NSI source code and I think it has to do with the automatic detection of the install mode. While `$InstMode` might be correctly guessed as `${ALL_USERS}` (didn't verify this), the uninstaller entry is written to `WriteRegStr SHCTX "${UNINSTREG}" ...`. I am guessing that using `SHCTX` here might not always reflect the choice taken by the automatic `$InstMode` guessing mechanism.

wulmer:
I have pinned down the problem to the `InstModeChanged` function in `main.nsi.tmpl`. In non-silent mode, the shell var context is set in [line 235](https://github.com/conda/constructor/blob/master/constructor/nsis/main.nsi.tmpl#L235) to the correct installation mode, but in silent mode, the call to `SetShellVarContext` never happens.

I think the ` ${IfNot} ${Silent}` block which is meant for protecting the `/D` command line option should not include the calls to `SetShellVarContext` as they don't have to be protected in silent mode.

I will prepare a PR for this change.

speleo3:
@wulmer can this be closed? #60 is merged

wulmer:
Sorry, yes, of course. Have been away for a few days...

PR #60, opened 2017-01-18
Set correct shell context even in silent mode

wulmer:
This PR fixes issue #49 by calling `SetShellVarContext` even in silent mode.


wulmer:
@msarahan Any chance this PR gets merged?

msarahan:
Thanks @wulmer - how did you find this?  What issue was triggered by it?

wulmer:
The problem was that when I tried to prepare an installer for an admin install under SCCM, I noticed that the uninstall information was at the wrong reg key, see #49.
This would make the uninstall information useless as you would need admin rights to uninstall even if you only saw the uninstaller from a normal user's account.
Thanks for merging it.

PR #61, opened 2017-01-31
Add unicode line, update version comment

msarahan:
so that installers use unicode plugins from nsis 3.01


waterbug:
Works for me!  :)  (P.S.: a nice enhancement would be a 'constructor clean' command... but that's for a separate PR, obviously! ;)

ilanschnell:
Thanks for the PR @msarahan!

Adding a constructor clean command is easy @waterbug.

ilanschnell:
8175e5150646c394f43fa39fd9dc3f70c5074001

PR #66, opened 2017-03-02
Add optional readme.txt support

mbargull:
This adds the ability to show an included readme during installation, similar to how it's done with the license file.
The readme will only be included if the new key `readme_file` defined in `construct.yaml`.

### DIFF
--- a/CONSTRUCT.md
+++ b/CONSTRUCT.md
@@ -85,6 +85,12 @@ The filename of the installer being created.  A reasonable default filename
 will determined by the `name`, `version`, platform and installer type.
 
 
+`readme_file`:
+----------------
+Path to a readme file, which, if provided, is being displayed by the installer
+during the install process.
+
+
 `license_file`:
 ----------------
 Path to the license file being displayed by the installer during the install
--- a/constructor/construct.py
+++ b/constructor/construct.py
@@ -100,6 +100,11 @@ The type of the installer being created.  Possible values are "sh", "pkg",
 and "exe".  By default, the type is "sh" on Unix, and "exe" on Windows.
 '''),
 
+    ('readme_file',           False, str, '''
+Path to the readme file being displayed by the installer during the install
+process.
+    '''),
+
     ('license_file',           False, str, '''
 Path to the license file being displayed by the installer during the install
 process.
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -121,6 +121,12 @@ then
 
     echo "
 Welcome to __NAME__ __VERSION__"
+#if has_readme
+    echo "About __NAME__ __VERSION__:"
+    more <<EOF
+__README__
+EOF
+#endif
 #if has_license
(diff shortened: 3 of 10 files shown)


megies:
I think this would be a nice addition.

Especially since it would be nice to show our license and then also show a notice about Anaconda EULA.

msarahan:
@mbargull I'm in favor of merging this.  Do you have time to fix the merge conflict?  I currently do not, but I can add it to the list...

mingwandroid:
I'm in favour too but please add macOS pkg support too.

mbargull:
I rebased this on master.
But I don't think there is currently much for me to do regarding the macOS pkgs. Those diverge from the other systems in that a readme is already generated. Hence, before I could add support for macOS, it has to be decided whether an explicit `readme_file` would replace the generated one or not (which would remove the macOS-only functionality of listing the to-be-installed package list). What's your opinion on that @mingwandroid?

mingwandroid:
Maybe using the readme identified but having a flag as to whether to augment it or not would be best? Then we could add the augmentation to the other platforms too?

I should say, most of the active development is happening in this branch:

https://github.com/conda/constructor/tree/cross_conda_interface

.. so you may want your furture PRs to target that to avoid significant rebasing.

Your work on constructor is appreciated btw.

mbargull:
> Maybe using the readme identified but having a flag as to whether to augment it or not would be best? Then we could add the augmentation to the other platforms too?

A single flag might be too inflexible though. Picture a readme like:
```text
Very informative read me. I haz packages:
__GENERATED_PACKAGE_LIST__
Some important footer line.
```
So, possibly the package list cannot always be simply appended to the readme. Maybe the readme files should thus be some kind of template file which can be augmented with a package list (or even, additionally, with information from the `info` dictionary, e.g., `info['name']` and `info['version']`.). Especially given the current osx `readme.rtf` use case, the package list might even need to be formatted in a certain way.
Maybe we could leverage `jinja2`'s capabilities for this, e.g., turn the current `constructor/osx/readme_header.rtf` into a `jinja2` template with access to the `info` dictionary? This way the loop over `info["_dists"]` inside `constructor.osxpkg:write_readme` could be implemented in `jinja2` and replicate the current behavior. The same could then be done for the other platforms.

Here's what I propose: For *this* PR I can implement the simple replacement for osx, i.e., if `readme_file` is present, simply insert this plain text file (as is done for the other platforms) instead of the `readme.rtf`. The other aspect, i.e., augmenting readme files with package lists, will be discussed/implemented in another issues/PR. Are you good with this proposal?
>I should say, most of the active development is happening in this branch:
>
>https://github.com/conda/constructor/tree/cross_conda_interface
>
>.. so you may want your furture PRs to target that to avoid significant rebasing.

Yeah I noticed -- would've done so anyway ;).
>Your work on constructor is appreciated btw.

Thanks, you're welcome!

PR #68, opened 2017-03-02
Add ability to provide defaults for custom options

mbargull:
This adds to new keys (`add_to_path_default`, `register_python_default`) with which the installer's custom options can be given explicit default values in `construct.yaml`. Useful if one wants to create an installer which, by default, does not interfere with the system (i.e. both values set to `False`).

### DIFF
--- a/CONSTRUCT.md
+++ b/CONSTRUCT.md
@@ -132,3 +132,17 @@ The color of the default images (when not providing explicit image files)
 used on Windows.  Possible values are `red`, `green`, `blue`, `yellow`.
 The default is `blue`.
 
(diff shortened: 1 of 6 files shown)


astaric:
@ilanschnell is this change something that could be included in constructor?

I am currently trying to build win installer for a package that installs its own menu shortcuts and thus does not need to change default path. Personally, I would not mind setting the defaults and not showing the choice screen to users at all, but providing different defaults is a good start.

msarahan:
Thanks for this helpful change, @mbargull.  Merging.

PR #71, opened 2017-04-11
Remove bashisms from header.sh, using only POSIX, split tar and bunzip2

mingwandroid:
This makes the installation procedure compatible with systems that use
Busybox (integrated ash shell, tar from Busybox) and also for people who
use dash instead of bash.

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 #
 # NAME:  __NAME__
 # VER:   __VERSION__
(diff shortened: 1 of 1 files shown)


mingwandroid:
Please review this @ilanschnell.

mingwandroid:
I need to do some more work on this @ilanschnell.

ilanschnell:
Thanks Ray.  What is `RUNNING_SHELL` used for?  Are we sure that all the extra code related to that is not going to cause problems?

mingwandroid:
Some systems do not have `bash`, so this does a best guess effort to find an alternative. It sets a variable that is used instead of these two lines:
https://github.com/conda/constructor/pull/71/files#diff-3e6c886485792c86488973e8e7295183L228
https://github.com/conda/constructor/pull/71/files#diff-3e6c886485792c86488973e8e7295183L276

.. however I need to revise this part of macOS before it's ready for review, so I'll close the issue for now, thanks.

PR #72, opened 2017-04-18
Remove bashisms from header.sh, using only POSIX, split tar and bunzip2

mingwandroid:
This makes the installation procedure compatible with systems that use
Busybox (integrated ash shell, tar from Busybox) and also for people who
use dash instead of bash.

This is ready for review now @ilanschnell.

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -6,6 +6,7 @@
 # BYTES: @SIZE_BYTES@
 # LINES: @LINES@
 # MD5:   __MD5__
+
 #if osx
 unset DYLD_LIBRARY_PATH
 #else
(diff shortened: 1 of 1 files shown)


mingwandroid:
Ping @csoja, @ilanschnell, @kalefranz. I really want to see this merged before 4.4.

constructor/header.sh
```diff
@@ -1,89 +1,125 @@
...
+OPTS=$(getopt bfhp:su "$*" 2>/dev/null)
+if [ ! $? ]; then
+    echo "$USAGE"
+    exit 2
+fi
+
+eval set -- "$OPTS"
+
+while true; do
+    case "$1" in
```
kalefranz: This isn't actually trying to parse flags properly, is it?  I use `-bfp` all the time, and have so publicly in a lot of issues.  With the parsing now, wouldn't that have to change to `-b -f -p`?  Should we use `getopts` here instead?

kalefranz:
I've done a full review.  LGTM except for that one comment.

mingwandroid:
getopts isn't available on as many systems. This script needs to target the very lowest common denominator. I'll check tomorrow but I think you can still do -bfp just fine.

ilanschnell:
Thanks Ray for working on this.  However, I don't see why this is relevant for the Anaconda 4.4 release, as the Anaconda/Miniconda installers are created using a different bash header file, see: https://github.com/ContinuumIO/anaconda/blob/master/ibuild/build.sh
Since this other header file has been used for years, and as proven to be stable, I don't see any reason to change it for the Anaconda 4.4 release.

kalefranz:
Because we need to be able to make a busybox miniconda installer.

kalefranz:
Also, why would we want to use different code for constructor vs the other installers?

ilanschnell:
Because of stability concerns.  Also, the code is different for specific reasons, and enabling constructor to exactly the Anaconda/Miniconda installers we provide would be extra work.

constructor/header.sh
```diff
@@ -1,89 +1,125 @@
...
-THIS_DIR=$(cd $(dirname $0); pwd)
-THIS_FILE=$(basename $0)
+# Determine RUNNING_SHELL; if SHELL is non-zero use that.
+if [ -n "$SHELL" ]; then
+    RUNNING_SHELL="$SHELL"
+else
+    if [ "$(uname)" = "Darwin" ]; then
+        RUNNING_SHELL=/bin/bash
+    else
+        RUNNING_SHELL=$(ps -p $$ -oargs=)
```
pzwang: How robust is part of the branch?  AFAICT most shells *do* define `$SHELL`, so how well is this logic understood/tested/etc.?
Also, why a special case for Darwin?  Is this to handle non-OSX Darwin, or...?

mingwandroid: It's less well tested. I'll undefine SHELL and make sure it's doing the right thing in some tests.

mingwandroid:
Adds more ways of finding `RUNNING_SHELL` and a helpful error message when everything fails.

constructor/header.sh
msarahan: @mingwandroid any result here?  Is this PR still worth pursuing?

mingwandroid:
This work is superceded by https://github.com/conda/constructor/tree/cross_conda_interface

Please base everything on that branch for now. I haven't merged it to master but perhaps I should?

msarahan:
I think it's best to merge it, yes, so that it doesn't become a gargantuan mega-PR.  I'm going to close this PR since it is superseded.

PR #74, opened 2017-04-30
Add basic jinja2 support

mbargull:
This adds support for `jinja2` template parsing in `construct.yaml`. The error handling is essentially copied from `conda-build`. Unlike `conda-build` it has only basic `jinja2` support, e.g., it is not possible to access environment variables etc.

Possible use case: Build an installer with Miniconda-esque version naming:
```yaml
{% set py_major_ver = "3" %}
{% set conda_ver = "4.3.16" %}

{% version = conda_ver %}
{% name = "InstallMyConda" + py_major_ver %}

name: {{ name }}
version: {{ conda_ver }}

specs:
  - python >={{ py_major_ver }}
  - conda =={{ conda_ver }}
...
```
Note: The `jinja2` processing is done *after* applying platform selectors.
Needs documentation (e.g., example and note above). If you find this PR viable, I could add some.

### DIFF
--- a/constructor/construct.py
+++ b/constructor/construct.py
@@ -170,27 +170,16 @@ def select_lines(data, namespace):
 
 
 # adapted from conda-build
-def render_jinja(data):
-    try:
-        import jinja2
-    except ImportError as ex:
-        raise exceptions.UnableToParseMissingJinja2(original=ex)
-    env = jinja2.Environment()
-    try:
-        template = env.from_string(data)
-        rendered = template.render()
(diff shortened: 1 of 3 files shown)


loriab:
+1 that this capability would be really helpful on multiple projects, if acceptable to the developers.

msarahan:
Merging, but I also agree that the jinja stuff here should remain simpler than it is in conda-build.  I don't really see the need for the equivalent of conda-build 3's variants here.

PR #75, opened 2017-05-02
Remove menus of all conda envs during uninstall

mbargull:
This makes the uninstaller remove not only the root environment's menus but also the menus from all other environments (if possible).
It tries to get the list of environments via the `conda.cli.python_api`.
An alternative implementation is provided, which uses `libconda.config.envs_dirs` and code copied from `conda.misc.list_prefixes` to get the list of environments. Since `conda.misc.list_prefixes` should be considered internal/subject to change, I feel more comfortable using `conda.cli.python_api` instead, which is thus the active implementation in this PR.

Use case:
An user creates new environments with packages containing `menuinst` menus. The user then uninstalls the whole `constructor`-created installation without explicitly removing those environments beforehand. Prior to this PR the menu entries created for packages outside the root environment would persist after the uninstallation.

### DIFF
--- a/constructor/nsis/_nsis.py
+++ b/constructor/nsis/_nsis.py
@@ -10,7 +10,7 @@
 import os
 import sys
 import traceback
-from os.path import isfile, join, exists
+from os.path import isdir, isfile, join, exists
 
 # Install an exception hook which pops up a message box.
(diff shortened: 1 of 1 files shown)


msarahan:
Thanks for this helpful change, @mbargull.  Merging.

PR #76, opened 2017-05-02
setup.py enhancements

mbargull:
This updates `setup.py` to use `entry_points`, `install_requires`, `package_data` from `setuptools`.
It allows the creation of egg- and wheel-distributions that include the platform specific files (`['header.sh']` or `['nsis/*', 'ttf/*']`, resp.) and removes the hard-coded post-processing copying.

### DIFF
--- a/bin/constructor
+++ b/bin/constructor
@@ -1,6 +1,0 @@
-#!/usr/bin/env python
-if __name__ == '__main__':
-    import sys
-    from constructor.main import main
-
-    sys.exit(main())
--- a/setup.py
+++ b/setup.py
@@ -7,11 +7,75 @@
 import re
 import sys
 from os.path import abspath, dirname, join
+from contextlib import contextmanager
 
-from setuptools import setup
+from setuptools import setup, Distribution
+from distutils.util import get_platform
+try:
+    from wheel.bdist_wheel import bdist_wheel
+except ImportError:
+    bdist_wheel = None
 
 
 SETUP_PY_DIR = dirname(abspath(__file__))
+
+
+def get_package_data(platform):
+    platform_package_data = {
+        "win32": ["nsis/*", "ttf/*"],
+        "unix": ["header.sh"],
+    }
+    if platform is None:
+        package_data = sum(platform_package_data.values(), [])
+    else:
+        platform = platform if platform in platform_package_data else "unix"
+        package_data = platform_package_data[platform]
+    return {"constructor": package_data}
+
+
+class PlatformSpecificDistribution(Distribution):
+    def run_commands(self):
+        for cmd in self.commands:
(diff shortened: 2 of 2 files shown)


msarahan:
@mbargull I merged this into what will become 1.7 at https://github.com/conda/constructor/commit/5995ea78ac926fa4ab0671a5625bb0fda6f39c17

Thanks for sorting this out!

msarahan:
@mingwandroid, let's close this PR whenever you merge in your other branch.

mbargull:
Note that this was written prior to osx pkg support. Thus the `platform_package_data = {...}` in `get_package_data(...)` will need an entry `"darwin": ["header.sh", "osx/*"],` or the like. @ilanschnell or @mingwandroid should now what to include here.
(BTW: It's nice to see `constructor`'s development picking up some pace lately! Thanks to everyone involved!)

mbargull:
If `"header.sh"` is also included for osx, maybe
```python
    if platform is None:
        package_data = sum(platform_package_data.values(), [])
```
might need to be changed to something like
```python
    if platform is None:
        package_data = list(set(chain.from_iterable(platform_package_data.values())))
```
? I'm not sure whether `setuptools` gracefully handles duplicates in `package_data`-lists.

ilanschnell:
Hmm, I would prefer if `setup.py` didn't use `setuptools`.  I don't see any good reason for doing so.

mbargull:
Hi @ilanschnell, which disadvantages to you see in using `setuptools`? (Asking out of genuine curiosity!)
All in all the advantages of using `setuptools` mainly aid to run/install `constructor` in more arbitrary, i.e. not only `conda`, environments. If it's preferred to only run `constructor` from `conda` environments, I could agree that there might be no good reason for this. Though, I can't see -- given the low (maybe even non-existent) maintenance needed -- why `constructor` shouldn't support other common installation methods like `wheel`s and others.
Advantages enabled by `setuptools` (if not installed as a `conda` package):
- Being able to build `wheel`s (`egg`s).
- Entry points created for all platforms.
- Non-Python files installed into the correct locations if `package_data` is used; since `sys.prefix` is not necessary the root path of these locations, e.g., see https://docs.python.org/3/install/#alternate-installation.
- Being able to specify run requirements via `install_requires`.

msarahan:
I have personally had trouble developing constructor.  I support
@mbargull's PR here because it makes it easier to develop constructor using
a "normal" workflow.  "Normal" meaning I don't need to install the
constructor conda package in order to hack on constructor.  Without that, I
don't have the entry point, which means I can't test constructor.

Thank you @mbargull for sorting this out.

mbargull:
Merged into master with #124.

Issue #77, opened 2017-05-17
Constructor and "installers" can not handle packages that track features.


hsparra:
The current version of constructor does not appear to work if one of the packages in the specs tracks a feature. As an example, in one situation the error message is
```
Error: Dependency missing in current linux-64 channels:
- mycerts ==1.0.4 ->@my
```
The construct.yaml file contains the channels where "mycerts" is located.

looooo:
I think I ran into the same issue:
```
channels:
  - https://conda.anaconda.org/conda-forge    # hosts freetype2.7
```
```
Error:  Dependency missing in current linux-64 channels: 
  - occt 7.1.0* -> freetype 2.7.*

```

mcg1969:
Sorry for the form letter here—I'm doing some issue cleanup right now and closing out some stale issues. Constructor 3.x was a significant rewrite, and so many of the issues filed before its release would need to be re-tested / re-verified. If you have good reason to believe this still needs to be addressed in 3.x, please feel free to re-open or file a new issue with additional information on how to reproduce the issue on the latest versions of constructor.

Issue #54, opened 2017-01-06
channel priority for constructor


csoja:
Feature request to add channel priority for constructor

joelhullcio:
Can we get this feature officially added to Constructor soon?  It would be very helpful.

darkdreamingdan:
Can someone explain this bug in a bit more detail?  If I add a list of channels into constructor, and a list of specs, how are channels currently prioritised? Randomly?

jjhelmus:
Currently all channels listed in the `construct.yaml` file are given equal priority.  The solver examines all of the packages in all channels with no preference for the source.  Note that the solver used comes from `libconda` not `conda` and may give different results that those produced from the equivalent `conda create --no-channel-priority` command.

I also believe there may be an issue with the current behavior where packages with identical names coming from different channels may clobber each other when constructor creates an index of all available packages.

jakirkham:
Is there any reason to not just use `conda` during construction?

jjhelmus:
> Is there any reason to not just use conda during construction?

conda (the python package) is only accessible from the root environment.  If constructor used `conda` to solve the package spec, then constructor would need to be installed in the root environment.  Using `libconda` allowed constructor to be installed in any environment.

jakirkham:
Sure. Am just concerned that using such an old version of `conda` in the form of `libconda` is incurring some significant penalties. Plus its concerning that two different packages are trying to be solvers. Are there any thoughts or plans as to how to mitigate these two issues?

jjhelmus:
> Are there any thoughts or plans as to how to mitigate these two issues?

Yes, see #79

mingwandroid:
New constructors are ready and version 2 has been made, thanks @nehaljwani 

Apart from needing to install constructor in your root env there are no other downsides AFAICT and if that's something you do not want then you can always remove it afterwards.

mingwandroid:
My opinion on libconda is that it's fine for it to make a comeback, but if it does then `conda` needs to depend upon it, cc @kalefranz.

hajapy:
I am not sure why this issue was closed as it seems constructor does not respect channel priority the same way conda does yet. https://github.com/conda/constructor/pull/79 is still open and in my opinion is greatly needed!

PR #79, opened 2017-05-19
Optional use of conda to solve spec in order to provide channel priority

jjhelmus:
Add an `--use-conda` argument which uses conda rather than libconda to fetch the channel indexes and resolve the specs.  This allows for channel priority to be specified in the `constructor.yaml` file.

Add an `--dry-run` argument which stops the process prior to fetching packages.  Allows the solution to the resolved spec to be tested and examined without downloading packages or creating an installer.

### DIFF
--- a/constructor/fcp.py
+++ b/constructor/fcp.py
@@ -191,7 +191,10 @@ def main(info, verbose=True, dry_run=False, use_conda=False):
         if use_conda:
             from conda.models.channel import prioritize_channels
             from conda.exports import fetch_index
-            index = fetch_index(prioritize_channels(info['channels']))
+            channels = tuple('%s/%s/' % (url.rstrip('/'), platform)
+                for url in info['channels']
+                for platform in (info['_platform'], 'noarch'))
+            index = fetch_index(prioritize_channels(channels))
         else:
             from libconda.fetch import fetch_index
             index = fetch_index(
--- a/constructor/fcp.py
+++ b/constructor/fcp.py
@@ -172,7 +172,10 @@ def fetch(info, use_conda):
         else:
             pkginfo = index[dist]
 
-        info['_urls'].append((urljoin(pkginfo['channel'], fn), pkginfo['md5']))
+        if not pkginfo['channel'].endswith('/'):
+            pkginfo['channel'] += '/'
+        assert pkginfo['channel'].endswith('/')
+        info['_urls'].append((pkginfo['channel'] + fn, pkginfo['md5']))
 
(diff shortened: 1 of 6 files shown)


jjhelmus:
This addressed #54.

looooo:
using this branch and `--use-conda` fixes https://github.com/conda/constructor/issues/77
Is there a chance this gets included in the next release?

hajapy:
I'd really like to encourage the addition of this feature. 

The difference between conda and constructor in terms of channel prioritization is causing major headaches for my team as we build environments from multiple source (our channel + conda-forge + defaults, in that order) and want to deploy equivalent environments using constructor for its simplicity and self-contained nature. Our environments get built by conda with very few packages coming from defaults and is mostly our stuff + conda-forge, but constructor makes a package with many packages from defaults. Our only mitigation is to mirror all the packages that conda picks when solving the environment into our own repo, which seems wasteful and adds complexity to our tooling.

mingwandroid:
This has been in constructor already for a long time. @hajapy I don't know what version or build you are using, is it very old?

hajapy:
I do not see a `--use-conda` argument for the cli interface, though it appears some methods accept the use_conda argument. 

```
constructor -V
constructor version: 2.1.0
```

```
constructor -h
Usage: constructor [options] DIRECTORY

build an installer from <DIRECTORY>/construct.yaml

Options:
  -h, --help           show this help message and exit
  --debug
  --output-dir=PATH    path to directory in which output installer is written
                       to, defaults to CWD
                       ('/Users/aachkar/workspace/deepnet/tools/jenkins')
  --cache-dir=PATH     cache directory, used for downloading conda packages,
                       may be changed by CONSTRUCTOR_CACHE, defaults to
                       '~/.conda/constructor'
  --clean              clean out the cache directory and exit
  --platform=PLATFORM  the platform for which installer is for, defaults to
                       'osx-64'
  --test               perform some self tests and exit
  --dry-run            solve package specs but do not create installer
  -v, --verbose
  -V, --version        display the version being used and exit
```

mingwandroid:
It is the only behaviour we support and thus the default.

hajapy:
I see, thanks for the help. It seems despite that I am still getting a different environment solution between constructor and conda install. I will see if I can create a minimal reproducible example and post an issue if so.

jjhelmus:
Closing, the logic in this PR was merged into constructor in #124.

jschueller:
I believe this should be reopened, because of #204 which I can confirm. A workaround can be to pin conda to 4.3.

PR #85, opened 2017-07-02
Installation instruction for gsutil

Amet13:


msarahan:
Sorry, this seems dedicated to your purpose more than it is a helpful example of how to use constructor.  I think this might be better suited as a blog post, combined with package uploads to anaconda.org as necessary.

PR #94, opened 2017-07-15
pass through verbose option to makensis

megies:
We're currently debugging a problem with nsis (building installers in appveyor) and our problem is that the verbosity CL-option is not passed through to `makensis` which is always running on low verbosity.. I think verbosity should definitly also be set for `makensis`.


msarahan:
Thanks @megies, merging.

PR #97, opened 2017-08-03
Added support for building conda installers for aarch64 (like the Jetson TX1/TX2)

thomasantony:
Includes both a minor change to `constructor.py` and an example based on Maxiconda.


msarahan:
Thanks.  merging, though I am somewhat sad to see that "bird.png" duplicated.

PR #100, opened 2017-08-17
Add option to use hardlinks to pkg cache

nehaljwani:
Constructor just extracts package tarballs to the
root env, does patching, etc and there is no package
cache. This patch enables the creation of package
cache and uses hardlinks to install pacakges inside
the root env.

### DIFF
--- a/constructor/construct.py
+++ b/constructor/construct.py
@@ -106,6 +106,13 @@ installer in the `pkgs` directory.  Using this option changes the default
 behavior.
 '''),
 
+    ('use_hardlinks',          False, bool, '''
+By default, conda packages are extracted into the root environment and then
+patched. Enabling this option will result into extraction of the packages into
+the `pkgs` directory and the files in the root environment will be hardlinks to
+the files kept in the `pkgs` directory and then patched accordingly.
+'''),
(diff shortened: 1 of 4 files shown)


nehaljwani:
Need some discussion on `post_extract` before merging this.

ilanschnell:
Constructor has already the ability to keeps the conda packages in the cache using  `keep_pkgs: True` in the `construct.yaml` file.  However, it does not extract the packages there and then hard link into the root prefix.  This PR changes fundamentally how constructor works by default, and therefore does not get my approval.  I would, however, agree to such a change, if a setting in `construct.yaml` allows hard linking (while keeping the default behavior the same).

nehaljwani:
Patch updated with the option `use_hardlinks`

Update to the description:

    This patch enables creation of hardlinks for files
    inside the root env to the files inside the package
    cache directory, using the option: `use_hardlinks`
    The default value for this option is False, for
    maintaining compatibility with the old behaviour.

mingwandroid:
Thanks for addressing @ilanschnell's objections @nehaljwani. I am merging so I can try to help out.

PR #101, opened 2017-08-20
Add support for pre-populating repodata cache

nehaljwani:
I am not sure if this is the right way to do it, since I am importing something from `conda.core.repodata`, which is internal to coda and this restricts the user to run constructor in the root env only.


mingwandroid:
@kalefranz, please review this.

mingwandroid:
Merging for now, if @kalefranz has a better way we can address later. Thanks @nehaljwani

PR #102, opened 2017-08-21
Compress non tarball files

nehaljwani:
preconda.tar.bz2 = Everything - (License.txt + pkg tarballs)


mingwandroid:
Thank you @nehaljwani

PR #109, opened 2017-09-11
Warn user if PYTHONPATH env var is set

nehaljwani:


mingwandroid:
Thanks

mingwandroid:
Hi @nehaljwani, I think we want a newline at the end of this, otherwise:

```
WARNING:
    You currently have a PYTHONPATH environment variable set. This may cause
    unexpected behavior when running the Python interpreter in Anaconda3.
    For best results, please verify that your PYTHONPATH only points to
    directories of packages that are compatible with the Python interpreter
    in Anaconda3: /root/anaconda3[root@cento632 ~]#
```

mingwandroid:
(The `#` there is my prompt after installation).

nehaljwani:
Ah, my zsh shell kept me in the dark about this. I pushed the fix.

Issue #86, opened 2017-07-02
Support for noarch conda packages?


ceball:
https://docs.continuum.io/anaconda-repository/admin/noarch-packages includes this:

> NOTE: Noarch packages are not compatible with Anaconda constructor. If you intend to use the packages with Anaconda constructor, build the packages for specific operating systems.

I couldn't find an issue about this here, so I wanted to check: is this still the case?

mingwandroid:
If it is we should fix it.

ceball:
Sorry, I missed this near the end of the README:

> Constructor does not work with noarch-Python packages. All conda packages must be available for the platform you are building the installer for.

In that case my question is: can this be a feature request, or is it not part of the plan?

Thanks

ericdill:
@mingwandroid @csoja is noarch support going to be added to constructor?

parente:
To add a concrete example: If I add a norarch package to my constructor definition today, it builds fine but upon running it, all the noarch packages wind up in `<env root>/site-packages` instead of in `<env root>/lib/python3.6/site-packages` (or whatever version of Python I've specified in the constructor). As a result, they're not in the python path and not importable. In contrast, when I `conda install` a noarch package, it winds up in `<env root>/lib/python<version>/site-packages`.

Perhaps a path prefix missing somewhere in the noarch case?

mingwandroid:
I've been cautioning people not to use `noarch: python` for a while now. It's just not ready for prime-time I'm afraid and this is a good example of that (others include permissions with .pyc files, menu shortcuts .. the list probably goes on and on).

It'd be nice to get some of this fixed, contributions are welcome.

parente:
Interesting. We're just starting to dabble in the noarch space and I, at least, was not aware there was a litany of issue.

Happy to try to contribute a fix for this initial problem here, but no one should wait on me if they get to it first.

ericdill:
@mingwandroid are you recommending against the use of `noarch: python` in general or just in the context of constructor?

mingwandroid:
I wouldn't use it until all of the issues are fixed. It's a bit chicken and egg though, I mean if no one uses it then no one will run into the problems and attempt to fix them. I'm not even sure what can be done about .pyc files in locked down scenarios.

Everyone's use-case is different though. With my Anaconda Distro hat on I have to take a very conservative approach.

mbargull:
[...]

PR #115, opened 2017-09-21
Blacklist noarch python packages

ericdill:
Constructor is not yet compatible with noarch python packages as per the
discussion in https://github.com/conda/constructor/issues/86

This "fix" allows us to continue using constructor like we have been
previously before noarch: python packages started appearing. When the
fixes described in #86 land this fix can be reverted.

### DIFF
--- a/constructor/fcp.py
+++ b/constructor/fcp.py
@@ -181,7 +181,8 @@ def main(info, verbose=True):
                         for url in info['channels']
                         for platform in (info['_platform'], 'noarch')))
         # drop all noarch: python packages because constructor is not noarch:
-        # python compatible
+        # python compatible. Track that progress here
+        # https://github.com/conda/constructor/issues/86
         index = {pkg_name: pkg_info for pkg_name, pkg_info in iteritems(index)
                  if pkg_info.get('noarch') != 'python'}
 
--- a/constructor/fcp.py
+++ b/constructor/fcp.py
@@ -16,6 +16,7 @@ from os.path import isdir, isfile, join
 
 from libconda.fetch import fetch_index, fetch_pkg
 from libconda.resolve import Resolve, NoPackagesFound
+from libconda.compat import iteritems
 
 from constructor.utils import md5_file
 from constructor.install import name_dist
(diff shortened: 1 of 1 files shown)


jakirkham:
Would be great if someone could look at this. It is at least some improvement on the `noarch` situation with `constructor`.

mingwandroid:
There's no point looking at this while it has conflicts.

AFAICT the end result of this approach is that old software that predates conda-forge's decision to prefer `noarch python` packages will get selected. Is that *really* a good outcome?

jakirkham:
The result will be not producing broken installers quietly, which actually happens now and bit me recently.

mingwandroid:
Broken because noarch python doesn't work? It will produce broken installers though and it will drop new packages for old ones, and it will do that silently. I honestly do not see the overall benefit of this.

jakirkham:
No. Broken because `constructor` does not understand how `noarch: python` works. While `constructor` tries to include them, they cannot be imported, cannot be remove (by `conda`), and cannot be upgraded. It might be possible to force install them, but have no idea what kind of break one gets then.

mingwandroid:
You have ignored the main thrust of my question twice now.

What use is an installer generator that (silently, or at all) creates installers based upon old software?

mingwandroid:
I should also point out, since `noarch: python` will always require `conda` to be able to install it, this precludes `conda-forge` from marking *any* of `conda`'s dependencies as being `noarch: python`. I hope that fact new to you though!

mingwandroid:
@ericdill, if you are still interested in this PR, then please address these issues:

1. It needs to be rebased.
2. It needs to warn in the strongest possible terms that it is dropping newer noarch packages for older non-noarch ones  (but not that it is dropping hundreds of packages that would never have made it into the installer finally anyway).

ankostis:
Really stryugling with this `noarch` packages hidden in the stack...isn't it possible to had-fail (raise an exception) if a `noarch` package is met, until #86 is fixed?

msarahan:
constructor supports noarch packages as of #263 - closing

PR #126, opened 2017-11-03
remove broken test test_link_idists

mbargull:
The test case is outdated since `install.link_idists` is re-implemented now.


nehaljwani:
Thanks @mbargull !

PR #134, opened 2017-11-09
Write basic system information into output

### LINKED ISSUE #130: Store build system information
Constructor currently stored the absolute set of of packages (URL/version/hash) in a file `pkgs/urls` within the tarball. This is great! However, that is not all that affects the binary that is produced; I would like to see another file alongside with the information on the system that produced it, including versions of constructor, python, conda and platform information.

martindurant:
Fixes #130

### DIFF
--- a/constructor/preconda.py
+++ b/constructor/preconda.py
@@ -15,7 +15,7 @@ try:
 except:
     import ruamel_json as json
 
-files = 'urls', 'urls.txt', '.install.py'
+files = 'system.info', 'urls', 'urls.txt', '.install.py'
 
 def write_index_cache(info, dst_dir):
     global files
@@ -87,7 +87,21 @@ def get_final_url(info, url):
             return new_url
     return url
 
+
+def system_info():
+    import constructor, conda, sys
+    return {'constructor': constructor.__version__,
+            'conda': conda.__version__,
+            'platform': sys.platform,
+            'python': sys.version,
+            'python_version': tuple(sys.version_info)}
+
+
 def write_files(info, dst_dir):
+    print('Write Files')
+    with open(join(dst_dir, 'system.info'), 'w') as fo:
+        json.dump(system_info(), fo)
+
     with open(join(dst_dir, 'urls'), 'w') as fo:
         for url, md5 in info['_urls']:
             fo.write('%s#%s\n' % (get_final_url(info, url), md5))
--- a/constructor/preconda.py
+++ b/constructor/preconda.py
@@ -89,7 +89,8 @@ def get_final_url(info, url):
 
 
 def system_info():
-    import constructor, conda, sys, platform
+    import constructor, sys, platform
+    import conda.exports, conda.config
     out = {'constructor': constructor.__version__,
            'conda': conda.__version__,
            'platform': sys.platform,
--- a/constructor/preconda.py
+++ b/constructor/preconda.py
@@ -89,16 +89,31 @@ def get_final_url(info, url):
 
 
 def system_info():
-    import constructor, conda, sys
-    return {'constructor': constructor.__version__,
-            'conda': conda.__version__,
-            'platform': sys.platform,
-            'python': sys.version,
-            'python_version': tuple(sys.version_info)}
+    import constructor, conda, sys, platform
+    out = {'constructor': constructor.__version__,
+           'conda': conda.__version__,
+           'platform': sys.platform,
+           'python': sys.version,
+           'python_version': tuple(sys.version_info),
+           'machine': platform.machine(),
+           'platform_full': platform.version()}
+    if sys.platform == 'darwin':
+        out['extra'] = platform.mac_ver()
+    elif sys.platform.startswith('linux'):
+        out['extra'] = platform.dist()
+    elif sys.platform.startswith('win'):
+        out['extra'] = platform.win32_ver()
+        prefix = os.environ.get('CONDA_PREFIX', conda.config.default_prefix)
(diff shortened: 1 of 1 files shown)


martindurant:
xref https://github.com/Anaconda-Platform/anaconda-platform/pull/982 , https://github.com/Anaconda-Platform/anaconda-platform/pull/2602

constructor/preconda.py
```diff
@@ -15,7 +15,7 @@
 except:
     import ruamel_json as json
 
-files = 'urls', 'urls.txt', '.install.py'
+files = 'system.info', 'urls', 'urls.txt', '.install.py'
```
kalefranz: So it appears this is going in the `pkgs/` package cache directory.  Is `system.info` the correct name for the file?  I'm not sure.  Which "system" are we talking about?  The system we're unpacking to, or the system that built the installer...   Would something like `constructor-build.info` be a better name?  Maybe better would be `.constructor-build.info`.

constructor/preconda.py
```diff
@@ -87,7 +87,20 @@ def get_final_url(info, url):
...
 
+
+def system_info():
+    import constructor, conda, sys
+    return {'constructor': constructor.__version__,
+            'conda': conda.__version__,
+            'platform': sys.platform,
+            'python': sys.version,
+            'python_version': tuple(sys.version_info)}
+
```
nehaljwani: Should we also dump the nsis version here (for the windows installer)?

martindurant:
@nehaljwani - sure! How do I get that?

constructor/preconda.py
nehaljwani: Not sure if this is the most correct way, but maybe like this?
```
(root) F:\nwani\constructor>%CONDA_PREFIX%\NSIS\makensis.exe /version
v3.01
```
Or maybe we could somehow ask conda to give the exact package version (along with complete build string) of the installed package with the name 'nsis'

nehaljwani: Not sure how useful this might be, but since we are talking about system info, for macOS, I'd also like to see the value for `sw_vers` and for Windows, the value of: 
```
(root) F:\nwani\constructor>systeminfo | findstr /B /C:"OS Name" /C:"OS Version"
OS Name:                   Microsoft Windows 7 Professional
OS Version:                6.1.7601 Service Pack 1 Build 7601
```

kalefranz: In conda 4.4, one way is

```python
pd = PrefixData(target_prefix_path)
prefix_rec = pd.get('nsis')
if prefix_rec:
    print(prefix_rec.version)
```

The `PrefixData` class should have a `query()` method, just like `PackageCacheData` and `SubdirData` (i.e. repodata).  Unfortunately it doesn't currently it looks like.  We could add that though.  And in that case, the preferred way would be

```python
pd = PrefixData(target_prefix_path)
prefix_rec = next(pd.query(MatchSpec('nsis')), None)
if prefix_rec:
    print(prefix_rec.version)
```

In conda 4.3, which will also work for conda 4.4 because of backward compatibility maintained between 4.3 and 4.4 (might not be maintained indefinitely)

```python
from conda.exports import linked_data
prefix_records = linked_data(target_prefix_path).values()
nsis_prefix_rec = next((rec for rec in prefix_records if rec.name == 'nsis'), None)
if nsis_prefix_rec:
    print(nsis_prefix_rec.version)
```

constructor/preconda.py
```diff
@@ -87,7 +87,35 @@ def get_final_url(info, url):
...
+           'python_version': tuple(sys.version_info),
+           'machine': platform.machine(),
+           'platform_full': platform.version()}
+    if sys.platform == 'darwin':
+        out['extra'] = platform.mac_ver()
+    elif sys.platform.startswith('linux'):
+        out['extra'] = platform.dist()
+    elif sys.platform.startswith('win'):
+        out['extra'] = platform.win32_ver()
+        prefix = os.environ.get('CONDA_PREFIX', conda.config.default_prefix)
```
nehaljwani: On Windows, If I just call system_info(), this fails with:
```
AttributeError: module 'conda' has no attribute 'config'
```

constructor/preconda.py
```diff
@@ -87,7 +87,35 @@ def get_final_url(info, url):
...
+        prefix_records = conda.exports.linked_data(prefix).values()
```
nehaljwani: On Windows, this fails with:
```
AttributeError: module 'conda' has no attribute 'exports'
```

nehaljwani: This should probably be: `from conda.exports import linked_data`

goanpeca:
We should write some tests and add CI ? Seems like some things could be easily caught by ci

nehaljwani:
@goanpeca PRs are most welcome for tests!

goanpeca:
Yep i will add something :-)

goanpeca:
@nehaljwani, @martindurant https://github.com/conda/constructor/pull/137

martindurant:
@nehaljwani - I can't merge this.
@goanpeca - tests can follow in your PR after merger here (although I cannot push to your branch)

kalefranz:
I was out sick on Friday.  This PR has a few issues with interacting with conda that I'll submit another PR to resolve, since this one has been merged.

constructor/preconda.py
martindurant: @kalefranz , the file name changed below but not here! Can you please throw in a quick change to fix this. Also, double second @goanpeca to actually do some testing along the way.

martindurant: https://github.com/conda/constructor/pull/142/

PR #141, opened 2017-11-14
reorder paths

msarahan:
Having Library/usr/bin and Library/mingw-w64/bin ahead of Library/bin bit me pretty hard today with curl.  There was a git4win curl version shadowing ours, and that git4win one was busted.


msarahan:
Corresponding conda PR at https://github.com/conda/conda/pull/6307

msarahan:
@mingwandroid and I discussed this on flowdock. He is concerned that this will break msys2 stuff. I have tried to bundle a more minimal git4win package, and we'll try that before we muck with this. Closing for now.

PR #154, opened 2018-01-05
Change ownership back to $USER after updating dot profile

nehaljwani:
If the user chooses to install to a custom location, he is prompted for
a password and the modification to the dot profile file for PATH
modification happens as root user. Since $USER is still set to the
original user who spawned the installer, we can use it to reset the
ownership of the dot profile and it's backup back to $USER.


mingwandroid:
LGTM.

Issue #164, opened 2018-01-19
Warn if trying to install on wrong platform


birdsarah:
A feature request to protect me from myself!

I decided to can my current miniconda3 installation and start a new one.

I downloaded the installer from https://conda.io/miniconda.html ran `bash Miniconda......` but it wouldn't install - see error below. This has never happened before. I tried again. I re-downloaded it. I panicked. I went to see @kalefranz, he asked me where I downloaded it from. I showed him.

He pointed out that I'd downloaded the Linux sh, not the OSX one. 🤦‍♀️ 

As someone who self-identifies as a Linux user but is on OSX this was an easy enough mistake to make. It would be a lovely feature, if instead of getting the failure below, there was a check and a simple "cannot install Linux version on OSX" message.

```
........

Do you accept the license terms? [yes|no]
[no] >>> yes

Miniconda3 will now be installed into this location:
/Users/sbird/miniconda3

  - Press ENTER to confirm the location
  - Press CTRL-C to abort the installation
  - Or specify a different location below

[/Users/sbird/miniconda3] >>> /Users/sbird/wrong_install_platform
PREFIX=/Users/sbird/wrong_install_platform
Miniconda3-latest-Linux-x86_64.sh: line 315: md5sum: command not found
WARNING: md5sum mismatch of tar archive
expected: 9cd52bb8c124af765e86192894e41638
     got: 
installing: python-3.6.3-h6c0c0dc_5 ...
Miniconda3-latest-Linux-x86_64.sh: line 353: /Users/sbird/wrong_install_platform/pkgs/python-3.6.3-h6c0c0dc_5/bin/python: cannot execute binary file
```

PR #165, opened 2018-01-21
Add more os, arch checks to the shell installer

nehaljwani:
Fixes https://github.com/conda/constructor/issues/164
Fixes https://github.com/ContinuumIO/anaconda/issues/1126

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -213,6 +213,57 @@ then
     fi
 #endif
 
+#if ppc64le
+    if [ "$(uname -m)" != "ppc64le" ]; then
+        printf "WARNING:\\n"
+        printf "    Your machine hardware does not appear to be Power8 (little endian), \\n"
+        printf "    but you are trying to install a ppc64le version of __NAME__.\\n"
+        printf "    Are sure you want to continue the installation? [yes|no]\\n"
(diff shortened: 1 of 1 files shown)


constructor/header.sh
```diff
@@ -213,6 +213,57 @@ then
     fi
 #endif
 
+#if ppc64le
+    if [ "$(uname -m)" != "ppc64le" ]; then
+        printf "WARNING:\\n"
+        printf "    Your machine hardware does not to appear be Power8 (little endian), \\n"
```
mingwandroid: Please change:

> does not to appear be

.. to:

> does not appear to be

(in all of these)

mingwandroid: Can we also have a way to skip this check? If `binfmt_misc` supports the architecture of the installer then I would not like to be forced to interactively confirm this, also what about when using `--force`? That usually means "no user interaction", should this abort automatically in this case, then do we need `--really-force` to indicate we don't care that `uname` is mismatched?

nehaljwani: All of these checks are skipped if -b is specified.

nehaljwani: Fixed in the commit below.

mingwandroid:
Thanks Nehal

PR #166, opened 2018-01-22
Bye bye libconda!

### LINKED ISSUE #139: Consider removing libconda support from constructor
As @nehaljwani mentioned in #130, with libconda growing further and further out of date from conda, it might be appropriate to remove libconda support from constructor.  When constructor is using libconda, end-users will often end up with broken behavior that is subtle and not easy to trace back to libconda vs conda.

The drawback is that as of conda 4.3, conda can only be installed in the root conda environment.  This restriction is lifted in conda 4.4.

nehaljwani:
Fixes #139

### DIFF
--- a/constructor/conda_interface.py
+++ b/constructor/conda_interface.py
@@ -9,12 +9,8 @@ try:
     from conda import __version__ as CONDA_INTERFACE_VERSION
     conda_interface_type = 'conda'
 except ImportError:
-    try:
-        from libconda import __version__ as CONDA_INTERFACE_VERSION
-        conda_interface_type = 'libconda'
-    except ImportError:
-        raise RuntimeError("Either conda or libconda must be installed for python interpreter\n"
-                           "with sys.prefix: %s" % sys.prefix)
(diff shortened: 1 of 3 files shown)


kalefranz:
If we merge this, we should probably bump the major version of constructor to 3.0, no?

nehaljwani:
libconda was already unusable with the 2.0.0 release, which was a major bump. This commit just removed the unnecessary burden. Should we still make the next release 3.0.0 ? Thoughts, @mingwandroid ?

kalefranz:
> libconda was already unusable with the 2.0.0 release, which was a major bump

In that case I'm fine keeping major version at 2.

mingwandroid:
2.x is ok with me. 3 should be a rewrite to use a proper installation system and a static installer executable ;-)

kalefranz:
And `libconda` should actually be... you know... a dynamic library you can link compiled code against.

mingwandroid:
it could be built as both static and shared, incase you wanted to make a static executable out of it.

kalefranz:
Yes, exactly.

PR #173, opened 2018-02-12
Update construct.py

jschueller:


nehaljwani:
With this change, run `python construct.py` and it will generate CONSTRUCT.md

Check that in as well! I probably forgot to mention that part in the previous PR :(

PR #180, opened 2018-03-07
Rework markdown generation

### LINKED ISSUE #179: Valid keys in construct.yaml
Would you accept a PR that updates the doc generation code in construct.py to whether or not each key is required and what the valid argument types are?

ericdill:
* Use jinja2 for cleaner templating
* Update markdown syntax
* Move doc generation into scripts/ dir
* Update ``make doc`` to use the script in scripts/

Closes #179

### DIFF
--- a/CONSTRUCT.md
+++ b/CONSTRUCT.md
@@ -49,9 +49,10 @@ required: False
 
 argument type(s): ``list``, 
 
-List of (src, dest) channels, from which, channels from 'src' are also
+List of `(src, dest)` channels, from which, channels from `src` are also
 considered while running solver, but are replaced by corresponding values from
-dest when writing urls{,.txt}. Example use:
+dest when writing `urls{,.txt}`. Example use:
+```
 channels_remap:
   -
       src: file:///tmp/a3/conda-bld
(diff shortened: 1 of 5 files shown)


ericdill:
Hi @mingwandroid @nehaljwani -- Here's my first pass. Changes that may be controversial:

1. using jinja2
2. moving doc generation out of the source code
3. using `#`/`##` markdown syntax instead of `====`/`-----`

Thoughts?

msarahan:
Seems fine to me.  Does jinja2 need to be added as a dependency somewhere/somehow?

ericdill:
> Seems fine to me. Does jinja2 need to be added as a dependency somewhere/somehow?

Needs to be installed on the host wherever the markdown doc is generated. I don't think that's happening on travis or appveyor but I only just skimmed those CI files.

I didn't want to pick up jinja2 as a runtime dep for constructor which is why I moved that doc generation out of the package and into scripts. Lemme see if there's a section in the readme for how to make the docs and, if not, i'll add one

ericdill:
attn @mingwandroid @nehaljwani

PR #188, opened 2018-05-15
resolve #138 properly populate conda-meta/history file

### LINKED ISSUE #138: Properly populate conda-meta/history file
Follow-on issue from #130 and https://github.com/conda/constructor/issues/130#issuecomment-342587752

> constructor should probably be writing the history file correctly anyway.
> 
> I guess by including this information in `conda-meta/history`, I proposing something like
> 
> ```
> ==> 2017-11-07 12:28:46 <==
> # cmd: /Users/kfranz/continuum/conda/conda/__main__.py create -n _newer python constructor -c > defaults
> # constructor: {"constructor_version": "2.0.0", "conda_interface_type": "conda", "conda_interface_version": "4.3.30"}
> +defaults::asn1crypto-0.22.0-py36hb705621_1
> +defaults::ca-certificates-2017.08.26-ha1e5d58_0
> +defaults::certifi-2017.7.27.1-py36hd973bb6_0
> ...
> +defaults::xz-5.2.3-ha24016e_1
> +defaults::yaml-0.1.7-hff548bb_1
> +defaults::zlib-1.2.11-h60db283_1
> # create specs: python,constructor
> ```
> 
> The new part here is the `constructor:` line.  But I guess I'm also advocating here that constructor should be including all the relevant information in the `conda-meta/history` file.  This will be especially important starting for conda 4.4, when conda starts to treat different packages that result from an explicit user-supplied spec, and packages that are brought in as dependencies to satisfy requirements.

Currently, constructor only touches the `conda-meta/history` file to ensure that it exists.

kalefranz:
resolve #138

### DIFF
--- a/constructor/construct.py
+++ b/constructor/construct.py
@@ -56,7 +56,9 @@ channels_remap:
     ('specs',                  False, (list, str), '''
 List of package specifications, e.g. `python 2.7*`, `pyzmq` or `numpy >=1.8`.
 This list of specifications if given to the conda resolver (as if you were
-to create a new environment with those specs.
+to create a new environment with those specs. The packages may also be 
+specified by their entire URL, 
+e.g.`https://repo.continuum.io/pkgs/free/osx-64/openssl-1.0.1k-1.tar.bz2`.
 '''),
 
     ('user_requested_specs',                  False, (list, str), '''
@@ -69,14 +71,6 @@ specs will fall back to 'specs'.
 List of package names to be excluded, after the '`specs` have been resolved.
 For example, you can say that `readline` should be excluded, even though it
 is contained as a result of resolving the specs for `python 2.7`.
-'''),
-
-    ('packages',               False, (list, str), '''
-A list of explicit conda packages to be included, e.g. `yaml-0.1.6-0.tar.bz2`.
-The packages may also be specified by their entire URL,
-e.g.`https://repo.anaconda.com/pkgs/main/osx-64/openpyxl-2.4.10-py36_0.tar.bz2`.
-Optionally, the MD5 hash sum of the package, may be added after an immediate
-`#` character, e.g. `readline-6.2-2.tar.bz2#0801e644bd0c1cd7f0923b56c52eb7f7`.
 '''),
 
     ('menu_packages',           False, list, '''
--- a/constructor/construct.py
+++ b/constructor/construct.py
@@ -59,6 +59,12 @@ This list of specifications if given to the conda resolver (as if you were
 to create a new environment with those specs.
 '''),
 
+    ('user_requested_specs',                  False, (list, str), '''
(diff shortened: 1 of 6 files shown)


constructor/install.py
```diff
@@ -228,8 +228,8 @@ def create_meta(prefix, dist, info_dir, extra_info):
     meta_dir = join(prefix, 'conda-meta')
     if not isdir(meta_dir):
         os.makedirs(meta_dir)
-        with open(join(meta_dir, 'history'), 'w') as fo:
-            fo.write('')
+    #     with open(join(meta_dir, 'history'), 'w') as fo:
+    #         fo.write('')
     with open(join(meta_dir, dist + '.json'), 'w') as fo:
```
kalefranz: Don't leave this commented out.  Actually remove it.

constructor/preconda.py
```diff
@@ -36,9 +37,7 @@ def write_index_cache(info, dst_dir):
         write_repodata(cache_dir, url)
 
     for cache_file in os.listdir(cache_dir):
-        if cache_file.endswith(".json"):
-            files += join(cache_dir, cache_file),
```
kalefranz: This `+=` pattern had no effect because `files` was already imported in `shar.py`.

constructor/shar.py
```diff
@@ -92,6 +92,7 @@ def create(info, verbose=False):
         for cf in os.listdir(cache_dir):
             if cf.endswith(".json"):
                 p_t.add(join(cache_dir, cf), 'pkgs/cache/' + cf)
+    p_t.add(join(tmp_dir, 'conda-meta', 'history'), 'conda-meta/history')
```
kalefranz: Decided to do this at installer create time rather than install time because, if we did it at install time, we'd have to carry some of this information around in another data structure somehow.  Doing it at installer create time is just easier.

constructor/construct.py
```diff
@@ -74,6 +74,12 @@
 List of package specifications, e.g. `python 2.7*`, `pyzmq` or `numpy >=1.8`.
 This list of specifications if given to the conda resolver (as if you were
 to create a new environment with those specs.
+'''),
+
+    ('user_requested_specs',                  False, (list, str), '''
+List of package specifications to be recorded as "user-requested" for the 
+initial environment in conda's history file. If not given, user-requested
+specs will fall back to 'specs'. 
 '''),
```
kalefranz: Note the new config parameter.  We'll need to use this config parameter for Miniconda and Anaconda.  We should discuss what the values should be.

nehaljwani: Could we use this and drop the 'packages' option? It doesn't work anyway.

kalefranz:
@nehaljwani From my perspective I believe this is ready. Do you see anything else that we should modify?

PR #191, opened 2018-05-18
fix url translation

kalefranz:
This fix is **not** compatible with conda 4.6 / current conda master branch.  However that's beyond the scope for the Anaconda 5.2 release.


nehaljwani:
Thanks @kalefranz !

PR #196, opened 2018-05-22
Fix location of components.plist files and add cleanup

nehaljwani:


nehaljwani:
xref: https://github.com/ContinuumIO/anaconda/issues/1436

PR #198, opened 2018-05-24
Report maximum size that the pkg installer can take

nehaljwani:
Complete explanation for this patch is in the dev comment under
pkgbuild()

TL;DR: Double the size of each .pkg component in distribution.xml (post
creation), so that the pkg installer is never given the chance to report less
size; i.e, play safe.

This shields us from two problems:
  - hard-linking fails and a fall back to copy takes place
  - finder doesn't account for hardlinks in get-info for the installation folder

### DIFF
--- a/constructor/osxpkg.py
+++ b/constructor/osxpkg.py
@@ -1,4 +1,5 @@
 import os
+import re
 import shutil
 import tarfile
 from os.path import dirname, exists, join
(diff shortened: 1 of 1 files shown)


nehaljwani:
1. Copy will take place if the destination filesystem does not support hard linking.
2.  Pretty much, yeah

The third alternative is not use hard links at all while creating the AD installers. So no finder bug and no fallback copy and the reported size is sane.

mingwandroid:
> 1. Copy will take place if the destination filesystem does not support hard linking
APFS and HFS+ both support this, can we agree we are talking about a tiny % of users here who'd use anything else?

I'd just waive this bug and anyone who complains we'd point them to a FAQ item personally and make the PR do the doubling if `attempt_hardlinks` is set only.

nehaljwani:
I'm fine with that.

cc @csoja

mingwandroid:
My biggest (but not only) concern here is that people who have enough space would end up unable to install the software btw, I should have made that clear earlier.

mcg1969:
Based on what I'm reading here I think we should close this. I do think it is reasonable for us to assume that the strong majority of `.pkg` installer users will be on a filesystem that supports hard linking. And even for those it will only be an issue if someone is truly near filling that disk. Please feel free to re-open if I'm wrong here.

PR #208, opened 2018-07-11
If prefix exists and it's a force/update, say reinstall

### LINKED ISSUE #207: Update option in generated installer says 'installing'
This is straight forward:

When you call an installer created with constructor with -u, which in the help says 'update' the message displayed says 'installing' instead of update and it seems like a new installation

Basically `-u` option in the `header.sh`   recalls ./install.py which I guess redisplays 'installing message' instead of something saying 'updating' or 'reinstalling'

nehaljwani:
Resolve #207

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -66,6 +66,7 @@ BATCH=0
 FORCE=0
 SKIP_SCRIPTS=0
 TEST=0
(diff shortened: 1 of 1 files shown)


nehaljwani:
@alep @fpliger

Does this change seem okay to you?

Issue #88, opened 2017-07-05
On Windows, constructor-generated installer creates an uninstall entry under 'python'.


ceball:
When a user comes to uninstall an application, they would expect to find it listed under 'application name' in the control panel, but instead it appears as an entry starting with 'python'.

mcg1969:
Looks like it was fixed in #212

PR #212, opened 2018-07-25
new 'uninstall_name' config param

speleo3:
Example which puts the application name first (instead of "Python X.Y", see #88):
```yaml
uninstall_name: "${NAME} ${VERSION} (Python ${PYVERSION} ${ARCH})"
```


speleo3:
thanks @nehaljwani for your comment. What do you suggest instead? Make `"${NAME} ${VERSION} (Python ${PYVERSION} ${ARCH})"` the new default? That would certainly work for me.

PR #217, opened 2018-09-03
fcp refactor for better use of conda

kalefranz:


nehaljwani:
The python2.7 test is failing because:
Installing pillow also installs xz into the package cache directory. And then because of the issue highlighted at https://github.com/conda/conda/issues/7769 , conda actually still honors both package cache directories, ~/condacache/pkgs (via pkgs_dirs in ~/.condarc) and ~/.conda/constructor/linux-64 and doesn't download xz as it thinks it is already there, which breaks the assumption of the `info/repodata_record.json` injection code while creating preconda.tar.bz2.

PR #219, opened 2018-09-14
Pave the path for conda init

nehaljwani:
- Replace add_to_path_default with initialize_by_default
- If $PREFIX/etc/profile.d/conda.sh exists, prefer sourcing it

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -474,15 +474,15 @@ if [ "$BATCH" = "0" ]; then
     BASH_RC="$HOME"/.bashrc
     DEFAULT=no
 #endif
-#if add_to_path_default is True
+#if initialize_by_default is True
     DEFAULT=yes
 #endif
-#if add_to_path_default is False
+#if initialize_by_default is False
     DEFAULT=no
 #endif
 
-    printf "Do you wish the installer to prepend the __NAME__ install location\\n"
-    printf "to PATH in your %s ? [yes|no]\\n" "$BASH_RC"
+    printf "Do you wish the installer to initialize __NAME__\\n"
+    printf "in your %s ? [yes|no]\\n" "$BASH_RC"
     printf "[%s] >>> " "$DEFAULT"
     read -r ans
     if [ "$ans" = "" ]; then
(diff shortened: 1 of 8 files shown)


constructor/header.sh
```diff
@@ -492,28 +492,42 @@ if [ "$BATCH" = "0" ]; then
...
         printf "\\n"
-        printf "You may wish to edit your .bashrc to prepend the __NAME__ install location to PATH:\\n"
+        printf "You may wish to edit your $BASH_RC to setup __NAME__:\\n"
         printf "\\n"
-        printf "export PATH=%s/bin:\$PATH\\n" "$PREFIX"
+        if [ -f "$PREFIX/etc/profile.d/conda.sh" ]; then
+            printf "source $PREFIX/etc/profile.d/conda.sh\\n"
+        else
+            printf "export PATH=\"$PREFIX/bin:\$PATH\"\\n"
+        fi
```
kalefranz: I would make this part just

    printf "source $PREFIX/etc/profile.d/conda.sh\\n"

Not any of the other fallback code.

constructor/header.sh
```diff
@@ -492,28 +492,42 @@ if [ "$BATCH" = "0" ]; then
...
+# added by __NAME__ __VERSION__ installer
+# >>> conda initialize >>>
+# !! Contents within this block are managed by 'conda init' !!
+if [ -f "$PREFIX/etc/profile.d/conda.sh" ]; then
+    . "$PREFIX/etc/profile.d/conda.sh"
+else
+    export PATH="$PREFIX/bin:\$PATH"
+fi
+# <<< conda initialize <<<
+EOF
```
kalefranz: We should go ahead and make this now what `conda init` will eventually do.  So the full block that `conda init` adds now, and actually exercise the first fallback until people get the new conda.

```
# >>> conda initialize >>>
# !! Contents within this block are managed by 'conda init' !!
__conda_setup="\$('$PREFIX/bin/conda' shell.bash hook 2> /dev/null)"
if [ \$? -eq 0 ]; then
    eval "\$__conda_setup"
else
    if [ -f "$PREFIX/etc/profile.d/conda.sh" ]; then
        . "$PREFIX/etc/profile.d/conda.sh"
    else
        export PATH="$PREFIX/bin:$PATH"
    fi
fi
unset __conda_setup
# <<< conda initialize <<<
```

constructor/osx/update_path.sh
```diff
@@ -22,13 +22,25 @@ BASH_RC_BAK="${BASH_RC}-__NAME_LOWER__.bak"
...
 # added by __NAME__ __VERSION__ installer
-export PATH=\"$PREFIX/bin:\$PATH\"" >>$BASH_RC
+# >>> conda initialize >>>
+# !! Contents within this block are managed by 'conda init' !!
+if [ -f "$PREFIX/etc/profile.d/conda.sh" ]; then
+    . "$PREFIX/etc/profile.d/conda.sh"
+else
+    export PATH="$PREFIX/bin:\$PATH"
+# <<< conda initialize <<<
+EOF
```
kalefranz: Same as above.  Use what `conda init` will be putting in I think.

constructor/header.sh
nehaljwani: Seems legit, since we are going to bump the minimum req on conda anyway.

constructor/header.sh
nehaljwani: I deliberately removed
```
'$PREFIX/bin/conda' shell.bash hook 2> /dev/null
```
Because it throws an exception at the moment and that makes opening a new shell incredibly slow.

constructor/osx/update_path.sh
nehaljwani: Same as before. Don't want users to blame Anaconda for making their shell instantiation slow.

constructor/header.sh
nehaljwani: To be more specific, this throws a Not Implemented Exception.

kalefranz:
xref: https://github.com/conda/conda/issues/7781

PR #214, opened 2018-07-31
Ignore noarch python packages

ericpre:
Update PR #115.

I have used this branch to build installer for linux, windows and mac and it seems to do what it is supposed to do.

PR #225, opened 2018-10-18
Overload Solver._prepare method to ignore noarch packages.

H0R5E:
Overload the _prepare method and apply the code in PR #214. This is a temporary fix for issue #86.

As the new code is appended to the superclass method, an additional initialisation of Resolve is required, which is not ideal. I can't see an easy way around that, as I don't want to rewrite the whole method.

I have also added new examples (one which deliberately fails) and updated run_examples to test them. I know this is a bit weird to have a failing example, but I didn't see another way to test excluding noarch easily. Also, I've only tried this out on Windows.


nehaljwani:
This PR looks good to me, however, if the support for noarch packages does get added in the future, should we make the skipping optional? Something like `skip_noarch_packages:` and default it to `True`?

@kalefranz What do you think?

H0R5E:
@nehaljwani I would agree that is a good idea as some people will prefer to post-process noarch packages, as mentioned in [this comment](https://github.com/conda/constructor/issues/86#issuecomment-358895352).

I will have a go at adding the option.

Mat

H0R5E:
OK, so I added an keyword option called `exclude_noarch_packages` (which is set to `True` by default) immediately following the `exclude` keyword .

I switched to decorating the _prepare method rather than creating a subclass so that it was easier to turn on and off the new functionality.

I also added yet another example which will only compile if the exclude_noarch_packages option is set to `False`.

nehaljwani:
@kalefranz I know we talked about using `CONDA_SUBDIRS` to drop noarch (which wouldn't require any changes to constructor), but this PR specifically removes `noarch: python` and lets the developer still use `noarch: generic`

nehaljwani:
Also, a slightly related PR, which may change the way we compute channel urls: https://github.com/conda/constructor/pull/231 . If this PR might get approved in the future, should I avoid dropping 'noarch' in the subdirs argument to all_channel_urls?

kalefranz:
> I know we talked about using `CONDA_SUBDIRS` to drop noarch (which wouldn't require any changes to constructor), but this PR specifically removes `noarch: python` and lets the developer still use `noarch: generic`

That makes sense.  I get it now.

So I've looked at the code.  To me, the approach seems pretty brittle.  That `_prepare` method starts with an underscore for a reason.  A better approach would be to make a test-enforced change to conda itself implementing an `exclude_noarch_python` config parameter (start with it undocumented).

H0R5E:
Could it be combined with the functionality provided here?

https://github.com/conda/conda/pull/6932

msarahan:
as of #263 constructor supports noarch packages.  Thanks for helping to push this issue.

PR #231, opened 2018-11-03
Delegate channel_urls computation to Conda

nehaljwani:


constructor/fcp.py
```diff
@@ -220,7 +221,7 @@ def main(info, verbose=True, dry_run=False):
     version = info["version"]
     download_dir = info["_download_dir"]
     platform = info["_platform"]
-    channel_urls = info.get("channels", ())
+    channel_urls = all_channel_urls(info.get("channels", ()), subdirs=[platform])
```
nehaljwani: Deliberately leaving out noarch here.

kalefranz: I guess we're going to have to be more sophisticated here if we want to be able to include `noarch: generic`, but exclude `noarch: python`.

mcg1969:
It seems to me this can and should be closed since #306 addresses this (if not other future PRs).

Issue #226, opened 2018-11-02
Issue in Constructor 2.2.2


elijahomolo:
Constructor is throwing the following error:

```
Checking for duplicate files ...


Traceback (most recent call last):

File "C:\Users\toby\miniconda3-64\Scripts\constructor-script.py", line 10, in <module>
sys.exit(main())
File "C:\Users\toby\miniconda3-64\lib\site-packages\constructor\main.py", line 221, in main
dry_run=opts.dry_run)
File "C:\Users\toby\miniconda3-64\lib\site-packages\constructor\main.py", line 113, in main_build
fcp_main(info, verbose=verbose, dry_run=dry_run)
File "C:\Users\toby\miniconda3-64\lib\site-packages\constructor\fcp.py", line 237, in main
ignore_duplicate_files, verbose, dry_run
File "C:\Users\toby\miniconda3-64\lib\site-packages\constructor\fcp.py", line 207, in _main
pc_recs, platform, ignore_duplicate_files
File "C:\Users\toby\miniconda3-64\lib\site-packages\constructor\fcp.py", line 120, in check_duplicates_files
size = path_data.size_in_bytes or getsize(join(extracted_package_dir, short_path))
AttributeError: 'PathData' object has no attribute 'size_in_bytes'

```

Downgrading to version 2.1.1 fixes the issue. Modifying miniconda3-64\lib\site-packages\constructor\fcp.py line 120 by removing the path_data.size_in_bytes call and leaving the 'size' variable just as getsize(join(extracted_package_dir, short_path)) also worked as a workaround.

nehaljwani:
Which version of conda is present in the environment where you are trying to run constructor?

H0R5E:
I can confirm this. I've attached the patch I used to fix it with a try-except block, as the "or" statement didn't work when the size_in_bytes attribute was missing from the PathData variant.

```
>conda info

     active environment : base
    active env location : C:\Anaconda\envs\_constructor
            shell level : 1
       user config file : C:\Users\Work\.condarc
 populated config files : C:\Users\Work\.condarc
          conda version : 4.5.11
    conda-build version : not installed
         python version : 2.7.15.final.0
       base environment : C:\Anaconda\envs\_constructor  (writable)
           channel URLs : https://repo.anaconda.com/pkgs/main/win-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/free/win-64
[...]

PR #233, opened 2018-11-03
Replace "or" statement in first attempt to retrieve path_data size.

H0R5E:
Offers a more robust solution to #226 as it covers the case where size_in_bytes returns a value which casts to False.

### DIFF
--- a/constructor/fcp.py
+++ b/constructor/fcp.py
@@ -118,7 +118,8 @@ def check_duplicates_files(pc_recs, platform, ignore_duplicate_files=False):
         for path_data in paths_data:
             short_path = path_data.path
             try:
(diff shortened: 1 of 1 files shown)


H0R5E:
Hi @nehaljwani, I found an example of size_in_bytes being None in https://github.com/conda/conda/blob/49070cedd884bffb7e85b99948bb4ab149a69b9e/tests/test_create.py

```
{
                            "_path": sp_dir + "/" + "__pycache__/six.cpython-35.pyc",
                            "path_type": "hardlink",
                            "sha256": None,
                            "size_in_bytes": None
                        }
```

nehaljwani:
Thanks.

PR #235, opened 2018-11-10
Add logic to reset AutoRun reg keys on Uninstallation

nehaljwani:
This will be required starting conda 4.6.x

xref: https://github.com/conda/conda/issues/7594
xref: https://github.com/conda/conda/issues/7881


msarahan:
Merging this.  We need to add a test item for QA to verify that this works.  I will take care of that.

PR #238, opened 2018-12-04
Resolve slow terminal issue

soapy1:
The problem seems to be related to how __conda_setup is invoked.
It is resolved by using $HOME in the prefix opposed to using e.g.
/Users/<user>. This should be resolved without this fix when conda
4.6 is released

ref:
AI-69
ContinuumIO/anaconda-issues#10173

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -511,20 +511,21 @@ if [ "$BATCH" = "0" ]; then
             printf "\\n"
             printf "Initializing __NAME__ in newly created %s\\n" "$BASH_RC"
         fi
+        PROFILE_PREFIX="${PREFIX/$HOME/'~'}"
         cat <<EOF >> "$BASH_RC"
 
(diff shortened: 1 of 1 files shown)


soapy1:
In ubuntu this speeds up from 
```
$ time source ~/.bashrc

real	0m0.883s
user	0m0.479s
sys	0m0.146s
```
to
```
$ time source ~/.bashrc

real	0m0.061s
user	0m0.014s
sys	0m0.023s
```

On mac this speed up from
```
$ time bash .bash_profile
bash .bash_profile  0.54s user 0.37s system 80% cpu 1.127 total
```
to
```
$ time bash .bash_profile
bash .bash_profile  0.00s user 0.00s system 76% cpu 0.009 total
```

Since this only updates the path when $HOME matches the prefix, this will only work for when the user installs this for themselves. Installations into other prefix's should not be impacted.
Looking into other side effects of this, and why this it faster than using $HOME

PR #239, opened 2019-01-07
rewrite for bundling envs instead of pkgs

msarahan:
This is something of an experimental rewrite.  It dramatically reduces the amount of work that constructor does, deferring instead to conda for env creation.  This library then packages the env that conda created.  It may package envs that already exist, too - so pip and other such stuff is possible in ways that didn't exist before.

There's a few concerns with this approach that need to be addressed before we can really consider switching to it:

1. we lose the packages by doing this method.  The package cache would need to be re-created or sacrificed.
2. by sacrificing the packages, we also lose the ability to test installers.  This one will be harder to re-create.  We'd probably need to re-download packages.
3. This rewrite depends on a standalone executable for unpacking the created env archive.  I've tested pyinstaller with some success, but it has also given me some confusing stuff that makes me think it's not packaging libarchive totally successfully.  It may be better or outright necessary to have a nicer standalone implementation of the install.py file.


dakcarto:
Very interesting approach. I couldn't tell immediately from the code changes, but wouldn't this mean the `post-link` scripts need to be idempotent, i.e. wouldn't they have be already run on the initial env setup?

While I suppose skipping `post-link` scripts could be part of an automated installer generation, where the env creation could be controlled, this would not work for an existing starting env, where the `post-link` script(s) would have presumably been already run.

---

Would another efficient approach be...

- construct installer
  - export an existing env to an env.yaml file
  - cache (mirror) the pkgs in env.yaml to temp local conda channel and pip cache, including no-arch, based upon listed channels
  - roll env.yaml and local channel and cache into installer payload

- on target install
  - install `base` env and use (or python and install scripts) to recreate the env, based off temp channel consisting of payload pkgs and pip cache
  - run `post-link` scripts, if not already run
  - run `menuinst` setup (on Win) like already done with constructor installers

This would essentially end up like installing Miniconda then importing the custom env, but via the user-constructed installer, with control over `constructor` config'd things like `menuinst` for cetain pkgs, etc.

Obviously, this creates a child env, though it seems the current install scripts could be adapted to handle env creation, if resulting only in a `base` env (like now) is the goal.

The difference between this approach and yours here is that it still works with pkgs, instead of a copying the env prefix as-is. The latter seems to remove the ability to apply `constructor`-config'd operations on a pre-pkg basis.

**Side note:** I started looking into the approach outlined above as a means of using `constructor` with a [Qt Installer Framework](https://doc.qt.io/qtinstallerframework/ifw-overview.html) install backend. Whereby, the payload would be a converted `conda` channel --> [QtIFW package directory](https://doc.qt.io/qtinstallerframework/ifw-tutorial.html). This would allow the same installer for all platforms and possibly use `meta` packages as a way to indicate 'optional' component installs in the installer's GUI.

dakcarto:
Followup to my QtIFW backend suggestion...

Using your per-env-prefix payload approach (or maybe a variation of my suggested approach), the optional installs in the QtIFW GUI could be _envs_. So, instead of converting a conda channel to a QtIFW package directory for a given env, the QtIFW package directory would be a collection of envs, where any could be optional, i.e. a _multi-env_ installer.

msarahan:
@nehaljwani implemented the standalone executable with #263 and other work to be added soon, and constructor is now capable of doing this idea.

msarahan:
PS: standalone executables are at https://repo.anaconda.com/pkgs/misc/conda-execs/

mcg1969:
I would like to close this because it just doesn't seem likely we'll get this updated to work with the existing code base. And the use of the standalone conda.exe has proven reasonably robust.

PR #240, opened 2019-01-21
Use repodata information to put into conda-meta

soapy1:


soapy1:
This PR is meant to address the constructor side of the issue from https://github.com/conda/conda/issues/8076

This change should put hotfixed metadata into conda meta when constructor installs all of the requested packages opposed to unhotfixed metadata sourced from the `index.json` of the target package.

msarahan:
Can we add a test for this?  We need a way to verify that this works somehow, to preserve the functionality for the future.

Issue #241, opened 2019-02-10
post_install fails with installer_type: pkg on OSX


newville:
This is related to #220.

I would like to be able to use a `post_install` script when making on OSK pkg installer.  FWIW, I'm using `post_install` script to run a freshly installed python program to create a desktop App using https://github.com/newville/pyshortcuts  This works well for all other `constructor` products (Unix shell, Windows installer), but not for OSX PKG.

It appears that a user-specified `post_install` script listed in `construct.yaml` is simply not included in any of the "packages" used by `pkgbuild`, and so is not run at the end of the installation.   

An example  showing the problem would have `construct.yaml` of

```
name: osxpkgtest
version: 1

channels:
    - http://repo.anaconda.com/pkgs/main/

attempt_hardlinks: True

specs:
    - python
    - conda
    - openssl
    - asn1crypto
    - readline
    - setuptools
    - numpy

post_install: post_install.sh   [unix]
installer_type: pkg         [osx]
```

and `post_install.sh`  of
```
#!/bin/bash
echo '## Hello from Post_install script ' > $HOME/postinstall.txt
printenv >> $HOME/postinstall.txt
```

This will create a pkg installer and that installer will run to completeion, but the user-supplied `post_install` script will not be included or run.

I believe there should be another "package" built to run the users `post_install` script.   I will propose a change in `osxpkg.py`

nehaljwani:
Happy to look at a PR, if you have one!

mcg1969:
Fixed, by you! in #242. Thank you!

PR #242, opened 2019-02-10
support for `post_install` in OSX Packages 

newville:
This is an attempt to address #241.  Here, a final "package" is built with `pkgbuild` to run a user-supplied `post_install` script, if provided.

This approach works for me (a simple example is added) and seems 'simple enough', but I do not know if this is the preferred approach.

There is at least one outstanding issue:   In the post install script with Unix shell or Windows, the environment variable `PREFIX` is set correctly.   But `PREFIX` is not set here, and one would have to the sort of magic done at the top of `update_path.sh`, namely 

```
# $2 is the install location, which is ~ by default, but which the user can
# change.
PREFIX=$(echo "$2/__NAME_LOWER__" | sed -e 's,//,/,g')
```

I'm not sure if  a replacement as is done for `__NAME_LOWER___` from inside the `move_script()` function could be done directly for `$PREFIX`.   The  `__NAME_LOWER___` trick would be an acceptable workaround, but it would be sort of unfortunate to need that for Mac OSX packages, and not any other  `post_install` scripts.

### DIFF
--- a/constructor/osxpkg.py
+++ b/constructor/osxpkg.py
@@ -254,6 +254,17 @@ def create(info, verbose=False):
     # can be disabled.
     pkgbuild_script('pathupdate', info, 'update_path.sh')
 
+    post_packages = ['postextract', 'pathupdate']
+
+    # Next, the users post_install script, if specified
+    if info.get('post_install', None) is not None:
+        scripts_dir = join(CACHE_DIR, "scripts")
+        fresh_dir(scripts_dir)
+        move_script(info['post_install'], join(scripts_dir, 'postinstall'), info)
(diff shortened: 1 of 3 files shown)


newville:
The CI failures seem unrelated to this PR.

newville:
@nehaljwani @csoja any chance of reviewing and/or commenting on this?   It does not seem to be that big of a change to me.

nehaljwani:
Apologies for the delay. Thank you for your contribution, @newville !

newville:
@nehaljwani Great! Thanks very much -- this will be really helpful for one of my projects.

PR #245, opened 2019-03-01
Fast uninstall

soapy1:


constructor/nsis/_nsis.py
```diff
@@ -229,6 +230,18 @@ def rm_regkeys():
             pass
 
 
+def win_del(dirname):
+    try:
+        out = check_output('DEL /F/Q/S {} > NUL'.format(dirname), shell=True, stderr=STDOUT, cwd=dirname)
+    except CalledProcessError as e:
+            if e.returncode != 5:
+                print("Removing folder {} the fast way failed. Output was: {}".format(dirname))
```
wulmer: Hi, isn't there an argument missing?
```suggestion
                print("Removing folder {} the fast way failed. Output was: {}".format(dirname, out))
```

constructor/nsis/_nsis.py
```diff
@@ -229,6 +230,18 @@ def rm_regkeys():
...
+                raise
+            else:
+                print("removing dir contents the fast way failed. Output was: {}".format(dirname))
```
wulmer: Isn't the argument `dirname` wrong here?
```suggestion
                print("removing dir contents the fast way failed. Output was: {}".format(out))
```

constructor/nsis/_nsis.py
```diff
@@ -229,6 +230,18 @@ def rm_regkeys():
             pass
 
 
+def win_del(dirname):
+    try:
+        out = check_output('DEL /F/Q/S {} > NUL'.format(dirname), shell=True, stderr=STDOUT, cwd=dirname)
```
wulmer: Wouldn't it be easier to provide the arguments for `DEL` as list? For example:
```suggestion
        out = check_output(['DEL', '/F', '/Q', '/S', dirname], shell=True, stdout=subprocess.DEVNULL, stderr=STDOUT, cwd=dirname)
```

msarahan: You may know more about this than me, so take this with a grain of salt.  My experience is that for things that are shell functions, not actual programs (which is what DEL is here), you must use shell=True, and you must make a single string, not a list of strings.

msarahan: DEL operates on files, not on folders.  I think you need to make this just *.*, and rely on the cwd argument to put you in the right place.

constructor/nsis/_nsis.py
```diff
@@ -229,6 +230,18 @@ def rm_regkeys():
...
+    except CalledProcessError as e:
+            if e.returncode != 5:
```
msarahan: Probably worth a comment here to indicate that error code 5 indicates a permission error.  We ignore those, but raise for anything else.

msarahan:
@wulmer thanks for the review here!  Do you have any comment on the general strategy?  We have lots of people complaining about the uninstall speed, and we believe it is due to GUI updates of the uninstall window.  We think this subprocess approach should be much faster (though it sacrifices any visual feedback).

soapy1:
Looks like the tests on appveyor are failing on the readline package not being available. @msarahan where does this package come from?

msarahan:
looks like it might be a bad test: https://ci.appveyor.com/project/ContinuumAnalyticsFOSS/constructor/builds/22910420/job/7ar3xhn55v165ira#L440

If we're trying to build osx installers on windows, it's probably not going to work.

nehaljwani:
Add a selector to readline or just drop it from the example. It's presence in the example is irrelevant to the test.

wulmer:
> @wulmer thanks for the review here! Do you have any comment on the general strategy? We have lots of people complaining about the uninstall speed, and we believe it is due to GUI updates of the uninstall window. We think this subprocess approach should be much faster (though it sacrifices any visual feedback).

Definitely the right choice. I had a similar issue some years ago with NSIS where the detail view was first hidden by default, but when you opened it you couldn't close it but the GUI updates were extremely slowing down uninstall progress.

constructor/nsis/_nsis.py
wulmer: After your comment I wasn't sure any more, but I just tested it and if `shell=True`, you can split the arguments to shell commands like `dir` or `del`. So something like `check_output(['del', some_dir], shell=True)` should work.
The solution with a single string should also work, however I always fear issues with broken quoting of arguments with whitespace, so I always prefer argument lists instead of strings.

PR #246, opened 2019-03-08
Fix "conda init" section for shell configs

sscherfke:
The *conda init* section that `conda init` generates in, e.g., `.bashrc` looks different than what Constructor generates. That means that `conda init` will basically add the same stuff again if you run it.

With these changes, `conda init` will no longer modify the config that Constructor generated (or add something to it).

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -514,21 +514,20 @@ if [ "$BATCH" = "0" ]; then
         cat <<EOF >> "$BASH_RC"
 
 # added by __NAME__ __VERSION__ installer
-# >>> conda init >>>
+# >>> conda initialize >>>
 # !! Contents within this block are managed by 'conda init' !!
-__conda_setup="\$(CONDA_REPORT_ERRORS=false '$PREFIX/bin/conda' shell.bash hook 2> /dev/null)"
+__conda_setup="\$('$PREFIX/bin/conda' 'shell.bash' 'hook' 2> /dev/null)"
 if [ \$? -eq 0 ]; then
-    \\eval "\$__conda_setup"
+    eval "\$__conda_setup"
 else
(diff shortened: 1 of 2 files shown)


sscherfke:
Btw, what is the difference between
```bash
__conda_setup="\$('$PREFIX/bin/conda' 'shell.bash' 'hook' 2> /dev/null)"
if [ \$? -eq 0 ]; then
    eval "\$__conda_setup"
fi
unset __conda_setup
```
and 
```bash
if [ -f "$PREFIX/etc/profile.d/conda.sh" ]; then 
    . "$PREFIX/etc/profile.d/conda.sh"           
fi
```
?

The output of `conda shell.bash hook` is exactly the same as the contents of `conda.sh`.  Wouldn’t one of the two variants suffice?

kalefranz:
They’re the same right now, but looking toward the future they might not always be. The contents of the `conda.sh` file are POSIX-compliant. The hook command gives us the opportunity to add shell-specific code (like shell completions).  It’s also a command interface that’s reasonably consistent across all shells.

The hook pattern was modeled after https://direnv.net.

kalefranz:
Actually, given that we have so much of this in the wild now, I think the right fix would be to the regex in `conda init`. It should accept both of the partition markers.

sscherfke:
@kalefranz But that would have to happen in Conda, right?

Could you still merge this PR as a “workaround”, then? :-)

mcg1969:
@sscherfke I'm sorry this has remained open for so long but in fact I believe it has been fixed in another way. Would you agree? If so we can close this.

sscherfke:
I guess so. 🤷

mcg1969:
Let's do that then. But please don't hesitate to open a new issue and/or re-open the PR with explanation if this isn't the case.

PR #274, opened 2019-08-22
Updated note in README about noarch support

jlstevens:
Just a quick PR to update the README now that noarch packages are supported!


jjhelmus:
LGTM, thanks for the update.

Issue #269, opened 2019-08-13
Support local file:///-based channels


dsentinel:
Using constructor to build an installer with a locally built and hosted package like this:
```yaml
...
channels:
  - file:///Users/builder/miniconda3/conda-bld
  - https://conda.anaconda.org/conda-forge
  - http://repo.anaconda.com/pkgs/main/
...
```
No longer works.

On install, the local channel is not found as it is a path on the machine the constructor was created.
```
$ bash pkg-2019.219-MacOSX-x86_64.sh

Welcome to pkg 2019.219

pkg will now be installed into this location:
/Users/installer/pkg_env

  - Press ENTER to confirm the location
  - Press CTRL-C to abort the installation
  - Or specify a different location below

[/Users/installer/pkg_env] >>>
PREFIX=/Users/installer/pkg_env
Unpacking payload ...
Collecting package metadata (current_repodata.json): ...working... failed

UnavailableInvalidChannel: The channel is not accessible or is invalid.
  channel name: conda-bld
  channel url: file:///Users/builder/miniconda3/conda-bld
  error code: 404

You will need to adjust your conda configuration to proceed.
Use `conda config --show channels` to view your configuration's current state,
and use `conda config --show-sources` to view config file locations.

As of conda 4.3, a valid channel must contain a `noarch/repodata.json` and
associated `noarch/repodata.json.bz2` file, even if `noarch/repodata.json` is
empty. Use `conda index /Users/lloyd/miniconda3/conda-bld`, or create `noarch/repodata.json`
and associated `noarch/repodata.json.bz2`.
```

conda 4.7.11
constructor 3.0.0

astaric:
I have the same problem. Opened #276 with the change I use, soI can create the installer.

astaric:
@nehaljwani, in my case, there is no place to remap the local channel to. I build installer from local packages since these packages are not available anywhere else.

msarahan:
Reopening.  We are encountering this in building Anaconda installers, where the anaconda metapackage is not available on any public channel yet.  I think the best way to do it is probably to create the $PREFIX/conda-bld folder, populate it with these local file URLs, and copy the files in there.  We'll have to pre-generate the index in there as well, so that conda-build is not a requirement for each installer, or alternatively make a standalone executable that includes conda-build.

scottwsides:
[...]

PR #276, opened 2019-08-28
Do include local channels in installer.

astaric:
Do not include local channels in installer file. Conda will not be able to find them on the target machine.

Attempt to fix https://github.com/conda/constructor/issues/269

### DIFF
--- a/constructor/utils.py
+++ b/constructor/utils.py
@@ -133,5 +133,10 @@ def get_final_url(info, url):
 def get_final_channels(info):
     mapped_channels = []
     for channel in info.get('channels', []):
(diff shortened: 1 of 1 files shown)


astaric:
@nehaljwani, the check is done on the final url (as returned from the get_final_url function), so if a remap exists, it has already been applied.

I have added a warning that will be displayed when building installer if a local channel does not define a remap. I used print to show the warning, as this is how warnings are displayed in get_final_url function.

dsentinel:
Tested this for my case with a local channel. Worked!
Thanks @astaric & @nehaljwani !

PR #277, opened 2019-09-03
Not delete $INSTDIR while Python is still running

wulmer:
On Windows, it may occur that a process is locking access
to loaded DLLs or its EXE file. If the uninstaller tries to
call Python for deleting the Anaconda installation directory,
Python will prevent itself from deleting its own executable.

Therefore, NSIS should perform the deletion, not Python itself.
(which is done anyways in the next line)

### DIFF
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -876,7 +876,6 @@ Section "Uninstall"
 
     ExecWait '"$INSTDIR\pythonw.exe" -E -s "$INSTDIR\Lib\_nsis.py" rmpath'
     ExecWait '"$INSTDIR\pythonw.exe" -E -s "$INSTDIR\Lib\_nsis.py" rmreg'
-    ExecWait '"$INSTDIR\pythonw.exe" -E -s "$INSTDIR\Lib\_nsis.py" del "$INSTDIR"'
 
     RMDir /r /REBOOTOK "$INSTDIR"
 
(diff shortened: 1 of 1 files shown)


msarahan:
That deletion is meant as a first pass.  It shells out to the fastest way we know of to clean up files.  For ones that deletion fails on, it should leave the files in place.  It may rename them with a .conda_trash extension or something like that.

Are you seeing errors that indicate this failure handling isn't working properly?

mcg1969:
@wulmer I just want to make sure you've had a chance to consider the merit of the PR in light of Mike's comments above. If I'm reading him correctly, at least _attempting_ to delete the files using `_nsis.py` is the fastest approach. Certainly the RMDir command could do it but it would be slower. Do you have reason to believe his analysis is incorrect or introduces additional failure modes?

wulmer:
@mcg1969 Thanks for taking this up... sorry that I didn't respond for so long.

My PR was actually a bit half-baked. It is true (like Mike wrote) that `_nsis.py`'s deletion is much faster and should be preferred, but my concern (and my observation for one of my installers) was that if the call to `_nsis.py` fails because there are some files currently being used, the `ExecWait` will set the error flag of the uninstaller.
So, IMHO there should be some kind of error handling/tolerance instead of a simple `ExecWait`.

The best solution in my opinion should ignore the exit status of that particular `ExecWait` call. As we anyways clean up afterwards, the exit code is not relevant here, but if this command fails, the uninstall process fails.

I'll try to find some time today to check this again with the newest constructor code base.

wulmer:
OK, now I remember what was my original concern:

The uninstaller worked correctly, even if there were locked files. But in that case it showed a stack trace dialog from the `pythonw _nsis.py` call which blocks the application during silent mode.

When running the uninstaller non-interactively, this dialog window blocks the uninstallation process and leads to half-uninstalled application.

To prevent merging and a useless commit history, I'll create a new PR with my changes.

wulmer:
The new PR is https://github.com/conda/constructor/pull/345

PR #278, opened 2019-09-09
Standalone exec

msarahan:


msarahan:
standalone exec moved to own repo: https://github.com/AnacondaRecipes/conda-standalone-feedstock

PR #279, opened 2019-09-16
use dd to unpack shar instead of tail -c

### LINKED ISSUE #272: Miniconda breaks for old versions (or an old version?) of `tail`
Based on <https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh> at `v4.7.10`.

On an old Linux that has `tail` from GNU coreutils 8.12 / April 2015.

The file contains 
```shell
[...]
printf "Unpacking payload ...\n"
if ! tail -c +000000000000018891 "$THIS_PATH" | tail -c +9231072 | tail -c +2 | "$CONDA_EXEC" constructor --extract-tar --prefix "$PREFIX"; then
    printf "ERROR: could not extract tar starting at offset 000000000000018891+9231072+2\\n" >&2
    exit 1
fi
[...]
```
Here, the `tail`s later in the pipe are supposed to read from stdin but have no explicit `-`.  For more recent GNU coreutils, `tail -c` seems to default to stdin:
```bash
$ man -P cat tail | tail -n1
GNU coreutils 8.25                         February 2017                                    TAIL(1)
$ seq 1 9 | tail -c 2
9
```

For GNU coreutils 8.12, this is not true:
```bash
$ man -P cat 1 tail | tail -n 1
GNU coreutils 8.12                April 2015                           TAIL(1)
$ seq 1 9 | tail -c 2
tail: cannot open `2' for reading: No such file or directory
```

msarahan:
fixes #272

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -3,7 +3,6 @@
 # NAME:  __NAME__
 # VER:   __VERSION__
 # PLAT:  __PLAT__
-# BYTES: @SIZE_BYTES@
 # LINES: @LINES@
 # MD5:   __MD5__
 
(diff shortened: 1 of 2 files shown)


mingwandroid:
Does `dd` come with every linux distro?

msarahan:
it's part of coreutils, same as head & tail. This is not a new requirement at least.

PR #285, opened 2019-09-19
Write to tempdir in install dir if normal tempdir is mounted as noexec

soapy1:


nehaljwani:
Is adding a dependency on psutil really necessary? Why not just create a small file with execute permissions and subprocess it and try/catch it?

nehaljwani:
Also, we need to take care of the location where the self contained exec gets extracted and where it extracts its own contents.

msarahan:
@nehaljwani those are going into the install prefix.  Is that location possibly problematic in any way?

soapy1:
In order to avoid introducing psutil or writing out own check for a noexec mount, how about setting up the tempdir in the alternate location whenever the temporary directory is a mount?
eg. `if os.ismount(tempfile.gettempdir()`

msarahan:
I think if you do that, you might as well just always use the alternative tempdir.  I think the purpose of mounting it would be to put it on a faster volume (RAMdisk, or such).  If you're always avoiding that, you might as well just make the logic uniform.  I'm fine with that.

PR #295, opened 2019-10-02
trim repodata cache to include only pkgs included with installer

msarahan:
Addresses an issue reported by QA where conda would try to use packages from repodata.json that were not available when using --offline flag.  This limits the initial install with --offline to only exactly what is available, but future requests will obtain new repodata for the rest of the packages.

This may need a tiny bit more work to force the repodata to be refreshed when it is first used without the --offline flag.

### DIFF
--- a/constructor/conda_interface.py
+++ b/constructor/conda_interface.py
@@ -81,7 +81,7 @@ if conda_interface_type == 'conda':
         used_repodata['_mod'] = "Mon, 07 Jan 2019 15:22:15 GMT"
         for package in used_packages:
             for key in ('packages', 'packages.conda'):
-                if package in full_repodata[key]:
+                if package in full_repodata.get(key, {}):
                     used_repodata[key][package] = full_repodata[key][package]
         with open(join(cache_dir, repodata_filename), 'w') as fh:
             json.dump(used_repodata, fh, indent=2)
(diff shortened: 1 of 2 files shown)


msarahan:
"Works for me" - @forrestwaters please review

constructor/conda_interface.py
```diff
@@ -52,26 +53,36 @@
...
+                                                                                       'removed',))}
+        used_repodata['packages'] = {}
+        used_repodata['packages.conda'] = {}
+        used_repodata['removed'] = []
+        # arbitrary old, expired date, so that conda will want to immediately update it
+        # when not being run in offline mode
+        used_repodata['_mod'] = "Mon, 07 Jan 2019 15:22:15 GMT"
+        for package in used_packages:
+            for key in ('packages', 'packages.conda'):
+                if package in full_repodata[key]:
```
forrestwaters: It looks like this will break if any channel is used that doesn't have the `packages.conda` key which happens with the r channel, etc.

msarahan: ah, good catch.  That should be a .get().

forrestwaters:
I'm testing Anaconda installers with this and am hitting unsatisfiable errors I don't quite understand yet. It works fine with the 4.7.5 standalone but not 4.7.12

PR #296, opened 2019-10-04
Fix missing navigator menus

forrestwaters:
This adds the packages that navigator usually displays to our trimmed repodata cache for the installers.


forrestwaters:
appveyor failures are because of the windows standalone we're working on fixing.

PR #297, opened 2019-10-08
Fix handling of bools in construct.yaml

forrestwaters:


forrestwaters:
You can work around this with the following in your yaml file: 

```
  version: !!str "2019.10"
```

PR #292, opened 2019-09-30
Add no scripts option for windows installers

ericpre:
Add no scripts option to windows installers. The can be used when installing in silent mode:
```
Installer_name_exe /NoScripts=1 /S /D=installation_director
```
Or in through the gui by checking the "Do not run post install script.", which will be displayed if the `post_install_desc` key is provided in `construct.yaml`.

PR #306, opened 2019-10-31
Get channels urls using conda

ericpre:
Revisit #231 to allow using channels name in `construct.yaml`, for example:

```
channels:
  - defaults
  - conda-forge
```

### DIFF
--- a/constructor/conda_interface.py
+++ b/constructor/conda_interface.py
@@ -31,6 +31,7 @@ if conda_interface_type == 'conda':
     from conda.core.prefix_data import PrefixData as _PrefixData
     from conda.core.solve import Solver as _Solver
     from conda.exports import default_prefix as _default_prefix
+    from conda.models.channel import all_channel_urls as _all_channel_urls
     from conda.gateways.disk.read import read_paths_json as _read_paths_json
     from conda.models.dist import Dist as _Dist
     from conda.exports import MatchSpec as _MatchSpec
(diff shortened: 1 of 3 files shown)


ericpre:
Is there a plan for considering this PR? If so, having an idea of when it would be reviewed would be useful.
I will solve the conflict when someone get time to review it and give feedback on the approach, to avoid spending time on it for nothing.

Same for #292 and ##288.

msarahan:
sorry, constructor typically gets neglected in between Anaconda releases.  We will begin working on the next Anaconda release early in January, so I think you can expect to see more attention then if we don't get to it sooner.

ericpre:
OK, thanks for the update on the timing.

forrestwaters:
@ericpre thanks for the contributions and apologies for the poor response times. Looks like this one breaks the [channels_remap functionality](https://github.com/ericpre/constructor/blob/channels_urls/CONSTRUCT.md#channels_remap) which we unfortunately don't have tests for. 
I think we need to remove these trailing slashes: https://github.com/conda/constructor/blob/0d4df820864465cfc2de2e73c75b01d39e5d4d63/constructor/preconda.py#L47
The other CI test failure is unrelated and is something we will need to fix. We will be starting on the next release in the next few weeks and will be sure this and your other PRs get reviewed properly. Thanks again!

ericpre:
Yes, indeed, some trailing slashes needed to be removed. When solving the merge conflict, I checked it for [remapping from a folder to an url](https://github.com/conda/constructor/pull/306/files#diff-10afcf5d28ad04ccb09235831e8ccc11R61,) but it was still breaking remapping from one url to another.

I have tried with:
```
name: MinicondaX
version: X

channels:
  - defaults

channels_remap:
  - src: file:///opt/miniconda3/conda-bld
    dest: https://repo.anaconda.com/pkgs/main

  - src: https://repo.anaconda.com/pkgs/main/
    dest: https://repo.anaconda.com/pkgs/main/

specs:
  - python 3.7*
  - conda
```
and it seems to work.

forrestwaters:
Thanks!

PR #307, opened 2019-11-01
force prefix if passed as kwarg so standalone doesnt fallback to.…

forrestwaters:
…sys.prefix

### DIFF
--- a/constructor/nsis/_nsis.py
+++ b/constructor/nsis/_nsis.py
@@ -143,7 +143,10 @@ def get_conda_envs_from_python_api():
 get_conda_envs = get_conda_envs_from_python_api
 
 
-def rm_menus():
+def rm_menus(prefix=None):
+    from conda.base.context import context
+    if prefix is not None:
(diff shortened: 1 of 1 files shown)


msarahan:
We may want to think about a way for people to update their standalone. As it is, this will not fix old installations, only new ones created by installers that have this fix (which obviously won't be available until the next Anaconda and miniconda releases).

mcg1969:
Note: `normpath` does not reliably remove all duplicate slashes. Sophia's `normalize_path` does, but it doesn't work on Windows. However, that can be fixed; see, for example, #315

jjhelmus:
I think  6afc3d771f5160b9caaa07b3f9ca9b5b49a61714 and 6ffc7c2f1464c9ba5a3ec29015a9d893f5c91e33 can be dropped given the changes to `normalize_path` in #315

forrestwaters:
Thanks, I removed those changes and rebased

jjhelmus:
Thanks @forrestwaters

Issue #261, opened 2019-08-06
Allow channels to be empty if channel_urls is not


mcg1969:
https://github.com/conda/constructor/blob/7e4d2f455c695227fdc9a25a75e8841917bf8862/constructor/fcp.py#L232-L233

Since the channels in the `src` entries of `channels_remap` are included in the source channel list, there should be no need to include them in `channels` as well. But if you are remapping _all_ of your channels, you exercise the error above if you don't also include at least one of those channels in the `channels` list.

jjhelmus:
fixed in #315

PR #315, opened 2019-11-22
channel_remap fixes: unavailable destination, .conda file format

mcg1969:
There seems to have been a regression in behavior with regards to `channel_remap`. Specifically, if any of the destination channels are not in the `channels` list, you get KeyErrors, like so:
```
Traceback (most recent call last):
  File "C:\Users\builder\Miniconda3\Scripts\constructor-script.py", line 11, in <module>
    load_entry_point('constructor', 'console_scripts', 'constructor')()
  File "c:\users\builder\desktop\constructor\constructor\main.py", line 244, in main
    dry_run=args.dry_run, conda_exe=args.conda_exe)
  File "c:\users\builder\desktop\constructor\constructor\main.py", line 134, in main_build
    create(info, verbose=verbose)
  File "c:\users\builder\desktop\constructor\constructor\winexe.py", line 192, in create
    preconda_write_files(info, tmp_dir)
  File "c:\users\builder\desktop\constructor\constructor\preconda.py", line 102, in write_files
    write_index_cache(info, dst_dir, info['_dists'])
  File "c:\users\builder\desktop\constructor\constructor\preconda.py", line 58, in write_index_cache
    write_repodata(cache_dir, url, repodata, used_packages)
  File "c:\users\builder\desktop\constructor\constructor\conda_interface.py", line 78, in write_repodata
    repodata_filename = _cache_fn_url(used_repodata['_url'].rstrip("/"))
KeyError: '_url'
```
It is not feasible to expect the destination channels to also be listed in the `channels` list, because those channels might not be accessible from the location where we are building the constructor. Consider the case where an installer is being built outside of a company firewall with the intention of being used inside that firewall, connected to an internal repository. On the other hand, it _is_ reasonable to expect the source channels to be in the channel list.

### DIFF
--- a/constructor/fcp.py
+++ b/constructor/fcp.py
@@ -232,8 +232,8 @@ def main(info, verbose=True, dry_run=False):
     install_in_dependency_order = info.get("install_in_dependency_order", True)
     ignore_duplicate_files = info.get("ignore_duplicate_files", False)
 
-    if not channel_urls:
-        sys.exit("Error: 'channels' is required")
+    if not channel_urls and not channels_remap:
+        sys.exit("Error: at least one entry in 'channels' or 'channels_remap' is required")
 
     with env_vars({
         "CONDA_PKGS_DIRS": download_dir,
--- a/constructor/preconda.py
+++ b/constructor/preconda.py
@@ -40,23 +40,26 @@ def write_index_cache(info, dst_dir, used_packages):
 
     package_urls = dict(info['_urls'])
 
-    remap_urls = []
-    for subdir in _platforms:
-        for url in info.get('channels_remap', []):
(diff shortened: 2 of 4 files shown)


mcg1969:
Found more issues. Specifically, packages with the `.conda` format were not being remapped, because the repodata uses the `.tar.bz2` filename for the file key. Furthermore, it was looping through all of the source repodata, instead of the much shorter list of packages.

Also, `normalize_path` was not coded properly for Windows.

mcg1969:
I didn't have the `.conda` remap support correct before; I do now. Also fixed #261 as a bonus

jjhelmus:
LGTM, thanks for the fix @mcg1969

PR #324, opened 2020-02-04
docs around construct.yaml

msarahan:


forrestwaters:
👍 this all looks good to me. I think there may be one or two more keys that are no longer relevant (I don't think attempt_hardlinks does anything anymore) but I can clean that up in a future PR.

PR #334, opened 2020-04-06
ENH: use current shell instead of YOUR_SHELL_NAME

yarikoptic:
This should make it possible to just cut/paste the line.
Should work only only POSIXy shells, probably would not be good
for tcsh etc.

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -494,7 +494,7 @@ if [ "$BATCH" = "0" ]; then
         printf "You have chosen to not have conda modify your shell scripts at all.\\n"
         printf "To activate conda's base environment in your current shell session:\\n"
         printf "\\n"
-        printf "eval \"\$($PREFIX/bin/conda shell.YOUR_SHELL_NAME hook)\" \\n"
+        printf "eval \"\$($PREFIX/bin/conda shell.${SHELL##*/} hook)\" \\n"
         printf "\\n"
         printf "To install conda's shell functions for easier access, first activate, then:\\n"
         printf "\\n"


yarikoptic:
will deal with SLA if PR is told to be worthwhile consideration ;)

mcg1969:
I appreciate what you're trying to do here but I actually don't think this is correct. After all, `conda` is compatible with a variety of shells that are _not_ able to run the constructor itself, like `fish` and `dash`.

yarikoptic:
`##` is an operation supported by any posix compliant shell, see eg https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html for the description of that function.
dash is posix compliant - it will work.

 fish seems indeed not according to https://stackoverflow.com/questions/48732986/why-how-fish-does-not-support-posix and I found no ## in https://fishshell.com/docs/current/#expand-variable . But then - are you sure that this .sh is fully compatible with fish?

mcg1969:
I'm afraid I've not made myself clear—the point is that the instructions as written work with _any_ shell, even the ones that are not compatible with the installer itself. So for example, if you use `bash` to run the installer, but `fish` is your default shell, the command you really want to use is `$PREFIX/bin/conda shell.fish hook`.

yarikoptic:
and they still would if user specifies explicitly which shell, but `YOUR_SHELL_NAME` could even be an unknown thing to many naive new users, thus this PR was changing so they do not just cut/paste and get an error.  But I guess it would indeed be a bit too suggestive, thus closing.  But you might then want to change to `DESIRED_SHELL` or smth like that.

PR #341, opened 2020-05-18
FileNotFoundError

fbriol:
Some packages contain links that are not present in the archive. I work around the problem by ignoring inaccessible files. Maybe I need to do some more checking.

### DIFF
--- a/constructor/fcp.py
+++ b/constructor/fcp.py
@@ -142,6 +142,8 @@ def check_duplicates_files(pc_recs, platform, ignore_duplicate_files=False):
                         getsize(join(extracted_package_dir, short_path)))
             except AttributeError:
                 size = getsize(join(extracted_package_dir, short_path))
+            except FileNotFoundError:
+                continue
             total_extracted_pkgs_size += size
 
             map_members_scase[short_path].add(fn)


mcg1969:
I think a little more work needs to be done here. First of all—can you give us an example of a package that contains a link that is not in the archive itself? I'm just curious what the practical use case is for that. That said I don't need proof of that to proceed here.

So let's assume for the sake of argument that we need to handle this use case. This implementation still has a problem, because it is bypassing the duplicate file checks. In fact it's quite possible the current implementation is flawed w.r.t. symbolic links too.

Here's the problem: because we are using `os.path.getsize` in this loop, it is not going to return the correct size for symbolic links, because it returns the size of the _target_. And so if, say, 10 symbolic links point to the same file, the total size will by off by a factor of 10 or so.

But we still need to fully enumerate and account for all files and symbolic links in our packages, so that two objects with the same name & path are not found across the packages. So we can't just abort the loop when `getsize` fails. What we have to do instead is properly size symbolic links by using something other than `getsize`.

fbriol:
I was a little hasty in my explanations. I'm sorry.

The initial problem is probably not in this software, but in the gsl 2.6 package. The archive contains the following files:

```bash
ls -l /home/fbriol/.conda/constructor/linux-64/gsl-2.6-h294904e_0/lib/
total 11080
-rw-rw-r-- 1 fbriol fbriol 7010566 Sep 3 2019 libgsl.a
-rw-rw-r-- 1 fbriol fbriol 548418 Sep 3 2019 libgslcblas.a
lrwxrwxrwx 1 fbriol fbriol 13 Sep 3 2019 libgslcblas.so -> libcblas.so.3
lrwxrwxrwx 1 fbriol fbriol 13 Sep 3 2019 libgslcblas.so.0 -> libcblas.so.3
-rwxrwxr-x 1 fbriol fbriol 304552 Sep 3 2019 libgslcblas.so.0.0
lrwxrwxrwx 1 fbriol fbriol 16 Sep 3 2019 libgsl.so -> libgsl.so.25.0.0
lrwxrwxrwx 1 fbriol fbriol 16 Sep 3 2019 libgsl.so.25 -> libgsl.so.25.0.0
-rwxrwxr-x 1 fbriol fbriol 3485120 Sep 3 2019 libgsl.so.25.0.0
drwxrwsr-x 2 fbriol fbriol 4096 Jun 1 06:13 pkgconfig
```
The symbolic links to the dynamic library ``libgslcblas`` are broken. The ``libcblas.so.3`` file does not exist.

If you create the following minimum recipe: 

```yaml
name: dummy
version: 1.0.0

ignore_duplicate_files: True

channels:
  - https://repo.anaconda.com/pkgs/main/
  - https://conda.anaconda.org/conda-forge/

specs:
  - gsl=2.6
  - openblas
```

The construction comes out in error because the program tries to read the size of the file ``gsl-2.6-h294904e_0/lib/libgslcblas.so`` which does not exist.

I modified the script to work around this problem. But presumably, a warning should be raised, because the processed package is incorrect.

PR #342, opened 2020-05-22
Construct from an environment

scopatz:
This PR adds the ability to construct from an environment (name or prefix) or an environment file. This has a couple of main advantages: 

1. It separates the concerns of environment creation and installer creation, enabling the reuse of an `environment.yml` file that already exists for a project
2. It uses the same solve that conda itself produces when creating an environment. This is a useful consistency to have. It is also a mechanism to avoid a bug on certain platforms (Windows) where conda can create the environment just fine, but constructor's `Solver` call fails in weird and unexpected ways, even for trivial specs.

As always, I welcome any feedback!

### DIFF
--- a/constructor/fcp.py
+++ b/constructor/fcp.py
@@ -180,17 +180,17 @@ def check_duplicates_files(pc_recs, platform, ignore_duplicate_files=False):
     return total_tarball_size, total_extracted_pkgs_size
 
 
-def _precs_from_environment(environment, download_dir, conda_exe="conda.exe"):
+def _precs_from_environment(environment, download_dir, user_conda):
     from subprocess import check_output
 
     # get basic data about the environment's packages
     list_flag = "--prefix" if isdir(environment) else "--name"
-    json_listing = check_output([conda_exe, "list", list_flag, environment, "--json"])
+    json_listing = check_output([user_conda, "list", list_flag, environment, "--json"])
     listing = json.loads(json_listing)
     packages = {p["dist_name"]: p for p in listing}
     # get the package install order and MD5 sums,
     # creating a tuple of dist_name, URL, MD5, filename (fn)
-    explicit = check_output([conda_exe, "list", list_flag, environment,
+    explicit = check_output([user_conda, "list", list_flag, environment,
                              "--explicit", "--json", "--md5"],
                             universal_newlines=True)
     ordering = []
@@ -223,7 +223,6 @@ def _main(name, version, download_dir, platform, channel_urls=(), channels_remap
           exclude=(), menu_packages=(), install_in_dependency_order=True,
           ignore_duplicate_files=False, environment=None, environment_file=None,
           verbose=True, dry_run=False, conda_exe="conda.exe"):
-
     # Add python to specs, since all installers need a python interpreter. In the future we'll
     # probably want to add conda too.
     specs = list(concatv(specs, ("python",)))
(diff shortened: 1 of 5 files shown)


scopatz:
Pinging this

mcg1969:
@scopatz looks good to me. I'll check with the team

scopatz:
Thanks @mcg1969!

constructor/fcp.py
```diff
@@ -176,9 +180,49 @@ def check_duplicates_files(pc_recs, platform, ignore_duplicate_files=False):
     return total_tarball_size, total_extracted_pkgs_size
 
 
+def _precs_from_environment(environment, download_dir, conda_exe="conda.exe"):
+    from subprocess import check_output
+
+    # get basic data about the environment's packages
+    list_flag = "--prefix" if isdir(environment) else "--name"
```
mcg1969, line 187: I'm finding this isn't working for me in `--name` mode, and I think it's because the standalone conda executable has a different envs_dirs path than the surrounding conda environment. Does that make sense?

Here's what I did: I created a test environment called `ctest`, with the prefix `/Users/mgrant/miniconda3/envs/ctest`, and I added `environment: ctest` to my `construct.yaml`. Then running `constructor .` crashed with
```
subprocess.CalledProcessError: Command '['/Users/mgrant/miniconda3/standalone_conda/conda.exe', 'list', '--name', 'ctest', '--json']' returned non-zero exit status 1
```
When I try running that command, I get this:
```
{
  "caused_by": "None",
  "error": "EnvironmentLocationNotFound: Not a conda environment: /Users/mgrant/.conda/envs/ctest",
  "exception_name": "EnvironmentLocationNotFound",
  "exception_type": "<class 'conda.exceptions.EnvironmentLocationNotFound'>",
  "location": "/Users/mgrant/.conda/envs/ctest",
  "message": "Not a conda environment: /Users/mgrant/.conda/envs/ctest"
}
```
If I change my `environment:` value to a full path, it works fine. The solution, it seems to me, is either to consider using `os.environ['CONDA_EXE']` instead of `conda_exe`, or to override `CONDA_ENVS_DIRS` temporarily with the appropriate value so it sees the environment directory.

mcg1969: It makes sense when I look at the output of `/Users/mgrant/miniconda3/standalone_conda/conda.exe info`:
```
       envs directories : /Users/mgrant/.conda/envs
                          /var/folders/zr/v087ks_x1dz1vr46g7jqb1880000gp/T/_MEIuACwao/envs
```

mcg1969: The more I think about this, the more I believe that your _environment_ `conda` needs to be used to render the environment, no matter what the version of `conda.exe` is. After all, it seems to me that presumably you are interactively designing and adjusting this environment until it is precisely what you want to see. So the idiosyncracies of your local conda installation are exactly what you want to capture. If for any reason at all the render produced by constructor differs from your local conda, you're likely to prefer the latter.

scopatz: > I'm finding this isn't working for me in --name mode, and I think it's because the standalone conda executable has a different envs_dirs path than the surrounding conda environment. Does that make sense? 

Yeah that makes sense to me and is what I found as well. 

> your environment conda needs to be used to render the environment

I think this makes sense. I'll make the adjustments

scopatz:
Thanks @mcg1969 - I think this is ready for review again.

mcg1969:
Thanks @scopatz! Can we think of the conditions under which `CONDA_EXE` (the environment variable) will not be present? Because if it's not there, and it forces your code to fall back on the `conda_exe` standalone, environment by name will still fail. But I don't think that's likely with relatively recent conda versions.

scopatz:
Yeah, I think that is pretty unlikely that conda won't be in the environment, but I wanted to put the fallback in there. I think what would have to happen is that someone would have to go into their conda install directory and manually remove it, or manually remove the enviornment variable that was set by activation. These are both things that I would consider outside of the supported use cases of conda (but please correct me if I am wrong, I'd love to know :wink:)

I think this is especially true for constructor since there are only conda packages for it (ie it isn't on pypi). I suppose in a dev scenario, you could `pip install --no-deps` into a non-conda Python environment and then it wouldn't be present either.  This shouldn't affect users though.

mcg1969:
I think we're good. Thank you for the great work!

scopatz:
Woohoo! Thanks @mcg1969!

mcg1969:
You're very welcome. Heck I think we're going to use this ourselves!

jakirkham:
Would it be possible to include this in a release? 🙂

cc @jschueller

jschueller:
@scopatz ?

PR #343, opened 2020-05-29
Allow header.sh to operate regardless of content size

mcg1969:
The current version of `header.sh` requires precise knowledge of its own size in order to successfully locate and extract the embedded binaries. This modification allows it to compute those locations using POSIX shell arithmetic and some simple grep/wc logic to determine the `@@END_HEADER@@` location dynamically. The only information it needs inserted are the sizes, in bytes, of the two payloads, with no careful padding needed.

I have endeavored to make sure that only POSIX `sh` arithmetic expressions are used.

Not only does this simplify the `shar.py` logic considerably it opens up the door for us to offer constructor images that can be customized _after building them_. I have some specific applications in mind here.

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -336,18 +336,42 @@ export PREFIX
 
 printf "PREFIX=%s\\n" "$PREFIX"
 
+# 3-part dd from https://unix.stackexchange.com/a/121798/34459
+# this is similar below with the tarball payload - see shar.py in constructor to see how
+#    these values are computed.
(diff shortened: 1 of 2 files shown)


mcg1969:
Thanks @msarahan! I've verified this separately against sh, bash, and zsh as well

mcg1969:
Should test this on busybox first.

mcg1969:
Actually, scratch that, even the existing installers don't work on busybox:glibc

PR #345, opened 2020-06-01
Prevent error popups during silent uninstall mode

wulmer:
This PR is a follow-up to https://github.com/conda/constructor/pull/277.

It wraps the Python `_nsis.py` calls by a macro which either uses `python.exe` (in silent mode) or `pythonw.exe` (else).  The problem with `pythonw.exe` in silent mode is, that it shows a stack trace dialog upon exceptions which might block the silent uninstaller.

And in environments where the silent mode is used in an automated set up, the uninstaller eventually hangs indefinitely and leads to half-uninstalled applications.

### DIFF
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -874,6 +874,13 @@ Section "Install"
     WriteUninstaller "$INSTDIR\Uninstall-${NAME}.exe"
 SectionEnd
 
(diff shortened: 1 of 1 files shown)


mcg1969:
Thank you for working on this! @msarahan, this addresses your concern on #277 — the Python-based removal script still runs, but any errors are silently ignored knowing that they will be covered by the subsequent RMDir command.

mcg1969:
Thanks @wulmer !

PR #353, opened 2020-08-13
Check installation path length

ericpre:
The installation of nodejs and its highly nested directory structure requires installation directory path shorter than ~46 characters in order to avoid the default 260 characters path limits on windows. This PR add an option to check the path length and make sure it is not longer than an estimated limit (based on nodejs 10.13 from the defaults channel).
If the installation path is too long, it would raise an message to request the user to choose a shorter path or enable support for long path on windows 10 (requires admin right).

### DIFF
--- a/CONSTRUCT.md
+++ b/CONSTRUCT.md
@@ -349,6 +349,16 @@ Default choice for whether to register the installed Python instance as the
 system's default Python. The user is still able to change this during
 interactive installation. (Windows only)
 
+## `check_path_length`
+
+required: False
+
+argument type(s): ``bool``,
+
+Check the length of the path where the distribution is installed to ensure nodejs
+can be installed.  Raise a message to request shorter path (less than 46 character)
+or enable long path on windows > 10 (require admin right).  Default is True. (Windows only)
+
 
 ## List of available selectors:
 
--- a/CONSTRUCT.md
+++ b/CONSTRUCT.md
@@ -18,7 +18,7 @@ end, in order to allow customization for selected platforms.
 
 required: True
 
-argument type(s): ``str``,
+argument type(s): ``str``, 
 
 Name of the installer.  May also contain uppercase letter.  The installer
 name is independent of the names of any of the conda packages the installer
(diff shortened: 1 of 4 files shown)


mcg1969:
I really like the patch checking logic. But I'm a little concerned that we are imposing a path restriction _by default_ on Windows that isn't always strictly necessary. Even your documentation string for this option indicates that it's for nodejs. But why restrict path lengths, then, for users who aren't installing nodejs?

ericpre:
Happy to change the default value. The motivation to enable this path restriction by default is to make sure the installation of nodejs will work and no user will experience an installation failure (through the installer or `conda install nodejs`) with an error message not clear/relevant.

mcg1969:
Great motivation, for sure. And if we have some sort of data that suggests that only a small number of our users are _trying_ to install into paths with long lengths, then it would be fine to turn the warning on by default.

mcg1969:
Thanks @ericpre, I'm certainly convinced it would be good to have the option! My only remaining comment—can you please re-run `python scripts/make_docs.py` and commit?

ericpre:
Do I also need to manually update `docs/source/construct_yaml.rst` or is there a script to do that?

ericdill:
I think this script is what you're looking for https://www.github.com/conda/constructor/tree/master/scripts%2Fmake_docs.py

benpapworth:
I observe that with the fix to address the default domain user install to be in %USERPROFILE%\AppData\Local (which is necessary to prevent roaming profile sync breaks), at least 32 characters are already allocated in the default path, which brings this limit quite close.

Can we make this a warning instead of an enforced requirement? I note the text is "should" not "must" in the warning message. Maybe noting that nodejs is the reason why this check was added?

I am also not sure if it's good form to change the LongPathsEnabled registry setting, if the user has Admin privileges, and not advise them that this is an alternate solution to the problem. I don't know the consequences on other applications, nor whether this change would be overridden by group policy later.

Personally, I use subst to put the root of deeply nested directory hierarchies on a drive letter, which was necessary to make some TeamCity builds play nicely with miniconda build environments...

benpapworth:
Reviewing this further, I think the nodejs problem could be quite pernicious -
* conda build is fragile, unless the build path is redirected, as it adds install dir + conda-bld + temporary directory + build and test envs.
* new conda environments are fragile, unless the env path is redirected to be outside of the conda installation too, as it adds install dir + envs + env_name.

Probably better effort would be to make nodejs package report warnings on Windows about it's installation location being too deep!

PR #354, opened 2020-08-14
Allow keep_pkgs to be a command-line option

mcg1969:
I've added a `-k` option to the shell script header that instructs the installer not to remove the package cache contents after the installation. I'm using this to implement an automated workflow to build multiple environment formats from a single constructor specification, and due to `channels_remap` renaming it is important to use the cache provided by the installer itself, not the package cache used to build the installer in the first place.

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -62,6 +62,11 @@ THIS_PATH="$THIS_DIR/$THIS_FILE"
 PREFIX=__DEFAULT_PREFIX__
 BATCH=0
 FORCE=0
+#if keep_pkgs
+KEEP_PKGS=1
+#else
+KEEP_PKGS=0
+#endif
 SKIP_SCRIPTS=0
 TEST=0
 REINSTALL=0
@@ -74,6 +79,9 @@ Installs __NAME__ __VERSION__
              it is expected the license terms are agreed upon
 -f           no error if install prefix already exists
 -h           print this help message and exit
+#if not keep_pkgs
+-k           do not clear the package cache after installation
+#endif
 -p PREFIX    install prefix, defaults to $PREFIX, must not contain spaces.
 -s           skip running pre/post-link/install scripts
 -u           update an existing installation
@@ -81,7 +89,7 @@ Installs __NAME__ __VERSION__
 "
 
 if which getopt > /dev/null 2>&1; then
-    OPTS=$(getopt bfhp:sut "$*" 2>/dev/null)
+    OPTS=$(getopt bfhkp:sut "$*" 2>/dev/null)
     if [ ! $? ]; then
         printf "%s\\n" "$USAGE"
         exit 2
@@ -101,6 +109,10 @@ if which getopt > /dev/null 2>&1; then
                 ;;
             -f)
                 FORCE=1
+                shift
+                ;;
+            -k)
+                KEEP_PKGS=1
                 shift
                 ;;
             -p)
@@ -131,7 +143,7 @@ if which getopt > /dev/null 2>&1; then
         esac
     done
 else
-    while getopts "bfhp:sut" x; do
+    while getopts "bfhkp:sut" x; do
         case "$x" in
             h)
                 printf "%s\\n" "$USAGE"
@@ -142,6 +154,9 @@ else
                 ;;
             f)
                 FORCE=1
+                ;;
+            k)
+                KEEP_PKGS=1
                 ;;
             p)
                 PREFIX="$OPTARG"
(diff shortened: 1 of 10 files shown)


constructor/header.sh
```diff
@@ -556,20 +578,13 @@ if [ "$TEST" = "1" ]; then
...
          mkdir -p "$PREFIX"/conda-bld/__PLAT__
      fi
      cp -f "$PREFIX"/pkgs/*.tar.bz2 "$PREFIX"/conda-bld/__PLAT__/
      cp -f "$PREFIX"/pkgs/*.conda "$PREFIX"/conda-bld/__PLAT__/
      conda index "$PREFIX"/conda-bld/__PLAT__/
      conda-build --override-channels --channel local --test --keep-going "$PREFIX"/conda-bld/__PLAT__/*.tar.bz2
-#endif
-#if not keep_pkgs
-#          conda-build -t -k "$PREFIX"/pkgs/*
-     printf "ERROR: Testing not possible without keep_pkgs\\n"
```
mcg1969, line 570: There is no need for this error anymore because we can keep the package cache in place for testing if that option is selected. We still clear the package cache after tests are complete.

mcg1969: Or more precisely, after the packages have been staged in the `conda-bld/__PLAT__` subdirectory where the testing occurs.

constructor/winexe.py
```diff
@@ -75,10 +75,6 @@ def pkg_commands(download_dir, dists, py_version, keep_pkgs, attempt_hardlinks,
     yield "Pop $0"
     yield r'SetDetailsPrint both'
 
-    if not keep_pkgs:
-        yield ''
-        yield r'RMDir "$INSTDIR\pkgs"'
```
mcg1969, line 80: _Important note for the reviewer_: I have moved this logic into `main.nsi.tmpl` above (see accompanying comment there). However, _this code as written wasn't working_. Because it's not using the `/r` flag, it will not remove the directory since it is not empty.

It takes a non-trivial amount of time to remove the package cache, so fixing this results in a slowdown of installer performance. For this reason, it may be best to _default_ to keeping the package cache on Windows.

constructor/nsis/main.nsi.tmpl
```diff
@@ -803,7 +811,12 @@ Section "Install"
     @PKG_COMMANDS@
 
     # Cleanup
-    Delete "$INSTDIR\pkgs\env.txt"
+    SetOutPath "$INSTDIR"
+    ${If} $ARGV_KeepPkgCache == "1"
+        Delete "$INSTDIR\pkgs\env.txt"
+    ${Else}
+        RMDir /r "$INSTDIR\pkgs"
```
mcg1969, line 818: This is where the package cache is now removed. Note the use of the `/r` flag; without it, it will not work. Furthermore, the `SetOutPath` command a couple of lines up is necessary to remove the directory itself, because without it nsis will be inside `$INSTDIR\pkgs`. Remove that command, and the _contents_ of `$INSTDIR\pkgs` will be removed, but not the directory itself (which frankly would be fine).

mcg1969:
This is now implemented on Windows as well with the `/KeepPkgCache` option. Currently the default is `0`, which will remove the package cache.

mcg1969:
Per my Slack comment, I'm wondering if we should change the default value of `KeepPkgCache` to `1` so as to mimic our current apparently-unintended behavior.

One vote in favor: subsequent uses of `conda-pack` or `conda create --clone` on the installed environment will function properly; see, for instance, this issue: https://github.com/conda/constructor/issues/145

constructor/nsis/main.nsi.tmpl
```diff
@@ -803,7 +811,12 @@ Section "Install"
     @PKG_COMMANDS@
 
     # Cleanup
+    SetOutPath "$INSTDIR"
     Delete "$INSTDIR\pkgs\env.txt"
+    ${If} $ARGV_KeepPkgCache == "0"
+        DetailPrint "Clearing the package cache..."
+        nsExec::ExecToLog '"$INSTDIR\_conda.exe" clean --all --force-pkgs-dirs --yes'
```
mcg1969, line 818: This changed my mind: `conda` can do a better job removing the package cache than NSIS can. So it's reasonably to correct the default behavior to remove the package cache.

constructor/osxpkg.py
```diff
@@ -230,6 +233,8 @@ def create(info, verbose=False):
             ndist = dist.name
         fresh_dir(PACKAGE_ROOT)
         cph_e(join(CACHE_DIR, fn), join(pkgs_dir, dname))
+        if keep_pkgs:
```
mcg1969, line 236: The `.pkg` installer copies in just the extracted packages. But if `keep_pkgs` is True we want the tarballs as well.

mcg1969:
@ericdill thanks for the comments. I'm not fully convinced that this as big a clash as is being suggested though. Keep in mind that it is not constructor _itself_ that is being given a `-k` option. It's the shell script installers that are being generated that are getting it. Even if we wanted to give both `conda` and `constructor` a `-k` option that means "ignore ssl", this would not prevent that.

ericdill:
Ah yeah that makes sense. Ignore the peanut gallery and carry on 🤪

mcg1969:
No worries, @ericdill, I'm actually quite happy to have some peanut gallery reviews! As you can see I'm doing some non-trivial work on this right now so the extra eyeballs are helpful.

PR #355, opened 2020-08-14
Add conda_channel_alias

mcg1969:
With this PR, if `write_condarc` is true, then a `channel_alias:` entry will be created in `$PREFIX/.condarc` and its value will be set to `conda_channel_alias`. Combined with `conda_default_channels`, this allows the constructor to produce an installer preconfigured to point to a specific repository and/or a specific set of channels for `defaults`.


mcg1969:
This is arguably redundant if #357 is adopted

PR #356, opened 2020-08-15
Add multiple installer support

mcg1969:
This PR creates a new entry in `construct.yaml` called `installers`. When supplied, it supports the creation of multiple installers in a single `constructor` call. This is useful if you have multiple installers to build that have a lot of common configuration options that you _intend_ to build at the same time.

The `installers` entry is a dictionary. The keys of this dictionary will typically provide the `name` value for each installer, while the dictionary value is a dictionary that is a (possibly partial) standard constructor specification. If this child specification supplies a `name` value, then the dictionary key is ignored.

The top-level key/value pairs in the `construct.yaml` are then used to provide global default values, and a simple one-level merge strategy is employed:

- For string values, a child value, if present, overrides the default presented in the parent spec.
- For list values, the parent values _prepend_ the values in the child spec. For instance, you can define a global set of packages in `specs`, and then augment them in each child; e.g.
  ```
  merged_value = (parent_value or []) + (child_value or [])
  ```
- For dictionary values, each key/value pair in the parent value serves as a default for the child. That is:
  ```
   merged_value = (parent_value or {}).copy()
   merged_value.update(child_value or {})
   ```

I modified the PR code in a slightly less elegant fashion that I would prefer to emphasize how little had to change there to make this work.

### DIFF
--- a/CONSTRUCT.md
+++ b/CONSTRUCT.md
@@ -245,10 +245,11 @@ and a `.bat` file for Windows.
 
 _required:_ no<br/>
 _type:_ string<br/>
-Short description of the "post_install" script to be displayed as label of
-the "Do not run post install script" checkbox in the windows installer.
-If used and not an empty string, the "Do not run post install script"
-checkbox will be displayed with this label.
+A description of the purpose of the supplied post_install script. If this
+string is supplied and non-empty, then the Windows and macOS GUI installers
+will display it along with checkbox to enable or disable the execution of the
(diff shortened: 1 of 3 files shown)


mcg1969:
Example `construct.yaml`:
```
version: 2020.08.15
specs:
  - conda
channels:
  - defaults
installers:
  Miniconda: {}
  Anaconda:
    specs:
      - anaconda
    exclude:
      - anaconda
ignore_duplicate_files: true
```

PR #360, opened 2020-08-22
Experiment with github actions

mcg1969:


mcg1969:
All tests are now being run on the entire matrix. I've manually reviewed the Travis and AppVeyor CI to ensure we're reproducing their tests. The only thing I have not done is attempt to run this against a matrix of conda versions. I'm not convinced that is 100% necessary, but if so, I think we can bring that back in easily enough.

I'm also not sure the codecov upload is working properly, but that can also be fixed in a future PR. In fact what I would prefer we do is move the tests outside of the package altogether and run them outside of conda build.

PR #372, opened 2020-08-30
Only init and create envs if conda is installed in the prefix

hmaarrfk:
It is useful not to have init be a question during the install procedure if you aren't really installing conda, but other software as the main software.

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -450,6 +450,12 @@ fi
 
 __INSTALL_COMMANDS__
 
+if [ -f "$PREFIX/bin/conda" ]; then
+    HAS_CONDA_IN_INSTALL=1
+else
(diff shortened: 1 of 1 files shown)


mcg1969:
Nice! Thank you sir.

mcg1969:
This is marked `WIP`; is there a reason for this?

hmaarrfk:
I was waiting till I got my stuff in order to test this.

I can confirm that I used a patched version of constructor to test this out in action.

mcg1969:
Ah, to be fair, I wonder if we should endeavor to reproduce this behavior in `.pkg` and `.exe` installers too.

hmaarrfk:
yes, i've only tested on linux.

mcg1969:
Will merge this soon—Mac .pkg is handled by #395

PR #376, opened 2020-08-31
Removed deprecated platform.dist()

risyomei:
Since platform.dist() is deprecated, I have removed that from the code.
It was tested under centos7 and looks fine.

I  am not confident with the dependency part, though.
Any comment is appreciated.

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -29,6 +29,7 @@ requirements:
     - conda-standalone
     - pillow >=3.1     # [win]
     - nsis >=3.01      # [win]
+    - distro
 
 test:
   source_files:
--- a/constructor/preconda.py
+++ b/constructor/preconda.py
@@ -4,6 +4,7 @@
 # constructor is distributed under the terms of the BSD 3-clause license.
 # Consult LICENSE.txt or http://opensource.org/licenses/BSD-3-Clause.
 
+import distro
 import os
 from os.path import isdir, join, split as path_split
 import platform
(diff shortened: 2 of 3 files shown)


mcg1969:
Thank you so much for doing this! I must request however that we approach this differently. Anaconda has not yet added `distro` to the defaults channel, so this package would no longer build against defaults, and that is a requirement for us.

However, `conda` actually _vendors_ the `distro` package. So for most recent versions of `conda`, this will work:
```
from conda._vendor import distro
```
However, the intent is to use this only for Linux, so it would need to be loaded conditionally.

So I'm thinking of something like this:
1. Conditionally import `conda._vendor.distro`; if it succeeds, use that; otherwise,
2. Conditionally import `distro`; if it succeeds, use that; otherwise,
3. If `hasattr(platform, 'dist')`, use that; otherwise,
4. Return a safe default.

This logic would be best baked into `utils.py`.

Thoughts?

risyomei:
Hey, @mcg1969 
Thank you for the comment, I will try to implement what you mentioned~

mcg1969:
@risyomei I'm replacing this PR with #380, in which you still have authorship. We do need you to sign the CLA, however:
https://conda.io/en/latest/contributing.html#conda-contributor-license-agreement
Once you have done that, I'll be able to get your change in. Thanks!

mcg1969:
@risyomei I've decided that my implementation was sufficiently distinct from yours that we can proceed without your CLA signature. But I am tremendously grateful for bringing this issue to our attention for correction—and if you _would_ go ahead and sign the CLA, we can make sure that all future contributions are automatically covered. And I do hope you will continue to support the development of this package!

PR #377, opened 2020-08-31
Improve compatability of pre_install and post_install scripts 

### LINKED ISSUE #367: pre_install and post_install are not always run in a bash-like shell
## Problem

In https://github.com/conda/constructor/blob/master/CONSTRUCT.md it claims:

> ##### `pre_install`
> Path to a pre install script. This is available for Unix only, and must be a Bash script.
> 
> ##### `post_install`
> Path to a post install script. This must be a Bash script for Unix and a `.bat` file for Windows.

However the code actually runs the script in the current shell:

https://github.com/conda/constructor/blob/9ae44d4355da774220cd3aebf6ea40f58b51f37e/constructor/header.sh#L500-L503

This makes it quite hard to write pre/post install scripts that will reliably work on Linux and macOS.

## Proposed solutions

I see three options:

1. Allow the `pre_install.sh`/`post_install.sh` to be executable and just run them directly

2. Check the basename of the shell binary, and only allow known-good values such as `sh` and `bash`. If this isn't the use the fallbacks:

https://github.com/conda/constructor/blob/9ae44d4355da774220cd3aebf6ea40f58b51f37e/constructor/header.sh#L43-L52

3. Fix the documentation to make clear than the scripts must be able to run in any possible shell (bash/sh/fish/csh/xnosh/...)

I'm inclined towards 2. If people agree I can implement it in a PR.

chrisburr:
Fixes #367.


mcg1969:
Love the work here, in particular the simplification of the header code.

We need to make sure this is performing properly on Mac pkg installs. Will play with this

mcg1969:
I just realized that we don't have post_install script support for the `.pkg` format. That will need a doc fix, but it can be separate from this.

mcg1969 (approved):

mcg1969:
@chrisburr I pushed some doc fixes. Turns out you need to edit `construct.py` for that; `CONSTRUCT.md` is generated automatically. I cleaned up the language a bit too. But we'll merge this and cut a release once this passes again.

mcg1969:
Thank you for doing this @chrisburr !

PR #378, opened 2020-09-02
Update CONSTRUCT.md

marimeireles:
I believe there were some typos here, because if I use `win64` instead of `win-64` for example I get `Error: invalid platform string 'win64'` but when I use `win-64` it works.


jjhelmus:
@cla-bot check

mcg1969:
Thank you @marimeireles !

PR #397, opened 2020-09-27
Handle two leading slashes in the PREFIX for shell installers

### LINKED ISSUE #284: Errors when installing miniconda to an unnormalised prefix
## Current Behavior
Installing miniconda/anaconda into an unnomalised path results in the unnomalised path becoming part of the installation (e.g. `//home/myuser/miniconda`).

This has caused a reproducible relocation issue with a conda-forge package on macOS where anaconda was installed into `//anaconda3`. https://github.com/conda-forge/root-feedstock/issues/53

### Steps to Reproduce
```bash
# Install miniconda on macOS with a prefix which starts with //
conda create --name test-no-norm root python=3.7
conda activate conda
curl http://root.cern.ch/files/WM.root > test.root
root test.root
# See errors of the form
# Error in <TFile::TFile>: file anaconda3/envs/roottest/lib/libNet_rdict.pcm does not exist
```

## Expected Behavior
The install prefix given by the user should be automatically normalised.

## Environment Information
macOS with the latest miniconda installer.
Also seen with an older Anaconda install.

mcg1969:
Fixes #284.

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -332,7 +332,9 @@ if ! mkdir -p "$PREFIX"; then
     exit 1
 fi
 
-PREFIX=$(cd "$PREFIX"; pwd)
+# pwd does not convert two leading slashes to one
+# https://github.com/conda/constructor/issues/284
+PREFIX=$(cd "$PREFIX"; pwd | sed 's@//@/@')
 export PREFIX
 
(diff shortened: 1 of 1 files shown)


mcg1969:
https://github.com/conda/constructor/issues/284#issuecomment-699668923

PR #415, opened 2020-11-15
Fix default install folder domain user

### LINKED ISSUE #412: Windows: Install to %USERPROFILE% is not appropriate for roaming profiles
This fix did correct some bad logic with domain user selection:
https://github.com/conda/constructor/commit/d5a4ceba3bbec5f7568ba42311809787347fcc11

But, making the default for domain users install to %USERPROFILE% is a bad choice because the installation (which can grow to Gbs in size) will now be in the roaming profile by default; therefore sync'd with network at each login/logout. As soon as conda install becomes large enough, this will timeout and break the roaming sync process. Most end users will not be aware of this subtle problem until it significantly slows their log-in/log-out time or breaks their roaming profile synchronization.

It would be more appropriate to:
* Not ignore DEFAULT_PREFIX, if defined, for JustMe installs, or provide a DEFAULT_JUSTME_PREFIX to allow installer writers to control the option.
* The default install location should be %LOCALAPPDATA%, as this is not on the roaming profile, but is still owned by the domain user on the workstation. This will avoid future roaming profile sync problems.

ericpre:
As per one of @benpapworth's suggestion in #412: 
- change the default prefix for domain user to %LOCALAPPDATA%
- add an option to set its default value.

Closes #412.

### DIFF
--- a/CONSTRUCT.md
+++ b/CONSTRUCT.md
@@ -279,6 +279,23 @@ _type:_ string<br/>
 Path to a pre uninstall script. This is only supported for on Windows,
 and must be a `.bat` file.
 
+## `default_prefix`
+
+_required:_ no<br/>
+_type:_ string<br/>
+Set default install prefix. On Linux, if not provided, the default prefix is
+`${HOME}/${NAME}`. On windows, if not provided, the default prefix is
+`${USERPROFILE}\${NAME}`.
+
+## `default_prefix_domain_user`
+
+_required:_ no<br/>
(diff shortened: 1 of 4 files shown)


mcg1969:
@ericpre I'm grateful for your work here—I only have a small documentation comment/question/request, forthcoming.

mcg1969 (approved):
Actually nevermind, I think the doc is fine! Thanks again!

PR #428, opened 2021-01-16
Modify regex that finds platform selectors

### LINKED ISSUE #427: Bug: Regex for grabbing selectors also grabs any other square brackets
## Description:
This regex `(.+?)\s*\[(.+)\]$` looks for "selectors" anywhere in the line following at least one white-space instead of limiting itself to the comment at the end of the line.

https://github.com/conda/constructor/blob/b9e007b7170bdb7f54b1b52f376842e682d7524d/constructor/construct.py#L313

This means that any square brackets in a line that also has a selector will break the logic. This might not have been a problem before, but now that environment variables are injected into the jinja2 templating, this is effectively making it impossible to use both a platform selector and an environment variable on the same line.

## Example:
Consider the following line:

`{{ environ["FOO"] }} # [win64]`

Expected behavior is that the regex captures `{{ environ["FOO"] }}` in the first group, and `win64` in the second group.

Actual behavior is that the regex captures `{{ environ` in the first group and `"FOO"] }} # [win64` in the second group.

## Proposed Solution:
Modify the regex slightly to explicitly look for the selector in the comment at the end of the line:

```diff
-sel_pat = re.compile(r'(.+?)\s*\[(.+)\]$')
+sel_pat = re.compile(r'(.+?)\s*# \[(.+)\]$')
```
This forces the selector syntax to comply with that laid out [here](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#preprocessing-selectors): "A selector has the form `# [<selector>]` at the end of a line."

FaustinCarter:
This fixes the regex that grabs platform selectors to _only_ grab them from the comment at the end of the line, rather than grabbing them from anywhere on the line.

Closes #427

### DIFF
--- a/constructor/construct.py
+++ b/constructor/construct.py
@@ -13,7 +13,6 @@ from .utils import yaml
 from constructor.exceptions import (UnableToParse, UnableToParseMissingJinja2,
                                     YamlParsingError)
 
-
 # list of tuples (key name, required, type, description)
 KEYS = [
     ('name',                   True,  str, '''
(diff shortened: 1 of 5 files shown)


nehaljwani:
I like the change, but should it be considered a breaking change?

FaustinCarter:
> I like the change, but this would likely be considered a breaking change, right?

In the strictest sense, I don't think so. Constructor documentation has stated for ages that the conda-build documentation is the definitive documentation on selector format. Also, Miniconda uses the correct selector format in its construct.yaml.

That being said, a random Google search for `construct.yaml` showed at least two officialish repositories that are doing it the same way the examples did. Given that at least some folks are using the examples rather than reading the documentation, then it probably makes sense to consider it breaking.

Probably this means catching old/improper use and issuing a deprecation warning for at least one minor version before switching over entirely. I'll think on this some and push an update to this PR.

One possible interim fix would be to instead switch the regex to:

```diff
-sel_pat = re.compile(r'(.+?)\s*\[(.+)\]$')
+sel_pat = re.compile(r'(.+)\s*\[(.+)\]$')
```

This makes the first capture group greedy, which means that it will do its best to only grab the _last_ set of square brackets in the line.

FaustinCarter:
Broken windows builds: #429

FaustinCarter:
@nehaljwani After some thinking, it occurred to me to look and see what the `conda-build` project was doing here. Lo and behold, they have a fancy regex that does everything both of us want!  So I copied that over from [here](https://github.com/conda/conda-build/blob/fa451287861771106bf19abac168659042577219/conda_build/metadata.py#L159), and also copied their `select_lines` function from [here](https://github.com/conda/conda-build/blob/fa451287861771106bf19abac168659042577219/conda_build/metadata.py#L191) and modified it slightly to match what `constructor` needs.

mcg1969:
I'm working on the Windows failures.

mcg1969:
I'm good with this. Thanks!

FaustinCarter:
@mcg1969 thanks for fixing the duplicate files issue! Any chance for a patch release soon to get these two fixes into the wild?

mcg1969:
Yep, we can get one out soon. But remember you can always grab the bleeding edge builds from ctools/label/dev ...

PR #440, opened 2021-02-20
Add parameter to make batch mode the default

chrisburr:
This PR adds a parameter (`batch_mode`) to `construct.yaml` to make batch mode (`-b`) the default when running the installer. If both `batch_mode` and `license_file` are set an exception is raise as there isn't really a way to "agree" to the license.

Any thoughts?

### DIFF
--- a/CONSTRUCT.md
+++ b/CONSTRUCT.md
@@ -102,8 +102,8 @@ supplying this list allows a subset to be selected instead.
 
 _required:_ no<br/>
 _type:_ boolean<br/>
-By default, constructor will error out when adding packages with duplicate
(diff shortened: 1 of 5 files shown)


hmaarrfk:
seems like an `--interactive` flag should be added to allow users to use interactive mode?

chrisburr:
> seems like an `--interactive` flag should be added to allow users to use interactive mode?

Thanks for taking a look! Yeah that's a good idea, I've added a `-i` flag to match the style of the existing arguments.

mcg1969:
What a great idea. Love it.

PR #444, opened 2021-03-24
init all

cjmartian:
Init all shells so that the user can have a better experience if they are switching between shells. 

This was brought up due to osx's change to zsh as the default shell.

### DIFF
--- a/constructor/osx/update_path.sh
+++ b/constructor/osx/update_path.sh
@@ -6,9 +6,4 @@
 # change.
 PREFIX="$2/__NAME_LOWER__"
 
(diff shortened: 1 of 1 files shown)


mcg1969:
This reminds me that, to the best of my knowledge, the pkg installer does not currently provide a mechanism for disabling the init process. Seems like it should. But this doesn't really change that, it just makes it actually work, now that Apple has seen fit to change the default shell out from under us. So I'm fine with this if it works

cjmartian:
@mcg1969 Just tested and verified that this is working as expected.

PR #455, opened 2021-08-02
Fix windows shortcuts

ericpre:
- [x] Fix ignore `menu_packages` entry in `construct.yaml`
- [x] Pass `prefix` to `rm_menus` to fix removing menu - see https://github.com/AnacondaRecipes/conda-standalone-feedstock/pull/5
- [x] Pass `root_prefix` to `menu.install`, which can be used when build a standalone `conda`.

These two commits are used as patch in https://github.com/conda-forge/conda-standalone-feedstock/pull/18.


constructor/nsis/main.nsi.tmpl
```diff
@@ -881,7 +881,7 @@ Section "Install"
 
     SetDetailsPrint TextOnly
     DetailPrint "Setting up the base environment ..."
-    nsExec::ExecToLog '"$INSTDIR\_conda.exe" install --offline -yp "$INSTDIR" --file "$INSTDIR\pkgs\env.txt"'
+    nsExec::ExecToLog '"$INSTDIR\_conda.exe" install --offline -yp "$INSTDIR" --file "$INSTDIR\pkgs\env.txt" --no-shortcuts'
```
jezdez, line 884: What does this parameter do?

ericpre: This doesn't create the start menu shortcuts. We don't want to create the shortcuts with `conda install` because we do it in the following step with a dedicated option of conda-standalone in order to support the `menu_packages` option of `constructor`:
```
"$INSTDIR\_conda.exe" constructor --prefix "$INSTDIR" --make-menus @MENU_PKGS@
```
I added a comment explaining it.

jezdez: Ah, got it, thank you!

PR #458, opened 2021-09-10
Fix help and conda init

ericpre:
- [x] Fix help shell installer - the shell header wasn't parse correctly
- [x] Fix `initialize_by_default` constructor option - same as previous point
- [x] Add option to shell installer not to run conda init and don't ask the corresponding question when this option is used
- [x] Ready for review

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -42,13 +42,13 @@ Installs __NAME__ __VERSION__
 
 #if batch_mode
 -i           run install in interactive mode
-#else
-  #if has_license
+#endif
+#if not batch_mode and has_license
 -b           run install in batch mode (without manual intervention),
              it is expected the license terms are agreed upon
-  #else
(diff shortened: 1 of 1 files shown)


hmaarrfk:
xref: https://github.com/conda-forge/miniforge/issues/127#issuecomment-917380346

hmaarrfk:
See alternative: https://github.com/conda/constructor/pull/459

ericpre:
I have added the option to disable running `conda init` even if `initialize_by_default` is used when building the installer. The motivation was to be able to keep using the current behaviour even if `initialize_by_default` will be used in future release of `miniforge`. However, thinking again about it, I am not sure this is necessary because running shell installer in batch mode can be used instead and therefore this new option is redundant!

#459 is more simple, so I am closing this PR in favour of #459.

PR #459, opened 2021-09-11
Add tests for header template preprocessing and fix initialize by default

hmaarrfk:
It seems that nested if statements are not supported by the template processor.

The choice of template processor is rather secondary for me. This seems like the minimal fix things to pass.

I also added tests so that new features don't go silently "forgotten" so as to avoid this happening again.


ericpre (commented):
I have closed #458 in favour of this PR - see https://github.com/conda/constructor/pull/458#issuecomment-920287212

It looks good from reading the code but I haven't tested the installer.

hmaarrfk:
Kind ping

mcg1969:
Working on this now

mcg1969 (commented):
thanks!

PR #473, opened 2021-11-11
Avoid yaml warnings (and ensure compatibility with pyyaml 6)

hmaarrfk:
I had to manually downgrade pyyaml. Not sure if this is the only place where yaml load is used.


hmaarrfk:
Has anybody had time to put thoughts into this? This is currently breaking the conda forge ecosystem due to the latest release of pyyaml.

jezdez (approved):
Good catch, yeah, this seems way overdue. Refs https://github.com/yaml/pyyaml/pull/561.

PR #475, opened 2021-12-23
add windows installer signing

### LINKED ISSUE #453: makensis exit code not checked on Windows
In verbose mode the makensis return code is not properly checked. Any failure of makensis will be treated as a success by constructor. This is because `subprocess.Popen.communicate` (line 209) does note raise an error for non-zero return codes the same way that `subprocess.check_call` (line 217) does.

https://github.com/conda/constructor/blob/a7ad3d2807129b2cdc596337b5081df1967002d5/constructor/winexe.py#L207-L220

Suggested fix is to explicitly check `sub.returncode` and raise an error.

jaimergp:
Similar to `signing_identity` for macOS `pkg` installers, this adds some simple logic to sign the Windows installers with `signtool`.

This PR adds a new keyword to `construct.yaml`, `signing_certificate`, which should point to a PFX certificate in disk. Additional configuration options are exposed through environment variables if needed:

* `CONSTRUCTOR_SIGNTOOL_PATH`: absolute path to `signtool.exe`, in case this is not in PATH.
* `CONSTRUCTOR_PFX_CERTIFICATE_PASSWORD`: password to unlock the certificate, if
needed.
* `CONSTRUCTOR_SIGNTOOL_TIMESTAMP_SERVER_URL`: custom timestamping server. It needs to support RFC 3161. Default one is http://timestamp.sectigo.com.

***

It also fixes #453


hmaarrfk:
I'm not able to comment on the content, but make sure you test the uninstaller as well.

larsoner (approved):
@jaimergp can you comment on whether or not you've tested the uninstaller as mentioned by @hmaarrfk ?

jaimergp:
Isn't the uninstaller generated by NSIS itself on the user machine during installation? I don't think that requires a certificate, but I guess it doesn't hurt to check.

Does anybody have a Windows-recognized certificate? I could only test it with self-signed ones, so the SmartScreen stuff still popped up despite the binaries being signed.

jaimergp:
Ah wait, I am mistaken. Turns out we can use NSIS-native options for this.

https://stackoverflow.com/a/70797528/3407590

I'll update the PR.

jaimergp:
We need NSIS 3.08 for this to work. Submitted here: https://github.com/conda-forge/nsis-feedstock/pull/20

jaimergp:
We can't uses NSIS 3.08 til we figure out #526

jaimergp:
Once #563 lands we can resume this PR!

goanpeca (approved):
Awesome work @jaimergp !!! 🚀

mcg1969:
@jaimergp is on FIRE. Thank you so much for all this great work.

jaimergp:
This is ready for review @conda/constructor!

The CI can sign the installers too (with a self signed certificate that doesn't prevent SmartScreen, but the process is the same if we were using a real certificate). I downloaded the artifacts to make sure both the installer _and_ the uninstaller are signed. Proof:

Installer:

image

Uninstaller:

image


You can try too if you want. Just check the `Signed-X-Windows-x86_64.exe` installer :)

larsoner (approved):
> with a self signed certificate that doesn't prevent SmartScreen

Slightly off-topic, but I'd love to sign our installers with one that *does* prevent SmartScreen / is legitimate but haven't found a good solution. If you or someone else has a preferred solution, let me know!

Thanks @jaimergp !

jaimergp:
To be clear, this PR does provide what you need to sign the installer properly, provided you have a Microsoft-blessed certificate.

Getting one of those does require buying one from a provider like [Comodo](https://www.comodo.com/business-security/code-signing-certificates/code-signing.php). As long as you provide it as a PFX certificate, you should be ok. 

That said, I never got my hands into a _proper_ Microsoft-recognized certificate, so I never got that far. (I did test with self signed certificates added to the trusted roots in Windows). Let me know how it goes, I'll be happy to help and adapt the code or review the PR as needed!

PR #477, opened 2021-12-29
Fix environment and environment_file modes

mcg1969:
I was finding something interesting: if I built an installer using `environment` or `environment_file` modes—which causes the constructor to inspect an _existing environment_ (or one that it creates with the user's conda) to determine the installer recipe—then the conda environment metadata is corrupt. Specifically, the `depends` entries were empty, causing conda to think my environment was fully of PyPi packages, among other things.

I traced the cause to the `_precs_from_environment` function, which tried to use primarily external `conda` calls to inspect the environment, instead of using the full power of the `PrefixData` object. The `depends` field was simply not being populated. I've rewritten that function considerably here.

I've also modified the `environment_file` behavior. If the suffix is `.yml` or `.yaml`, it uses `conda env create --file` to create the environment; otherwise it uses `conda create --file`.

Finally, I've made it optional to supply `channels` or `channels_remap` list in these modes. `channels` is definitely unnecessary, but `channels_remap` might still be useful.

I'm reluctant to merge this without external review from someone who actually uses the `environment` approach to building installers, to make sure I didn't miss anything. That said, I did create three new examples that are run by the `run_examples.py` script...


hoechenberger:
Hello, I've run into the same issue and would appreciate if this could get reviewed & merged :) I'm also willing to help test things if you tell me what to do! I have access to Linux and macOS computers.

Thank you!

hoechenberger:
@mcg1969 I just created an installer using your `constructor` branch, and while `conda list` doesn't claim the packages were installed from `pypi`, the `Channel` column now is entirely empty. I don't think this is intended behavior, or is it?

constructor/fcp.py
```diff
@@ -198,51 +198,24 @@ def check_duplicates_files(pc_recs, platform, ignore_duplicate_files=True):
...
+def _precs_from_environment(environment, dir_path):
+    if not isdir(environment) and ('/' in environment or '\\' in environment):
+        env2 = join(dir_path, environment)
+        if isdir(env2):
+            environment = env2
+    if isdir(environment):
+        environment = abspath(join(dir_path, expanduser(environment)))
+    else:
+        environment = locate_prefix_by_name(environment)
+    pdata = PrefixData(environment)
```
jaimergp, line 210: The only drawback I can anticipate is that the user wanted to use their supplied `conda-exe` for this, but it's not like `PrefixData` has changed a lot in the past and it doesn't look like it will do so anytime soon.

constructor/fcp.py
```diff
@@ -198,51 +198,24 @@ def check_duplicates_files(pc_recs, platform, ignore_duplicate_files=True):
...
-    return precs
+def _precs_from_environment(environment, dir_path):
+    if not isdir(environment) and ('/' in environment or '\\' in environment):
+        env2 = join(dir_path, environment)
+        if isdir(env2):
+            environment = env2
+    if isdir(environment):
+        environment = abspath(join(dir_path, expanduser(environment)))
+    else:
+        environment = locate_prefix_by_name(environment)
```
jaimergp, line 209: If I am reading this correctly, `environment` can be a working path on its own, or a partial path that needs `dir_path`? Hence why all the checks are needed? This logic was not present in the previous code so I want to double check if this is necessary or not.

jaimergp (approved):
This LGTM now.

I adjusted the examples tests so they are part of the pytest module, which allows us to test more things too.

A cool side effect of this PR is that, since it adds support for TXT files, we can create installers from explicit lock files now! Added a test too just to make sure and it works beautifully ;)

I'll leave this open for some days in case other members of @conda/constructor want to review too.

larsoner (approved):

PR #479, opened 2021-12-31
DNM: Build matrix: drop 2.7, 3.6; add 3.9, 3.10

mcg1969:
https://endoflife.date/python
https://numpy.org/neps/nep-0029-deprecation_policy.html

In theory, we could just expand the build matrix here, without dropping Python 2.7 and Python 3.6. But while I support the continued use of Python 2.7 in legacy code, constructor is not really a legacy application.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest,ubuntu-latest,windows-latest]
-        pyver: [2.7,3.6,3.7,3.8]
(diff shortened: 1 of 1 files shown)


mcg1969:
Looks like Python 3.10 would be premature.

jaimergp:
Is it because it's relying on 3.10 deployment in defaults?

mcg1969:
I have no reason to believe 3.10 won't work. Must be a dependency issue in defaults.

mcg1969:
I actually think it's just our test deps. Should be fixable

jezdez:
Let's skip 3.10 for now and move 2.7 and 3.6 out for now.

PR #486, opened 2022-01-28
Add new options to the generated installers

chrisburr:
I've found myself wanting to use three these commits of functionality when using custom installers generating with constructor.

* Optionally disable the cleaning of `LD_LIBRARY_PATH`
* Allow the path to `conda.exe` to be replaced
* Allow pre/post installer scripts to see the installer's name/version/platform

To be more specific, I'm finding 81cab9eedecb043641a4fbb66efbf70b13cfd412 and de6c23db76a4cb063dd7ab2e7a55c581a9708267 to be useful when installing from a different architecture (in my case, to a read-only networked file system that is only writeable from a single x86 machine).

I considered adding this as environment variables instead to avoid bloating the number of CLI flags but I guess this way is more idiomatic.

### DIFF
--- a/CONSTRUCT.md
+++ b/CONSTRUCT.md
@@ -257,7 +257,9 @@ _type:_ string<br/>
 Path to a pre-install script. For Unix `.sh` installers, the shebang
 line is respected if present; otherwise, the script is run by the POSIX
 shell `sh`. Note that the use of a shebang can reduce the portability of
-the installer. This option is not supported for Windows `.exe` or macOS
+the installer. Metadata about the installer can be found in the
+`${INSTALLER_NAME}`/`${INSTALLER_VER}`/`${INSTALLER_PLAT}` environment
+variables. This option is not supported for Windows `.exe` or macOS
 `.pkg` installers.
 
 ## `post_install`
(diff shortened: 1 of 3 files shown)


chrisburr:
Unfortunately these commits conflict so separate PRs are a bit of a pain to do in parallel. I can rebase to remove any of these and open a new PR with the other(s) if you can tell me which you'd prefer:

1. 837b1d8
2. 3a89c2e <- I guess this one would be the most questionable?
3. 8ee1890

jezdez:
> Unfortunately these commits conflict so separate PRs are a bit of a pain to do in parallel.

Generally speaking it's an anti-pattern to lump a number of changes into one PR since it makes it harder to review and in total increases the amount of time to get it merged. I understand that given the current long response times for pull requests, it's attractive to cut corner of course, so let's just say I'd prefer if we could have separate reviews simply it allows me (a relative newbie with the constructor codebase) to learn why you found these changes useful.

> I can rebase to remove any of these and open a new PR with the other(s) if you can tell me which you'd prefer:
>
> 1. [837b1d8](https://github.com/conda/constructor/commit/837b1d80ce4f0e37f443c969b5f3e9b277f700f9)

This seems like a footgun and I want to get input from @chenghlee who has more experience using constructor.

> 2. [3a89c2e](https://github.com/conda/constructor/commit/3a89c2e58046c8ee582d1597b06dbc595b3f9e93) <- I guess this one would be the most questionable?

Questionable is too harsh, but I'd like to understand what the use case is before merging. Is this related to microconda as well? Is there a risk of this API to be misused?

> 3. [8ee1890](https://github.com/conda/constructor/commit/8ee18904df70fac2162e4b079283c77332198c28)

This seems like a useful change to make installers aware of its context.

chrisburr:
As suggested originally, I'm going to move this to separare PRs and will address the questions in those 😄

PR #487, opened 2022-01-28
Micromamba and transcoding fixes

chrisburr:
When transcoding to `.conda` and using `--conda-exe=/path/to/micromamba` I found that the installer would fail with a lot of error messages along the lines of:

```
Encountered problems while solving:
  - nothing provides ncurses >=6.2,<6.3.0a0 needed by python-3.9.9-h62f1059_0_cpython
  - nothing provides ncurses >=6.2,<6.3.0a0 needed by libedit-3.1.20191231-he28a2e2_2
  - nothing provides ncurses >=6.2,<6.3.0a0 needed by readline-8.1-h46c0cb4_0
  - nothing provides ncurses >=6.2,<6.3.0a0 needed by sqlite-3.37.0-h9cd32fc_0
  - nothing provides ncurses >=6.2,<6.3.0a0 needed by xrootd-5.4.0-py39hf708bbc_0
```

Initially this could be fixed by writing the cache filenames that `micromamba` expects to find. I think hit errors with the extracted `pkgs/*/info/repodata_record.json` files being invalid. This was caused by `micromamba` failing to find the [package name](https://github.com/mamba-org/mamba/blob/129920a304084a59bdfbe026b036185a83d94e40/micromamba/src/constructor.cpp#L100-L117) as the repodata wasn't being modified to look replace `.tar.bz2` names with `.conda` (and the fix the size/md5/sha256).

With this PR `constructor --conda-exe=/path/to/micromamba` works fully for every combination of options I can think of.

cc @wolfv

### DIFF
--- a/constructor/conda_interface.py
+++ b/constructor/conda_interface.py
@@ -126,6 +126,59 @@ if conda_interface_type == 'conda':
                 data["md5"] = hash_files([pkg_fn])
                 used_repodata[key][package] = data
 
+        write_cache_types = info.get("write_cache_types", None)
+        if not write_cache_types:
+            conda_exe_fn = os.path.basename(info["_conda_exe"])
+            if "conda.exe" in conda_exe_fn:
+                write_cache_types = ["conda"]
+            elif "micromamba" in conda_exe_fn:
+                write_cache_types = ["mamba"]
+            else:
+                write_cache_types = ["conda"]
+                print(
+                    "WARNING: Failed to detect write_cache_types using",
+                    conda_exe_fn,
+                    "defaulting to",
+                    write_cache_types,
+                )
+
+        for write_cache_type in write_cache_types:
+            if write_cache_type == "conda":
+                _write_repodata_conda(used_repodata, cache_dir, url)
+            elif write_cache_type == "mamba":
+                _write_repodata_mamba(used_repodata,cache_dir, url)
+            else:
+                raise NotImplementedError("unsupported write_cache_type: %s" % write_cache_type)
+
+
+    def _write_repodata_conda(used_repodata, cache_dir, url):
         repodata_filename = _cache_fn_url(used_repodata['_url'].rstrip("/"))
         with open(join(cache_dir, repodata_filename), 'w') as fh:
             json.dump(used_repodata, fh, indent=2)
+
+
+    def _write_repodata_mamba(used_repodata, cache_dir, url):
+        try:
+            import libmambapy
+        except ImportError:
+            raise ImportError("libmambapy is required for write mamba-style repodata caches")
+
+        # Mamba relies on the first line of the cache to get so metadata
+        full_url = url + "/repodata.json"
+        used_repodata.pop('_url')
+        _mod = used_repodata.pop('_mod')
+        mamba_repodata = json.dumps(used_repodata, indent=2)
+        mamba_repodata_header = json.dumps(
+            {
+                "_mod": _mod,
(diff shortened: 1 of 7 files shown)


constructor/conda_interface.py
```diff
@@ -87,20 +91,94 @@ def get_repodata(url):
...
+
+
+    def _write_repodata_conda(used_repodata, cache_dir, url):
+        repodata_filename = _cache_fn_url(used_repodata['_url'].rstrip("/"))
         with open(join(cache_dir, repodata_filename), 'w') as fh:
             json.dump(used_repodata, fh, indent=2)
+
+
+    def _write_repodata_mamba(used_repodata, cache_dir, url):
+        try:
```
wolfv, line 161: do we really need this? Because it looks a bit like a bug in mamba if that's the case. We should be able to read a repodata.json without any additional metadata

jezdez: I concur that if this is a hack to fix something in mamba's implementation, this shouldn't be added here. To get a better understanding, could you elaborate why this isn't using the same repodata cache format as conda?

jezdez (commented):
Hey @chrisburr, thanks for the patch, but I'm not sure I'm following why this is needed and I would be grateful if you could elaborate in the pull request description to better understand why micromamba needs a different repodata cache format and if there is a less hacky way to implement this.

FWIW I've looked at the code and it makes sense, but assumes that it's okay to keep innovating community standards by just splitting things up between "conda" and "mamba", but we know that got us in a lot of heartache around conda's long-term maintenance in the past. Let's try to figure out a way to make this happen without the split. I know I'm asking you to do extra work, but it would be cool to at least get the reasons spelled out so we know why we're doing this.

Some questions I'd love to have answered:
- Do we have a Conda Enhacement Proposal yet to coordinate the evolution of the repodata cache format?
- Is it needed to make this specific to mamba, or would conda also profit from this?
- Except from a fork that was suggested in https://github.com/conda/constructor/pull/462#issuecomment-1034601126, do we have other options to add this support for micromamba that would make this change future-proof?

constructor/conda_interface.py
chrisburr: @wolfv is probably in the best position to answer.

From my external perspective, assuming that there isn't any conda specification/documentation that I've missed:
* the cache format is an implementation detail that `constructor` is abusing
* either implementation could change how they cache the repodata at any time

The right solution would probably be to instead install from a local channel and let conda/mamba take care of handling that however they please (i.e. `conda.exe install -c file://$PREFIX/repodata.json ...`).

wolfv: We actually do use the same format that conda uses (which is writing a bit of metadata about the download at the beginning of the json file, namely `etag`, `timestamp` and `url` or something along those lines).

I think we maybe should ignore missing etag/... if we are in offline mode. Would that help @chrisburr ?

wolfv: Also for the filename we're supposedly using the same algorithm as conda, which is `MD5(url)[0:8]` or something along those lines.
It's possible that the URL we're using here is different for conda vs. mamba.

chrisburr:
Thank you @jezdez for the review and @wolfv for clearing up how the cache is supposed to work. We should be able to get this working without requiring any mamba specific code.

I'm going to close this PR in favour of it's components: #489, #490 and https://github.com/mamba-org/mamba/pull/1482

After we've got these three merged I'll look at adding a test for `micromamba`+`constructor` to the CI.

PR #489, opened 2022-02-10
Fixes for transmuting packages and generating repodata

chrisburr:
This cherry-picks some commits from #487 while avoiding the mamba-specific commits.

Currently constructor writes incorrect repodata which:
* assumes the host platform's `subdir` is the same as the target platform's `subdir`
* includes incorrect file hashes in the repodata
* ignores if packages have been transmuted to `.conda` (`packages` -> `packages.conda`)
* ignores the fact package filenames change when transmuting

`conda-standalone` is ran with enough checks disabled that this doesn't matter however when running with `micromamba` instead it recognises the issues and fails to install.

### DIFF
--- a/constructor/conda_interface.py
+++ b/constructor/conda_interface.py
@@ -2,8 +2,12 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import json
-from os.path import join
+import os
 import sys
+from copy import deepcopy
+from os.path import join
+
+from constructor.utils import hash_files
 
 NAV_APPS = ['glueviz', 'jupyterlab', 'notebook',
             'orange3', 'qtconsole', 'rstudio', 'spyder', 'vscode']
@@ -87,10 +91,9 @@ if conda_interface_type == 'conda':
 
         return full_repodata
 
-    def write_repodata(cache_dir, url, full_repodata, used_packages):
+    def write_repodata(cache_dir, url, full_repodata, used_packages, info):
         used_repodata = {k: full_repodata[k] for k in
                          set(full_repodata.keys()) - {'packages', 'packages.conda', 'removed'}}
-        repodata_filename = _cache_fn_url(used_repodata['_url'].rstrip("/"))
         used_repodata['packages.conda'] = {}
         used_repodata['removed'] = []
         # arbitrary old, expired date, so that conda will want to immediately update it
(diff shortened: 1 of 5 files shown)


constructor/fcp.py
```diff
@@ -336,14 +337,24 @@ def _main(name, version, download_dir, platform, channel_urls=(), channels_remap
...
-                    transmute_file_type, out_folder=download_dir)
+                if not os.path.exists(new_file_name):
+                    print("transmuting %s" % dist)
+                    failed_files = conda_package_handling.api.transmute(
+                        os.path.join(download_dir, dist),
+                        transmute_file_type,
+                        out_folder=download_dir,
+                    )
+                    if failed_files:
+                        raise RuntimeError("Transmution failed:", failed_files)
```
jezdez, line 348: Are we catching the arguments passed here somewhere or is this something that users will see? If the latter just printing the failed files like they are returned could be a bit messy UX wise. This is a dictionary or something? Anything we can do to list out the failed files nicely?

constructor/fcp.py
```diff
@@ -336,14 +337,24 @@ def _main(name, version, download_dir, platform, channel_urls=(), channels_remap
...
+                url, md5 = _urls[dist]
+                assert url.endswith(".tar.bz2")
```
jezdez, line 350: This is code smell since assert statements are easily disabled in production environments and shouldn't be used for validation. If you want this to fail loudly, please add a condition and fail with a RuntimeError like above.

constructor/conda_interface.py
```diff
@@ -87,20 +91,41 @@ def get_repodata(url):
...
+            # If we're transcoding packages, fix-up the metadata
+            if package.endswith(".conda"):
+                original_package = package[:-len(".conda")] + ".tar.bz2"
+                original_key = "packages"
+            elif package.endswith(".tar.bz2"):
+                original_package = package[:-len(".tar.bz2")] + ".conda"
+                original_key = "packages.conda"
+            else:
+                print("WARNING: package type is unknown for: %s" % package)
+                continue
```
jezdez, line 120: Shouldn't this fail loudly since that implies the metadata is corrupt?

jezdez (changes requested):
Hey thanks @chrisburr for the new PR, this is really appreciated. I'm trying to wrap my head around how this could have been overlooked in the past, but no matter what, this'll need updates to the test so it doesn't happen again.

Could you work on some to cover these cases?

Thank you for your contribution!

constructor/fcp.py
chrisburr: Users would see this, before this change you get a bizzare `FileNotFound` error at some much later point.

And `failed_files` is a dictionary of `path -> failure reason`.

I suspect this will only happen for issues like read-only filesystems or out of disk space so the failure reason is pretty helpful. (For miniforge we hit this with a read only bind mount and https://github.com/conda/conda-package-handling/issues/44.)

constructor/fcp.py
chrisburr: I'm pretty sure this is impossible and even if it fails I'm fairly sure the code would still work. If you'd rather not have `assert`s I propose drop it entirely:
```suggestion
```

chrisburr:
> I'm trying to wrap my head around how this could have been overlooked in the past, but no matter what, this'll need updates to the test so it doesn't happen again.

I think it's just that it wasn't need to get conda working (as conda skips as many sanity checks as possible when used for `constructor`).

> Could you work on some to cover these cases?


I'm not sure how useful it is to write more fine graned tests for generating the cache as ultimately the check is, "does the generated installer work?". I'm planning to add tests for `micromamba` to the CI and maybe something with building for other platforms if it isn't covered already. I think this is better as a sparate PR (partly because it would depend on https://github.com/conda/constructor/pull/490).

constructor/fcp.py
jezdez: Hm, let's try to format the `failed_files` a bit and return a good error message then.

constructor/fcp.py
jezdez: Yeah, unfortunately there are a few environments in which people specify [PYTHONOPTIMIZE](https://docs.python.org/using/cmdline.html#envvar-PYTHONOPTIMIZE) or `-O` in their build environment, which disables assert statements. It's quite the bummer. Let's remove the line then.

constructor/conda_interface.py
jezdez: > I'm in favour of that, how about:
> 
> ```diff
> -                print("WARNING: package type is unknown for: %s" % package)
> -                continue
> +                raise NotImplementedError("Package type is unknown for: %s" % package)
> ```

Looks good to me!

constructor/fcp.py
chrisburr: As there is no reasonable way of testing this in CI (for testing I'm exploiting the bug which is fixed by https://github.com/conda/conda-package-handling/pull/91), this is how the message looks:

```python
RuntimeError: Transmution failed:
    /home/cburr/miniconda3/pkgs/binutils_linux-64-2.36-hf3e587d_5.tar.bz2 failed with: [Errno 13] Permission denied: '/tmpf1pe2tay'
    /home/cburr/miniconda3/pkgs/gfortran_linux-64-9.4.0-hf0ab688_5.tar.bz2 failed with: [Errno 13] Permission denied: '/tmp5r2tkblz'
    /home/cburr/miniconda3/pkgs/constructor-3.2.2-py310hff52083_1.tar.bz2 failed with: [Errno 13] Permission denied: '/tmpqklcfzz5'
    /home/cburr/miniconda3/pkgs/fortran-compiler-1.3.0-h1990efc_0.tar.bz2 failed with: [Errno 13] Permission denied: '/tmpq1v0t3sb'
```

jezdez: Heh, thanks for filing that bug!

jezdez:
@chenghlee Is there anything you see out of the ordinary here?

jezdez (approved):

PR #490, opened 2022-02-10
Include cache metadata on the first line of the repodata cache

chrisburr:
Currently when constructor writes the repodata cache for conda it neglects to add some of the cache headers. Conda is forgiving about this however micromamba is stricter about this and I think it makes sense to make constructor more correct rather than making mamba more lenient.

For reference this is where conda adds these fields to the repodata for `pkgs/cache/*.json`:

https://github.com/conda/conda/blob/05b3e799daba158b74c892bf3aeb60ffe7485e3c/conda/core/subdir_data.py#L822-L834

### DIFF
--- a/constructor/conda_interface.py
+++ b/constructor/conda_interface.py
@@ -90,17 +90,27 @@ if conda_interface_type == 'conda':
     def write_repodata(cache_dir, url, full_repodata, used_packages):
         used_repodata = {k: full_repodata[k] for k in
                          set(full_repodata.keys()) - {'packages', 'packages.conda', 'removed'}}
-        repodata_filename = _cache_fn_url(used_repodata['_url'].rstrip("/"))
         used_repodata['packages.conda'] = {}
         used_repodata['removed'] = []
-        # arbitrary old, expired date, so that conda will want to immediately update it
-        # when not being run in offline mode
-        used_repodata['_mod'] = "Mon, 07 Jan 2019 15:22:15 GMT"
         used_repodata['packages'] = {
             k: v for k, v in full_repodata['packages'].items() if v['name'] in NAV_APPS}
         for package in used_packages:
             for key in ('packages', 'packages.conda'):
                 if package in full_repodata.get(key, {}):
                     used_repodata[key][package] = full_repodata[key][package]
+
+        # The first line of the JSON should contain cache metadata
+        # Choose an arbitrary old, expired date, so that conda will want to
+        # immediately update it when not being run in offline mode
+        url = used_repodata.pop('_url').rstrip("/")
+        repodata = json.dumps(used_repodata, indent=2)
+        repodata_header = json.dumps(
+            {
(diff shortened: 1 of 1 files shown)


constructor/conda_interface.py
```diff
@@ -90,17 +90,28 @@ def get_repodata(url):
...
+        # The first line of the JSON should contain cache metadata
+        # Choose an arbitrary old, expired date, so that conda will want to
+        # immediately update it when not being run in offline mode
+        url = used_repodata.pop('_url').rstrip("/")
+        repodata = json.dumps(used_repodata, indent=2)
+        repodata_header = json.dumps(
+            {
+                "_mod": "Mon, 07 Jan 2019 15:22:15 GMT",
+                "_url": url,
+                "_cache_control": "public, max-age=1200",
```
wolfv, line 111: it turns out that conda also only writes two of these keys (_mod and _url) when fetching any file from a local channel.

I am adding some handling in mamba to make sure that we can also parse that correctly.

chrisburr: Ah great, I guess that makes sense for local channels. Thanks for dealing with this!

In that case the only thing left for this PR would be to make sure that these keys are on the top line of the file? i.e.

```python
{"_mod": "Mon, 07 Jan 2019 15:22:15 GMT", "_url": url,
  "packaes": {
       ...
  },
  "packaes.conda": {
       ...
  }
}
```

wolfv: yes, mamba currently expect them at the beginning of the file. Is that not the case here?

chrisburr: Prior to this PR, constructor just writes `json.dumps(repodata, indent=2)`.

Conda does it in a way that's consistent with `mamba`.

wolfv: yeah i don't like that ... in that case we'd need to parse the entire file first. Would be nice to correct this in constructor :)

chrisburr: I've amended this so that the `_cache_control` and `_etag` fields aren't written.

wolfv:
This looks great to me.

jezdez (changes requested):
This looks good in theory, but needs tests to make sure, we're not regressing in the future on the "strict" behavior that micromamba expects.

chrisburr:
Once this and #489 are merged I'll add `micromamba` tests to the CI.

I've updated miniforge to use this two PRs and it works with the latest micromamba: https://github.com/conda-forge/miniforge/pull/263

jezdez:
@chrisburr With #489 merged this needs another rebase sadly.

chrisburr:
Thanks for taking care of the rebase (and generally picking up the constructor maintenance).

I'll make a PR with tests over the weekend.

jezdez:
@chrisburr No problem, thanks for working on the patches in the first place!

PR #492, opened 2022-02-14
Allow cleaning of environment variables to be disabled during installation

chrisburr:
When running the installer from a different architecture via qemu and binfmt_misc I find I need to set `LD_LIBRARY_PATH` so that `conda.exe` can find `libz.so`. 

```bash
$ QEMU_LD_PREFIX="$CONDA_ROOT/envs/sysroot_linux-ppc64le/powerpc64le-conda-linux-gnu/sysroot" \
    "$CONDA_ROOT/envs/constructor-linux-ppc64le/standalone_conda/conda.exe" -V
conda.exe: error while loading shared libraries: libz.so.1: cannot open shared object file: No such file or directory
```

The easiest solution I found is to set `LD_LIBRARY_PATH` pointing at a conda environment for ppc64le, i.e.

```bash
$ LD_LIBRARY_PATH="$CONDA_ROOT/envs/sysroot_linux-ppc64le/lib" \
    QEMU_LD_PREFIX="$CONDA_ROOT/envs/sysroot_linux-ppc64le/powerpc64le-conda-linux-gnu/sysroot" \
    "$CONDA_ROOT/envs/constructor-linux-ppc64le/standalone_conda/conda.exe" -V
conda 4.10.3
```

Obviously this is a horrible hack and their are other hacky ways of doing this. Despite this I think it's useful to be able to do these kind of tricks for the handful of people that need them. As such I propose to add an option to the installer so that it the unsetting of environment variables can be disabled.

If there are worries we could:
1. Show a big loud message that here-be-dragons and you should probably remove the option again
2. Hide it from the help menu

Extracted from https://github.com/conda/constructor/pull/486

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -5,13 +5,6 @@
 # PLAT:  __PLAT__
 # MD5:   __MD5__
 
-#if osx
-unset DYLD_LIBRARY_PATH DYLD_FALLBACK_LIBRARY_PATH
-#else
-export OLD_LD_LIBRARY_PATH=$LD_LIBRARY_PATH
-unset LD_LIBRARY_PATH
-#endif
-
 if ! echo "$0" | grep '\.sh$' > /dev/null; then
     printf 'Please run using "bash"/"dash"/"sh"/"zsh", but not "." or "source".\n' >&2
     return 1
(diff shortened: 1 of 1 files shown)


jezdez:
FWIW I'm not sure where this should be filed though given the divergence between the [anaconda](https://github.com/AnacondaRecipes/conda-standalone-feedstock) and the [conda-forge feedstock](https://github.com/conda-forge/conda-standalone-feedstock) for conda-standalone.

chrisburr:
Thanks @isuruf, that's definitely the best way to go.

For conda-forge this will be handled by https://github.com/conda-forge/pyinstaller-feedstock/pull/70 (and we can prevent regressions with https://github.com/conda-forge/conda-standalone-feedstock/pull/32).

PR #496, opened 2022-03-10
FIX: `conda_interface` should be able to handle alpha, beta, rc versions

jaimergp:
The previous approach will fail with non-final conda versions (e.g. 4.12a1). This uses `conda.models.VersionOrder` instead, which has been stable since 2016, so I'd say it's safe to use.

### DIFF
--- a/constructor/conda_interface.py
+++ b/constructor/conda_interface.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 import json
 from os.path import join
+from itertools import chain
 import sys
 
 NAV_APPS = ['glueviz', 'jupyterlab', 'notebook',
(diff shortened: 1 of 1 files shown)


constructor/conda_interface.py
```diff
@@ -16,7 +16,11 @@
                        "with sys.prefix: %s" % sys.prefix)
 
 if conda_interface_type == 'conda':
-    CONDA_MAJOR_MINOR = tuple(int(x) for x in CONDA_INTERFACE_VERSION.split('.')[:2])
+    # This import path has been stable since 2016
+    from conda.models.version import VersionOrder
+
+    _conda_version = VersionOrder(CONDA_INTERFACE_VERSION).version
+    CONDA_MAJOR_MINOR = _conda_version[1][0], _conda_version[2][0]
```
jezdez, line 23:
```suggestion
    # this is ignoring the first element since that's the version epoch (e.g. "1!4.12.0")
    CONDA_MAJOR_MINOR = tuple(itertools.chain.from_iterable(_conda_version))[1:3]
```
I'm not a huge fan of accessing deeply nested structures like this in API interface modules since they can lead to ugly errors when something changes in the version scheme, so let's use `itertools.chain` to flatten the list of lists returned from `VersionOrder` and ignore the epoch.

jezdez: Note, this needs an itertools import at the top of the file.

jezdez (approved):
Just a minor suggestion around accessing version specifiers with a bit more defensive programming.

PR #505, opened 2022-03-14
fix upload path for artifacts

jaimergp:
One more fix for #498 , since apparently the artifacts action doesn't support `../xxx` relative paths.

Checks

- [X] Artifacts are uploaded
- [X] Local constructor is installed in tests
- [X] Packages can also be uploaded

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,8 +19,6 @@ jobs:
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
         pyver: ["3.7", "3.8", "3.9", "3.10"]
-    env:
-      CONDA_BLD_PATH: ${{ runner.temp }}/conda-bld
     steps:
       - name: Print github context
         run: |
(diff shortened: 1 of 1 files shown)


jaimergp:
@jezdez sorry for the extra noise but this should be ready now (and hopefully for good).

jezdez:
There seems to be some overlap with @chrisburr's #503, too

jaimergp:
Is there? I am only touching the workflow file here. It should result in no conflicts.

chrisburr:
Unfortunately I will be touching them as I've found my PR is failing due to the CI testing the latest release rather than the current branch. (`conda install -c local constructor` randomly picks between the `defaults` and `local` channel)

Though it's no trouble for me to rebase on top of this once it's merged.

jaimergp:
Hm, yes, I am seeing this as well now @chrisburr

jaimergp:
Looks like it's working more reliably now, thanks for the tip!

PR #506, opened 2022-03-14
Fix truncated python version if minor has two digits (windows)

### LINKED ISSUE #495: Bug: version of python reported as 3.1 in the windows wizard
Hi,
I've been working on a constructor script for use with cadquery

  * https://github.com/CadQuery/cadquery/pull/1000

we've noticed that in the windows installer it mentions python 3.1 for some reason instead of 3.10
maybe the string is getting truncated
``` yaml
name: cadquery
version: master
installer_type: all

license_file: ../../LICENSE

# Don't add to the system Path
initialize_by_default: false
# Don't set as the default python
register_python_default: false

channels:
  - conda-forge
  - cadquery

write_condarc: true
conda_default_channels:
  - conda-forge
  - cadquery

specs:
  - python 3.10*
  - conda
# Occt is too big to bundle so install during post_install instead
#  - cadquery

post_install: post-install.sh           # [unix]
post_install: post-install.bat          # [win]
```

jaimergp:
Closes #495

### DIFF
--- a/constructor/winexe.py
+++ b/constructor/winexe.py
@@ -59,7 +59,7 @@ def make_nsi(info, dir_path):
         'VERSION': info['version'],
         'COMPANY': info.get('company', 'Unknown, Inc.'),
         'ARCH': '%d-bit' % arch,
-        'PY_VER': py_version[:3],
+        'PY_VER': ".".join(py_version.split(".")[:2]),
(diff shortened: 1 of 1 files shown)


jezdez (approved):
Thanks!

PR #507, opened 2022-03-14
add `initialize_conda` and `register_python` keys to control whether to show these options

jaimergp:
All installers will offer the user to initialize conda (all OS) and register Python as default (Windows only). Some applications might be using `conda` behind the scenes but don't want to expose this to the user, so asking for that kind of options can be strange and surprising if they have never heard about conda before.

This PR adds two keys to prevent this behavior:

* `initialize_conda` (True by default). If false, `conda init` prompts won't appear on `sh` installers. EXEs and PKGs won't show the checkboxes.
* `register_python` (True by default). If false,  EXEs won't offer this option, which was discouraged anyway.


jaimergp:
Blocked by #505

jaimergp:
Local tests:

- [X] PKGs and SHs work on macOS
- [ ] SHs work on Linux
- [ ] EXEs work on Windows

jaimergp:
Fixed locally! Screenshots of the different configs for `register_python` and `initialize_conda`, respectively:

With True / True (default):

image

With False / True:

image

With True / False:

image

With False / False:

image

larsoner (dismissed):

jezdez (approved):

hoechenberger:
Amazing, thank you!!!

PR #508, opened 2022-03-14
enable debugging and include license info in the dict

jaimergp:
Debugging flag was not fully implemented (guarded by `if 0:`, who knows). This PR fixes that and also adds a dictionary with a mapping of package-license-filenames so downstream users can quickly review which licenses are being bundled in the app (some of them DO require showing them in the EULA, specially proprietary packages).

As a result, when run with `--debug`, users will obtain a `pkgs-list.txt` and a `info.json` file in their working directory (added to gitignore too).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ _cenv/
 rever/
 .idea
 docs/build/
+pkgs-list.txt
+info.json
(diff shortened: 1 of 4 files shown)


hoechenberger:
This is an **extremely** useful addition. Any chance we could see this getting merged sometime soon?

jaimergp:
I am going to rework this feature @hoechenberger. It also enables ultra verbosity in conda so it's a bit convoluted. I'll tag you in a new PR.

jaimergp:
See https://github.com/conda/constructor/pull/595

Issue #319, opened 2020-01-07
Failures when "exclude"d packages cause inconsistent environments


amjames:
I am using the Maxiconda example, and the installer is built fine with no errors reported. When the installer is run there is no error reported however the installation is incomplete. The package cache is created but no packages are installed. If I open "show details" during the installation I can see that the environment creation failed because one of the excluded packages is required by another package:

```
Collecting package metadata (current_repodata.json): ...working... done
Solving environment: ...working... failed
Collecting package metadata (repodata.json): ...working... done
Solving environment: ...working... failed

PackagesNotFoundError: The following packages are not available from current channels:

  - matplotlib==3.0.0=py35hd159220_0 -> pyqt=5.9

Current channels:

  - http://repo.anaconda.com/pkgs/main/win-64
  - http://repo.anaconda.com/pkgs/main/noarch
```

You can see that the environment solution failed because matplotlib couldn't be installed due to the missing pyqt dependency (pyqt is excluded in the construct.yaml).

I would expect that either 1) the installer fails to build when the impossible environment is detected or 2) the installation fails when the environment creation fails. The way that the installer reports "Setup was completed successfully" smells like a bug to me. 

How are excluded packages supposed to work? Are the examples out of date?

I am using constructor 3.0.0 on windows 7.

forrestwaters:
Yeah I think that example is out of date. The installers created by constructor will do an offline install, so of course, all packages must be included in the installer for that install to be satisfiable. Perhaps previously you could have a set of packages in the installer, and others downloaded from the internet at installation time to save space in the installers themselves. 

Do you have a use-case where you would want to exclude a package? I'm not aware of a reason why we'd want to support that. Your other question is definitely a bit of a bug - we should handle and report errors on windows better.

amjames:
@forrestwaters There is the most widely applicable use case where someone wants to exclude a large package "X" which is brought in as a dependency of another package "Y", but "Y" can still be used without "X" for example like QT and matplotlib. Although this can be solved by using the matplotlib-base package from conda-forge for example.  

In my particular case what I need to do is remove some packages which we can't have on the machines where we are deploying the installer due to security restrictions. In particular we can't deploy software which uses openssl, QT, and some other libraries with reported vulnerabilities. 

As a work-around I have found that if a package we want in the environment brings in one we can't have as a dependency I can do `conda remove --force` in the post_install script. We then have to manually test if we can use that package without the dependency make sure it still works.
[...]

PR #509, opened 2022-03-14
add support for multi_env installers

### LINKED ISSUE #359: Allow for the creation of multiple environments with custom names
Would it be possible to update constructor to do something like this:

```yaml
name: Uberconda
version: 1.0
channels:

specs:
    - python 3.5*
    - conda

envs:
  - name: something_else
    specs:
    - python 3.5*
    - numpy
  - name: _something_private
    specs:
    - python 3.5*
    - pandas

license_file: EULA.txt

# Welcome image for Windows installer
welcome_image: bird.png                  [win]
```

jaimergp:
Closes #359 (possible solution)

***

In this PR I suggest adding a new keyword, `extra_envs`, that allows a user to specify additional environments in the same installer. It has some rules:

* `extra_envs` is a mapping of environment names to environment creation options. The options map supports:
  * `specs`, `channels`, `channels_remap`, `user_requested_specs`, `exclude`: same as the global value
  * `menu_packages` support is planned but not implemented yet (we need the menuinst stuff in place for a cleaner implementation)
* It requires `conda` on `specs` (either explicitly or implicitly, but better be explicit).
* `ignore_duplicate_files` will always be considered `True`

This approach guarantees backwards compatibility at the expense of a slightly more complicated install logic. The other option would have been reworking ALL specs-based installations to follow the `envs` mapping (including `base`). Maybe in a (much needed) refactor in the future.


constructor/construct.py
```diff
@@ -309,6 +333,20 @@
...
+    "specs": (list, tuple),
+    "environment": (str,),
+    "environment_file": (str,),
+    "channels": (list, tuple),
+    "channels_remap": (list, tuple),
+    "user_requested_specs": (list, tuple),
+    "exclude": (list, tuple),
+    # TODO: we can't support menu_packages for extra envs yet
+    # will implement when the PR for new menuinst lands
+    # "menu_packages": (list, tuple),
```
larsoner, line 346: Would this require https://github.com/conda/constructor/pull/474 (and thus https://github.com/conda/conda/pull/11035 and https://github.com/conda/menuinst/pull/91)? I'm having a bit of trouble following... I guess all of https://github.com/conda-incubator/ceps/pull/8 should be merged around the same time to enable "new menuinst"?

jaimergp: Yep. The new `menuinst` support will be trickier to land, as it requires sync across many projects.

larsoner:
This one needs a rebase/merge now

jaimergp:
So I added some tests to ensure things were working ok and I found out that `exclude` and `extra_envs` won't be good friends :(

The way constructor works is the following:

1. The installer unpacks the cache of packages and some synthetic channels repodata that only contain the needed bits.
2. Then it calls conda-standalone with a list of specs that exactly match the existing packages on cache. This is not a list of paths (e.g. `python-3.10-h123456.tar.bz2`, but a list of _specs_ like `python=3.10=h123456`). 
  * This means that `conda-standalone` WILL use the solver, which guarantees that the request is satisfiable and the environment is not shipped broken to begin with.
  * I am not sure why this is needed. I think it has to do with updating installations from an installer, but in some platforms this is actively prevented (Windows, MacOS PKG), so again, not clear.
3. The thing is that if we exclude non-leaf packages (a package that is a dependency of some other package in the same environment), we can't satisfy the solve and it breaks the installation.
4. Additionally, if we mix `extra_envs` and `exclude`, since the synthetic channels and pkg cache will be shared across environments, there's a chance that the solution of one environment provides the packages that were excluded in other envs.
  * For example: environment A ships `python 3.8` from conda-forge but excludes `tk`. Environment B ships `python 3.8` too, but doesn't exclude `tk`. When installing environment A, `tk` will be indeed found in the channel cache because it was needed for environment B. So both A and B will have `tk`, despite the exclusion.

To sum up, what I've done is the following:

* `exclude` can only be specified at the global level.
* If `extra_envs` are in use too, they will inherit the global `exclude` value to ensure they do not undo the exclusion accidentally (as commented in point 4 above).
* Extra envs won't error out if a package that needs to be excluded is not found in them (it does error out for the `base` environment).

***

Additionally, I found out that conda-standalone WILL inherit the user configuration in `~/.condarc` if it existed, which can change the behaviour of the solver (e.g. strict vs flexible priority). This needs to be fixed in conda and conda-standalone; i.e. provide a flag so no `condarc` file is loaded and it really works "standalone". But that's worth its own issue.

jaimergp:
@larsoner Ready for your eyes now :)

larsoner (approved):
I finally read through the code and it looks reasonable to me, and is tested by an example and things are green. Feel free to merge unless you want someone else to look @jaimergp !

jaimergp:
Thanks @larsoner! We'll merge for now. Hopefully we can re-enable the excludes once I figure out the solver issues mentioned in #319

scottwsides:
So the NSIS changes recently to constructor fixed the windows install issue and was working as of a couple of days ago. I just rebuilt... and windows is faiiling again due to "anaconda directories not being initialized". I suspect this commit (since its so recent has broken windows again). I'm buildng from trunk.

jaimergp:
@scottwsides Please create an issue with full details about your input file, full logs and screenshots of the error if necessary. Thanks!

PR #512, opened 2022-03-16
Fix constructor failures when the repo/subchannel only has noarch packages

bryan-hunt:
When the full_repodata is set to None it causes write_repodata to throw an AttributeError exception because it immediately tries to treat it as an initialized dictionary.

Instead of returning none this change returns an initialized repodata dictionary with the minimum expected data which allows constructor to continue properly.


bryan-hunt:
CLA has been signed

jaimergp:
Just ran into this issue too, applied your patch and it worked! I am re-running the CI to make sure everything is alright. Merging later if green!

PR #521, opened 2022-06-22
Fixes uninstaller breaking CMD.exe by corrupting registry keys

guimondmm:
This PR addresses https://github.com/ContinuumIO/anaconda-issues/issues/12938 and https://github.com/ContinuumIO/anaconda-issues/issues/12735.

On 23 Oct 2020, https://github.com/conda/conda/commit/b4f3210bab7819a9a08e3038276cbb2d2498421c changed the regex pattern used by the [`init_cmd_exe_registry`](https://github.com/conda/conda/blob/ae389d58d67dd8e29952e669c34f1495d034185a/conda/core/initialize.py#L1411-L1470) function in [conda/core/initialize.py](https://github.com/conda/conda/blob/master/conda/core/initialize.py) to configure the conda hook for CMD.exe on Windows:
```diff
@@ -1430 +1435 @@ def init_cmd_exe_registry(target_path, conda_prefix, reverse=False):
-            r'(\"[^\"]*?conda[-_]hook\.bat\")',
+            r'(if exist \"[^\"]*?conda[-_]hook\.bat\" \"[^\"]*?conda[-_]hook\.bat\")',
```

This changed the format of the string written by the `conda init` command to either of the following Windows Registry variables:

- HKEY_LOCAL_MACHINE\Software\Microsoft\Command Processor\AutoRun
- HKEY_CURRENT_USER\Software\Microsoft\Command Processor\AutoRun

However, the regex pattern used by the [`rm_regkeys`](https://github.com/conda/constructor/blob/e75e1562d1ffa520b675788f31b241da1f9c72c7/constructor/nsis/_nsis.py#L280-L291) function in [constructor/nsis/_nsis.py](https://github.com/conda/constructor/blob/master/constructor/nsis/_nsis.py) was not updated accordingly.

Because of that oversight, when a Windows user who has run `conda init` after Oct 23, 2020 attempts to uninstall Anaconda using `Uninstall-Anaconda3.exe`, the conda hook isn't deleted properly from the registry by the [constructor/nsis/_nsis.py](https://github.com/conda/constructor/blob/master/constructor/nsis/_nsis.py) script, which breaks CMD.exe and causes system instability.

This is because the old regex pattern `(\"[^\"]*?conda[-_]hook\.bat\")` deleted by [constructor/nsis/_nsis.py](https://github.com/conda/constructor/blob/master/constructor/nsis/_nsis.py) matches twice in the new pattern `(if exist \"[^\"]*?conda[-_]hook\.bat\" \"[^\"]*?conda[-_]hook\.bat\")` set by [conda/core/initialize.py](https://github.com/conda/conda/blob/master/conda/core/initialize.py), and leaves behind `if exist  ` with trailing spaces.

When CMD.exe is launched without the /D flag, it attempts to read the AutoRun variable from the registry, and if it is set, it tries to execute the specified string first. The syntax of the `if else  ` command is invalid, so CMD.exe silently terminates with exit code 1. This breaks the Command Prompt as well as batch files, and causes several apps to crash or misbehave. Windows repair tools like SFC.exe and DISM.exe don't detect the corrupted registry variables, and reinstalling Anaconda doesn't fix the issue. Unless the user knows about obscure CMD.exe flags and registry keys, their system is effectively hosed.

To address the issue, I've updated the regex string in the [`rm_regkeys`](https://github.com/conda/constructor/blob/e75e1562d1ffa520b675788f31b241da1f9c72c7/constructor/nsis/_nsis.py#L280-L291) function of [constructor/nsis/_nsis.py](https://github.com/conda/constructor/blob/master/constructor/nsis/_nsis.py) to match the one in the [`init_cmd_exe_registry`](https://github.com/conda/conda/blob/ae389d58d67dd8e29952e669c34f1495d034185a/conda/core/initialize.py#L1411-L1470) function in [conda/core/initialize.py](https://github.com/conda/conda/blob/master/conda/core/initialize.py).


jaimergp:
@conda-bot check

constructor/nsis/_nsis.py
```diff
@@ -280,7 +280,7 @@ def add_to_path(pyversion, arch):
 def rm_regkeys():
     cmdproc_reg_entry = NSISReg(r'Software\Microsoft\Command Processor')
     cmdproc_autorun_val = cmdproc_reg_entry.get('AutoRun')
-    conda_hook_regex_pat = r'((\s+&\s+)?\"[^\"]*?conda[-_]hook\.bat\")'
+    conda_hook_regex_pat = r'((\s+&\s+)?(if exist)?(\s?\"[^\"]*?conda[-_]hook\.bat\"))'
```
jaimergp, line 283:
```suggestion
    conda_hook_regex_pat = r'((\s+&\s+)?(if +exist)?(\s?\"[^\"]*?conda[-_]hook\.bat\"))'
```

I guess it doesn't hurt to allow more than one space in there?

jaimergp (approved):
Thanks for the detailed investigation and submitted patch @guimondmm!

I think it makes sense. Unfortunately we don't have tests for uninstallers yet, so I'll have to check manually... which involves creating a custom `conda-standalone.exe` 😬 

I'll see if I find the time to do it properly, but for now I trust you and will leave the merge to other members of @conda/constructor if they are more confident!

constructor/nsis/_nsis.py
guimondmm:
```suggestion
    conda_hook_regex_pat = r'((\s+&\s+)?(if +exist)?(\s*?\"[^\"]*?conda[-_]hook\.bat\"))'
```

Based on the regex pattern `r'(if exist \"[^\"]*?conda[-_]hook\.bat\" \"[^\"]*?conda[-_]hook\.bat\")'` in the init script, I don't foresee a situation in which multiple spaces would occur there, but it wouldn't hurt to allow for it.

In that case I'd also account for the possibility of multiple spaces between `exist` and the first `"`.

jaimergp (approved):

jaimergp:
Thank you so much @guimondmm!

YouJiacheng:
Thank you so much!
(but, why not use the same method as conda init --reverse?)

jaimergp:
That's an excellent point. We can investigate how that works in a separate PR. The reason would be that constructor wasn't using a full `conda` until version 3, so the uninstall code probably predates the possibility of using `conda init --reverse`.

AngusMaiden:
This is still an issue with the installers available from https://docs.conda.io/en/latest/miniconda.html
I'm not sure if it's relevant but when I look at the merge commit's associated workflow, the "Upload" action was skipped.

This is a critical issue as is pointed out in the fix, because it can leave the OS severely broken, and should make sure it's not only merged but also uploaded to the installers.

DHowett:
> The syntax of the if exist   command is invalid, so CMD.exe silently terminates with exit code 1. This breaks the Command Prompt as well as batch files, and causes several apps to crash or misbehave.

We just started seeing an uptick in reports of this (hi! I'm the engineering manager for the Windows Console and Terminal (and regrettably, CMD)). Without knowing what the root cause was, I added an early-out to Autorun processing :)

Hopefully this fix plus yours will help remediate a lot of the trouble here. :smile:

image

AngusMaiden:
> The installers you can currently find in that website are built by a separate `constructor` fork. Pinging @pseudoyim for visibility / awareness.

@jaimergp @pseudoyim was this ever fixed such that the old windows installers for Python 3.8 and Python 3.9 (the ones affected by this issue) on the miniconda install page were updated with the fixed version?

jaimergp:
Depends on what you mean by "old". The most recent ones should contain this fix. I think from January 2023 on, but someone else confirm. cc @dbast @marcoesters

AngusMaiden:
> Depends on what you mean by "old". The most recent ones should contain this fix. I think from January 2023 on, but someone else confirm. cc @dbast @marcoesters 

By "old" I just meant the installers for Miniconda for Python 3.8 and 3.9. Because I'm wondering if this fix was applied only to the latest version (for Python 3.10), or were these versions' installers rebuilt and re-uploaded for those older versions as well?

I can and should test myself, however, but I just thought you guys might know the answer already.

jaimergp:
Yea I think they are all built at the same time. You can check the Properties tab of the installer file too: https://conda.github.io/constructor/howto/#find-out-the-used-constructor-version

PR #525, opened 2022-07-07
Removes the usage of `conda._vendor.toolz` from constructor

travishathaway:
This pull request replaces the usage of `toolz` in the main conda project. In an upcoming `conda`, release this will be completely removed.

I chose to completely copy over the code from the `toolz` repository into this one. It now lives in the top level `toolz.py` file. I also ignored the Python 2 compatibility code that was still in the `conda` version and instead updated based on the most recent `toolz` version.

It should also be noted that the `get` function being imported from `conda._vendor.toolz` was not being used and was not ported over.

See this issue for more information: https://github.com/conda/conda/issues/11333


travishathaway:
The test coverage in this repository is pretty low, so I would not rely on the tests passing as the best way to determine this is ready to be merged.

Does anyone know of another way we could test out the changes to make sure they haven't broken anything?

larsoner:
> Does anyone know of another way we could test out the changes to make sure they haven't broken anything?

I guess we could figure out the right commands manually every time to test any changes 😱 But this isn't sustainable, and it's also limited to one OS, etc...

The best way forward I think would be to improve the tests:

1. One way would be via additional pytest code that targets individual functions.

2. Another would be to adapt the code we use to build-install-test the installers on all three OSes in MNE-Python:

    https://github.com/mne-tools/mne-installers/blob/main/.github/workflows/build.yml

3. Really the best combination would probably be some pytest fixture magic that had build-and-install (to tmpdir) capabilities, i.e., something that took what we do in mne-installers and built it into a fixture somehow. For this the closest thing I can think of is the [sphinx_app_wrapper](https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/tests/conftest.py#L212) from sphinx-gallery that makes it easy to do an isolated sphinx build via `sphinx_app_wrapper.create_sphinx_app`:

    https://github.com/sphinx-gallery/sphinx-gallery/blob/7347cd47a8623a89919fb3fd744150a074a49431/sphinx_gallery/tests/test_gen_gallery.py#L37-L41

    We'd have to see how long this took, but if it's only a few or tens of seconds per recipe, then we can try to make a diverse minimal set of recipes to hopefully hit most code paths...

kenodegard (changes requested):
As my comments suggest I'm clearly not a fan of copying code 😄

tests/test_fcp.py
```diff
@@ -0,0 +1,31 @@
...
+        (GenericObject('NameOne'), GenericObject('NameTwo')), 0
+    ),
+    (
+        (GenericObject('NameOne'), GenericObject('NameTwo'), GenericObject('NameThree')), 0
+    ),
+))
+def test_check_duplicates(values: tuple[..., GenericObject], expected_fails: int):
+    with mock.patch('constructor.fcp.sys.exit') as sys_exit:
+        check_duplicates(values)
+        assert len(sys_exit.mock_calls) == expected_fails
```
kenodegard, line 31: `sys.exit` raises `SystemExit`, no need to mock, see https://docs.python.org/3.10/library/sys.html#sys.exit

```python
from contextlib import nullcontext

class _NameFn:
    """We use this for testing the check_duplicates function"""
    def __init__(self, name):
        self.name = name
        self.fn = f'filename/{name}.txt'


@pytest.mark.parametrize('values,expected_fails', (
    ((_NameFn('One'), _NameFn('Two'), _NameFn('Two')), True),
    ((_NameFn('One'), _NameFn('Two')), False),
    ((_NameFn('One'), _NameFn('Two'), _NameFn('Three')), False),
))
def test_check_duplicates_failures(values: tuple[_NameFn, ...], expected_fails: bool):
    if expected_fails:
        context = pytest.raises(SystemExit)
    else:
        context = nullcontext()

    with context:
        check_duplicates(values)
```

constructor/fcp.py
```diff
@@ -44,11 +46,12 @@ def warn_menu_packages_missing(precs, menu_packages):
...
-    for precs in groups.values():
-        if len(precs) > 1:
-            sys.exit("Error: '%s' listed multiple times: %s" %
-                     (precs[0].name, ', '.join(prec.fn for prec in precs)))
+    prec_counts = Counter(prec.name for prec in precs)
+
+    for name, count in prec_counts.items():
+        if count > 1:
+            filenames = (prec.fn for prec in precs if prec.name == name)
+            sys.exit(f"Error: {name} listed multiple times: {' , '.join(filenames)}")
```
kenodegard, line 54: This is what I had been thinking:

```python
for name, precs in groupby(precs, lambda prec: prec.name).items():
    filenames = sorted(prec.fn for prec in precs)
    if len(filenames) > 1:
        sys.exit(f"Error: {name} listed multiple times: {' , '.join(filenames)}")
```

kenodegard (changes requested):
Oops! Sorry for not paying more attention and suggesting these changes during the last review. Most of these `itertools` use cases actually seem unnecessary and can be replaced with arg unpacking and list-comprehensions.

constructor/fcp.py
```diff
@@ -245,15 +249,17 @@ def _main(name, version, download_dir, platform, channel_urls=(), channels_remap
...
         print("specs: %r" % specs)
 
     # Append channels_remap srcs to channel_urls
-    channel_urls = tuple(concatv(
+    channel_urls = (
         channel_urls,
         (x['src'] for x in channels_remap),
-    ))
+    )
+    channel_urls = tuple(chain.from_iterable(channel_urls))
```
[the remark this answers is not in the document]
travishathaway: Definitely like this better 👌

constructor/fcp.py
```diff
@@ -245,15 +249,17 @@ def _main(name, version, download_dir, platform, channel_urls=(), channels_remap
           transmute_file_type=''):
     # Add python to specs, since all installers need a python interpreter. In the future we'll
     # probably want to add conda too.
-    specs = list(concatv(specs, ("python",)))
+    specs = (specs, ("python",))
+    specs = list(chain.from_iterable(specs))
```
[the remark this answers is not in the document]
travishathaway: Yep, totally agree.

tests/test_fcp.py
travishathaway: Cool use of `nullcontext`. I will update the tests.

kenodegard (approved):

jezdez (approved):
Noice!

constructor/fcp.py
```diff
@@ -8,17 +8,19 @@
 """
 from __future__ import absolute_import, division, print_function
 
-from collections import defaultdict
+from collections import defaultdict, Counter
```
kenodegard, line 11: Oh `Counter` is unused now

```suggestion
from collections import defaultdict
```

constructor/fcp.py
```diff
@@ -8,17 +8,19 @@
...
 import json
 import os
 from os.path import isdir, isfile, join, splitext
+from itertools import groupby, chain
```
kenodegard, line 15: `chain` is also unused

```suggestion
from itertools import groupby
```

kenodegard (approved):
Just removing two unused imports

constructor/fcp.py
travishathaway: Thanks. Not used to not having linter!

beeankha (approved):
👍

Issue #526, opened 2022-07-19
[Windows] Python Installer crashed using nsis 3.08


jwu9:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

We created a customized Python toolkit installer using constructor, the installer worked well when using nsis 3.01.
Recently we upgraded nsis to 3.08 which provides the capability to sign the uninstaller exe.
After switched to nsis 3.08(with no any other additional changes), we experienced crashes during installation process.
Here are the high level steps to observe the crashes:
1. Generate the customized Python toolkit installer using constructor (and nsis 3.08);
2. Install the generated installer for the first time, noticed that everything is installed correctly;
3. Uninstall the Python toolkit from "Add & Remove programs";
4. Run the installer again;
5. This time the installation window disappeared after a while, and checked the installation folder, there were just a few folders and the packages were not extracted, the subfolders and files are as below:
```
 conda-meta [folder]
 Lib [folder]
 pkgs [folder]
 _conda.exe [file]
```

6. Checked the task manager, noticed that there are 3 conda processes running under Background processes - even the installation window disappeared;
7. Checked the windows Event Viewer, noticed that there were some faulting errors for the mentioned installer:
```
Faulting application name: PythonToolkitInstallTool.exe, version: 2022.1.0.0, time stamp: 0x614f9b5a
Faulting module name: nsExec.dll, version: 0.0.0.0, time stamp: 0x614f9a06
Exception code: 0xc0000005
Fault offset: 0x00001564
Faulting process id: 0x1ac
Faulting application start time: 0x01d88c6c9f7ae701
Faulting application path: C:\Distr\PythonToolkitInstallTool.exe
Faulting module path: C:\Users\demo\AppData\Local\Temp\nsh4DF0.tmp\nsExec.dll
Report Id: b1533623-83c7-48e8-9778-48d2f9aad5be
Faulting package full name: 
Faulting package-relative application ID: 
```

and information report:

```
Fault bucket 1397276649508340299, type 1
Event Name: APPCRASH
Response: Not available
Cab Id: 0

Problem signature:
P1: PythonToolkitInstallTool.exe
P2: 2022.1.0.0
P3: 614f9b5a
P4: nsExec.dll
P5: 0.0.0.0
P6: 614f9a06
P7: c0000005
P8: 00001564
P9: 
P10: 

Attached files:
[...]

PR #532, opened 2022-07-27
Force use of nsis 3.01 on Windows

jaimergp:
Temporary workaround for the CI issues we are seeing as part of #526

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -29,7 +29,7 @@ requirements:
     - ruamel_yaml >=0.11.14,<0.16
     - conda-standalone
     - pillow >=3.1     # [win]
-    - nsis >=3.01      # [win]
+    - nsis ==3.01      # [win]
 
 test:
   source_files:


scottwsides:
Hi Jaime, I'm only asking this here because of your recent merge of the branding branch. A minor problem first... for a while I've been seeing

```
No local packages or working download links found for ruamel_yaml
error: Could not find suitable distribution for Requirement.parse('ruamel_yaml')
```

which normally just keeps going, but now no pkg file gets made, so I didnt know if this was something new. Also, since branding I see

```
Checking mac signature
pkgutil: option `--check-signature' requires an argument``
```

which was working in your branch. I'm building just by cloning source, going into my conda environment and doing

`python setup.py install`

Is this not the standard way?  Thanks!

jaimergp:
To get the dev version I'd do something like this:

```
$ conda create -n constructor-dev constructor pip
$ conda activate constructor-dev
$ conda remove --force constructor
$ pip install https://github.com/conda/constructor/archive/main.tar.gz --no-deps
```

Try that and let me know if it works!

scottwsides:
Thanks, I will try that. But I want to understand why building main doesnt work. Didnt you merge changes from osx-branding into main? I just want to be prepared for changes later (when I forget what has happened) and my installers might break.

jaimergp:
We don't use that option of `pkgutil` directly so no clue what might be wrong. If the issue persists please open an issue with more details about your installation and input files.

scottwsides:
Sorry.... the pkgutil error is a red-herring. Thats in my script and is dying simply because constructor is failing. The pip install is fine... but I'm getting

phoenix:hpccloud-trunk swsides$ cs-env/bin/constructor --verbose chemistream-constructor
platform: osx-64
conda packages download: /Users/swsides/.conda/constructor/osx-64
Traceback (most recent call last):
  File "cs-env/bin/constructor", line 8, in <module>
    sys.exit(main())
  File "/Users/swsides/hpccloud-trunk/cs-env/lib/python3.8/site-packages/constructor/main.py", line 244, in main
    main_build(dir_path, output_dir=out_dir, platform=args.platform,
  File "/Users/swsides/hpccloud-trunk/cs-env/lib/python3.8/site-packages/constructor/main.py", line 112, in main_build
    fcp_main(info, verbose=verbose, dry_run=dry_run, conda_exe=conda_exe)
  File "/Users/swsides/hpccloud-trunk/cs-env/lib/python3.8/site-packages/constructor/fcp.py", line 398, in main
    _urls, dists, approx_tarballs_size, approx_pkgs_size, has_conda = _main(
  File "/Users/swsides/hpccloud-trunk/cs-env/lib/python3.8/site-packages/constructor/fcp.py", line 251, in _main
    print("specs: %r" % specs)
TypeError: not all arguments converted during string formatting

Again... nothing else about my construct.yaml has changed that works with your osx-branding.

jaimergp:
That looks like your `construct.yaml` has some kind of issue in the `specs` field. Can you post the contents of that file here?

scottwsides:
So I switched back to your branch... and I'm still getting same errors. The first error is the ruamel_yaml 'Parse distribution' error. I dont think anything else in my build script has changed. So I cant build at all on mac now... which is kinda serious.

jaimergp:
Can you DM me on gitter? http://gitter.im/jaimergp

scottwsides:
Sure... I'm on gitter now, I think. I havent used gitter before.

PR #534, opened 2022-07-27
Fix string conversion issue in verbose mode

jaimergp:
If `--verbose` is enabled, `constructor` will print more stuff to STDOUT. However, under certain circumstances, the `specs` tuple will have non `%r`-formattable objects, which makes everything fail horribly. Replacing this `print()` with the default syntax provides the same verbosity requirements while being safe.


larsoner:
Seems reasonable to me, thanks @jaimergp !

PR #535, opened 2022-07-27
Use proper logging mechanism instead of notify on macOS

hoechenberger:
The notifications often go unnoticed if a focus mode is active.

And if they don't go unnoticed, we found they're irritating to users, as clicking on them spawns a script editor.

Lastly, the emitted messages are not logged inside the installer.

To resolve these issues, the changes here replace all `notify` calls with proper logger invocations.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,10 +6,6 @@ XXXX-XX-XX   X.X.X:
 Enhancements:
 -------------
 
-* PKG installers now default to the macOS logging system to log messages.
-  UI notifications can be enabled with `progress_notifications`
-  (off by default). (#535)
-
 * The graphical macOS installer now also displays the version number of the
   software in the window title. (#536)
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,4 +1,15 @@
 .. current developments
+
+XXXX-XX-XX   X.X.X:
+===================
+
+Enhancements:
+-------------
+
+* The graphical macOS installers now use the macOS logging system to log
+  messages. Previously, notification badges would be displayed. (#535)
+
+
 2022-03-14   3.3.1:
 ===================
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,8 +6,9 @@ XXXX-XX-XX   X.X.X:
 Enhancements:
 -------------
 
-* macOS installers now use the macOS logging system to log
-  messages. Previously, notification badges would be displayed. (#535)
+* PKG installers now default to the macOS logging system to log messages.
+  UI notifications can be enabled with `progress_notifications`
+  (off by default). (#535)
 
 * The graphical macOS installer now also displays the version number of the
   software in the window title. (#536)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@ XXXX-XX-XX   X.X.X:
 Enhancements:
 -------------
 
-* The graphical macOS installers now use the macOS logging system to log
+* macOS installers now use the macOS logging system to log
   messages. Previously, notification badges would be displayed. (#535)
 
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,9 @@ Enhancements:
 * macOS installers now use the macOS logging system to log
   messages. Previously, notification badges would be displayed. (#535)
 
+* The graphical macOS installer now also displays the version number of the
+  software in the window title. (#536)
+
 
 2022-03-14   3.3.1:
 ===================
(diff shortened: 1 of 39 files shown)


hoechenberger:
I've signed the CLA. It appears it needs to be manually verified first?

kenodegard:
@conda-bot check

kenodegard:
@conda-bot check

CHANGELOG.rst
```diff
@@ -1,4 +1,15 @@
 .. current developments
+
+XXXX-XX-XX   X.X.X:
+===================
+
+Enhancements:
+-------------
+
+* The graphical macOS installers now use the macOS logging system to log
```
hoechenberger, line 9:
```suggestion
* macOS installers now use the macOS logging system to log
```

jaimergp:
I added the notifications because we got feedback from users that thought the installer had hung. I get your point about intrusiveness and lack of logs, so I'd propose if we can make this an option? Something like:

```yaml
osx_pkg_progress_notifications: true  # default is false
```

The normal `echo` calls are logged too, by the way. I am not sure if using `logger` is advantageous in other ways, though.

hoechenberger:
> The normal `echo` calls are logged too, by the way. I am not sure if using `logger` is advantageous in other ways, though.

Are you sure about that? IIRC only with those logger calls I could properly see all messages in the GUI installer log window. But I'd have to check again.

hoechenberger:
> I added the notifications because we got feedback from users that thought the installer had hung. I get your point about intrusiveness and lack of logs, so I'd propose if we can make this an option? Something like:
> 
> ```yaml
> osx_pkg_progress_notifications: true  # default is false
> ```

I've received user feedback that the notifications were confusing.

It would be great if we could find a way to display "proper" notifications somehow and not have to abuse AppleScript for this stuff…

jaimergp:
@hoechenberger Would it be ok if I push some changes to your branch introducing the configuration option to enable notifications (off by default)? This would be the ideal case for a plugin, but we don't have support for that (maybe in the future?), so one more config key would have to do now... If you prefer I open a 2nd PR as an alternative implementation, that's ok too!

hoechenberger:
@jaimergp Please feel free to push changes here, I'm currently busy with other stuff and therefore didn't have the time to come back to this PR

jaimergp (dismissed):
I'd like to see this as an opt-in setting (notifications off by default). I'll implement it next week, please don't merge yet :)

CONSTRUCT.md
```diff
@@ -380,6 +380,16 @@ available and set to the destination causing the error. Defaults to:
...
 
+## `progress_notifications`
+
+_required:_ no<br/>
+_type:_ boolean<br/>
+Whether to show UI notifications on PKG installers. On large installations,
+the progress bar reaches ~90% very quickly and stays there for a long time.
+This might look like the installer froze. This option enables UI notifications
+so the user receives updates after each command executed by the installer.   
+(MacOS only)
```
hoechenberger, line 391:
```suggestion
(macOS only) 
```

constructor/construct.py
```diff
@@ -298,6 +298,14 @@
...
 
 (MacOS only)
+'''),
+
+    ('progress_notifications', False, bool, '''
+Whether to show UI notifications on PKG installers. On large installations,
+the progress bar reaches ~90% very quickly and stays there for a long time.
+This might look like the installer froze. This option enables UI notifications
+so the user receives updates after each command executed by the installer.   
+(MacOS only)
```
hoechenberger, line 308:
```suggestion
(macOS only) 
```

hoechenberger:
I think new commits should always refer to `macOS` (which is the correct spelling). We should do a separate PR to fix existing spelling…

CHANGELOG.rst
```diff
@@ -6,6 +6,9 @@ XXXX-XX-XX   X.X.X:
 Enhancements:
 -------------
 
+* macOS installers now use the macOS logging system to log
+  messages. Previously, notification badges would be displayed. (#535)
```
jaimergp, line 10:
```suggestion
* PKG installers now default to the macOS logging system to log messages.
  UI notifications can be enabled with `progress_notifications`
  (off by default). (#535)
```

jaimergp:
Hm, we have a `news/` folder but also a `CHANGELOG.rst`. I think the CHANGELOG is only updated by the release automations though, so we should turn the changes in CHANGELOG into a news piece :D I'll go for it.

CHANGELOG.rst
```diff
@@ -6,6 +6,10 @@ XXXX-XX-XX   X.X.X:
 Enhancements:
 -------------
 
+* PKG installers now default to the macOS logging system to log messages.
+  UI notifications can be enabled with `progress_notifications`
+  (off by default). (#535)
+
```
jaimergp, line 12:
```suggestion
```

hoechenberger (commented):
Very nice. I cannot approve since I was the one who originally opened this PR.

larsoner (approved):
Thanks @hoechenberger @jaimergp !

PR #543, opened 2022-08-12
Hardcode paths to Apple's system tools

jaimergp:
### Description

We ran into this issue in the https://github.com/napari/packaging/issues/11. Apparently some package in the dependency tree includes a `codesign` binary too, with a different CLI. This can get in the way of our signing process, resulting in a 109 error.

This happened because we let `subprocess` pick the first `codesign` in PATH. The fix is to hardcode the location to `/usr/bin/codesign`, as [`conda` itself does](https://github.com/conda/conda/blob/fdbedc49d4a61fb2121614de22795cfccc562b0b/conda/core/portability.py#L67).

I also applied the same patch to `productbuild` and `productsign`, just in case.

### Checklist - did you ...

### DIFF
--- a/constructor/osxpkg.py
+++ b/constructor/osxpkg.py
@@ -374,8 +374,9 @@ def create(info, verbose=False):
             }
             plist_dump(plist, f)
         check_call(
-            [
-                "/usr/bin/codesign",
+            [   
+                # hardcode to system location to avoid accidental clobber in PATH
+                "/usr/bin/codesign",  
                 "--verbose",
                 '--sign', notarization_identity_name,
                 "--prefix", info.get("reverse_domain_identifier", info['name']),
@@ -411,7 +412,8 @@ def create(info, verbose=False):
     # The default distribution file needs to be modified, so we create
     # it to a temporary location, edit it, and supply it to the final call.
     xml_path = join(PACKAGES_DIR, 'distribution.xml')
-    args = ["productbuild", "--synthesize"]
+    # hardcode to system location to avoid accidental clobber in PATH
+    args = ["/usr/bin/productbuild", "--synthesize"]
     for name in names:
         args.extend(['--package', join(PACKAGES_DIR, "%s.pkg" % name)])
     args.append(xml_path)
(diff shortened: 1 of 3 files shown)


isuruf (dismissed):

constructor/osxpkg.py
```diff
@@ -375,7 +375,7 @@ def create(info, verbose=False):
             plist_dump(plist, f)
         check_call(
             [
-                'codesign',
+                "/usr/bin/codesign",
```
goanpeca, line 378: Maybe add a comment here, with reference to the issue or why we hardcode?

constructor/osxpkg.py
```diff
@@ -420,15 +420,15 @@ def create(info, verbose=False):
 
     identity_name = info.get('signing_identity_name')
     check_call([
-        "productbuild",
+        "/usr/bin/productbuild",
```
goanpeca, line 423: Maybe add a comment here, with reference to the issue or why we hardcode?

constructor/osxpkg.py
```diff
@@ -420,15 +420,15 @@ def create(info, verbose=False):
...
         "--distribution", xml_path,
         "--package-path", PACKAGES_DIR,
         "--identifier", info.get("reverse_domain_identifier", info['name']),
         "tmp.pkg" if identity_name else info['_outpath']
     ])
     if identity_name:
         check_call([
-            'productsign', '--sign', identity_name,
+            '/usr/bin/productsign', '--sign', identity_name,
```
goanpeca, line 431: Maybe add a comment here, with reference to the issue or why we hardcode?

goanpeca (approved):

goanpeca:
Maybe we can use a variable for the PREFIX. 

```python
APPLE_TOOLS_PREFIX + "/codesign"
```

or something 🤷🏼

hoechenberger (dismissed):
The changes look good to me.

hoechenberger (approved):

jaimergp:
Thanks @hoechenberger - this automated review dismissal is a bit inconvenient :(

hoechenberger:
Yep, can this be disabled?

We could also set up automerge for when reviews are in and CI passes. WDYT?

jaimergp:
Agree, that'd be nice. I think we need to ask someone with more permissions though. Let's create an issue.

PR #546, opened 2022-08-15
Allow shortcut removal on uninstall even if `.conda/environments.txt` does not exist

### LINKED ISSUE #461: The `.conda` directory is not being created on Windows
When trying to build a Windows installer (win-64, tested on Windows 10), it was found that the uninstaller wasn't removing menu entries but only in some Windows environments and not others. After some investigation, it turned out that the environments where the uninstaller *does* work properly are the ones where a `.conda` directory already existed due to a previous Miniconda/Anaconda installation.

The problem seems to be that the `.conda` directory (usually in `$HOME`) is not created by the installers built by constructor. In turn, it seems that the issue with the uninstaller (i.e. not removing menu items) goes away when the available environments can be found in the `environments.txt` file of the `.conda` directory (when present).

In conclusion, constructor built installers do not appear to be creating the `.conda` directory when they should while the vanilla Miniconda/Anaconda installers have no such issue. I've discussed this with @chenghlee and the root cause of this discrepancy is currently being investigated.

jaimergp:
### Description

Fixes #461 

### Checklist - did you ...

### DIFF
--- a/constructor/nsis/_nsis.py
+++ b/constructor/nsis/_nsis.py
@@ -158,9 +158,8 @@ get_conda_envs = get_conda_envs_from_python_api
 
 def rm_menus(prefix=None, root_prefix=None):
     try:
-        import menuinst
+        import menuinst  # noqa
         from conda.base.context import context
-        menuinst
     except (ImportError, OSError):
         return
     try:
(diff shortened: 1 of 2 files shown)


larsoner (approved):
This looks straightforward as well, thanks @jaimergp !

PR #558, opened 2022-08-22
Make pre/post install scripts UX more consistent

jaimergp:
### Description

#491 exposed some metadata about the installer so the pre and post install scripts on SH installers could use that information. However, these values are not available for scripts on PKG and EXE installers.

Also, EXE and PKG installers do not have pre-install scripts.

This PR will:

- [X] Make installer metadata available for all installer types
- [X] Enable pre install scripts on Windows EXEs
- [x] Enable pre install scripts on MacOS PKGs
- [X] Enable pre-install script descriptions, which control whether the user can disable its execution or not.
- [X] Rework macOS PKG inner structure to insert pre-install logic after tarball extraction
- [x] Wrap user-supplied scripts with some parent script for better UX on PKG installers
- [x] Test all of these changes work
- [x] Update docs
- [x] Merge #560 for proper Windows tests

Note I had to rephrase some of the checkbox descriptions on Windows so all the extra checkboxes fit (sorry for the potato quality, VM over browser is not excellent):

image

MacOS screenshot:

image

### Checklist - did you ...


FaustinCarter:
This is fantastic!

larsoner:
Looks like there is now a conflict to resolve @jaimergp

jaimergp:
Ready, @larsoner!

CONSTRUCT.md
```diff
@@ -318,31 +318,40 @@ Defaults to `${NAME} ${VERSION} (Python ${PYVERSION} ${ARCH})`.
...
+etc). See `post_install` for information about available environment variables.
+
+## `pre_install_desc`
+
+_required:_ no<br/>
+_type:_ string<br/>
+A description of the purpose of the supplied pre_install script. If this
+string is supplied and non-empty, then the Windows and macOS GUI installers
+will display it along with checkbox to enable or disable the execution of the
+script. If this string is not supplied, it is assumed that the script
```
larsoner, line 333: And what will happen on Linux, a Y/N prompt, or nothing?

larsoner (commented):
Looks good to me other than one small point of clarification

CONSTRUCT.md
jaimergp: Nothing. These are meant to be used in the GUI installers, so they are not used at all in the shell installers. For the `sh` installers, users can disable scripts via the `--skip-scripts` command-line flag. Maybe we can clarify this in the docs. Thanks for the suggestion!

PR #553, opened 2022-08-15
catch install-time errors on windows

jaimergp:
### Description

Closes #544 

Closes #551 

In #509 I forgot to add one piece of the conda packages extraction logic. This resulted in failed installs. Readding this piece fixes #544.

However, this error wasn't caught by the CI pipeline because of #551!

Note that the silent mode of NSIS installers (`/S`) does _not_ print anything to STDOUT. Additionally, most of the dialogs are configured to "ignore" the warning if an error occurred, thus moving on. The exit code does not change with errors! We can't rely on that to detect whether something failed. Instead, the solution was:

* The CI now uses a special NSIS build with log-dumping abilities (normal NSIS does not have this feature). I want to add [`*log*` variants](https://github.com/conda-forge/nsis-feedstock/pull/21) to the `nsis` package in conda-forge/defaults, but for now we are manually patching.
* Errors are printed to the log file, even if ignored
* `conda install ...` step will throw an error too, if needed (it didn't before!)
* The CI will inspect the log for `:error:` strings planted by our NSI script
* If there were any errors, these are printed and reported the installer is reported as failing.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [X] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #560, opened 2022-08-30
Improve testing on Windows

jaimergp:
### Description

(Comes from #449 and extends #553)

We had some issues with the Windows tests:
  - The NSIS installers always exit with a `0` exit code, regardless it worked or failed. Not a good test 😬 
  - The silent mode `/S` on Windows is REALLY silent. Nothing (and I mean _nothing_) gets printed to stdout or stderr. If something fails, it generally keeps going on (the error dialogs are configured to "ignore" in silent mode; check the `/SD` parameter next to each `MessageBox`). 
  - To make things worse, the `conda install` command was _not_ reporting its possible errors via a dialog, so I changed this to throw an error message that aborts the installation. Again, remember, none of these errors get reported to stdout or stderr.
  - What if we manually print to a file then? Enabling log files with _some_ of the output required using a special build for NSIS, which was added in #553. We then slightly modified our error messages to always include the `::error::` sentinel value, which we can `grep` out of the logfile _after_ the installation has concluded. If we find this sentinel value, then we tell `run_examples.py` to report that installation as failed.
  - Note that these errors won't include the output from the failing command. That output will only appear on the GUI logs due to `nsExec` limitations.  In other words, CI will only report whether the error occurred, but further debugging needs to be done locally on a Windows machine with desktop and graphical installer.
  - Another note: errors due to path issues (spaces if enabled, unicode, permissions, etc) won't appear in the logfile because the destination path does not exist yet at that. Hence, a `FileNotFoundError` could mean an issue with the paths.
  - Lastly, `_nsis.py` was not reporting the error code of the scripts it was running (pre-uninstall or post-install). This might have been done on purpose, so I didn't want to change that default behaviour. Instead, it will only report subprocess errors (via exit code) if a special environment variable is set. We are setting it in our CI, since end users might not need it. That said, I am open to changing this conservative decision and leave it as "always tell me if something went wrong".
  - 
### Checklist - did you ...

### DIFF
--- a/constructor/construct.py
+++ b/constructor/construct.py
@@ -158,12 +158,23 @@ default will determined by the `name`, `version`, platform, and installer type.
     ('installer_type',     False, (str, list), '''
 The type of the installer being created.  Possible values are:
 - `sh`: shell-based installer for Linux or macOS;
-- `pkg`: macOS GUI installer
-- `exe`: Windows GUI installer
+- `pkg`: macOS GUI installer built with Apple's `pkgbuild`
+- `exe`: Windows GUI installer built with NSIS
 
 The default type is `sh` on Linux and macOS, and `exe` on Windows. A special
 value of `all` builds _both_ `sh` and `pkg` installers on macOS, as well
 as `sh` on Linux and `exe` on Windows.
+
+Notes for silent mode `/S` on Windows EXEs: 
+- NSIS Silent mode will not print any error message, but will silently abort the installation.
+  If needed, [NSIS log-builds][nsis-log] can be used to print to `%PREFIX%\\install.log`, which can be 
+  searched for `::error::` strings. Pre- and post- install scripts will only throw an error
+  if the environment variable `NSIS_SCRIPTS_RAISE_ERRORS` is set.
+- The `/D` flag can be used to specify the target location. It must be the last argument in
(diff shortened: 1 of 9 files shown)


constructor/construct.py
```diff
@@ -158,12 +158,23 @@
...
 value of `all` builds _both_ `sh` and `pkg` installers on macOS, as well
 as `sh` on Linux and `exe` on Windows.
+
+Notes for silent mode `/S` on Windows EXEs: 
+- NSIS Silent mode will not print any error message, but will silently abort the installation.
+  If needed, [NSIS log-builds][nsis-log] can be used to print to `%PREFIX%\\install.log`, which can be 
+  searched for `::error::` strings. Pre- and post- install scripts will only throw an error
+  if the environment variable `NSIS_SCRIPTS_RAISE_ERRORS` is set.
+- The `/D` flag can be used to specify the target location. It must be the last argument in
+  the command and should NEVER be quoted, even if it contains quotes. For example:
```
hoechenberger, line 174: Typo or intentional?
```suggestion
  the command and should NEVER be quoted, even if it contains spaces. For example:
```

jaimergp: Good catch! Thanks!

goanpeca (approved):
Looks great @jaimergp !

PR #569, opened 2022-09-12
test uninstallers

jaimergp:
### Description

We were not testing the uninstallers generated on Windows, so at least I want to make sure we can run them and the files are (mostly) deleted. Other side effects are not tested (like registry cleanup or menu item removal). Maybe in another PR, after maybe revamping the testing infrastructure so we use pytest or similar.

### Checklist - did you ...

### DIFF
--- a/news/569-test-uninstallers
+++ b/news/569-test-uninstallers
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* Run Windows uninstallers as part of the examples CI (#569)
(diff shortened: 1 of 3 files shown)


jaimergp:
We are running into some IO lock issues where `ExecWait` leaves a Python process open but returns control the uninstaller, so the temporary directory cleanup fails. We have seen reports of this problem in some issues, but hopefully #558 alleviates this problem a bit (in that PR we switched from `ExecWait` to `nsExec` which is a bit more robust).

jaimergp:
Actually it wasn't `ExecWait` but the self-copy mechanism of the uninstaller, which returns control immediately. Using `_?=<path>` prevents the copy and hence it doesn't return immediately.

Ready for review!

larsoner (approved):
Thanks @jaimergp

PR #570, opened 2022-09-12
Fix tk grep tests

jaimergp:
### Description

In the `extra_envs` example we check that we can indeed exclude `tk` from the installed packages in the `base` environment. For that we pipe `conda list` into `grep` and check if we match a line with `tk`. However the query can also match the temporary path of the environment if we are unlucky enough to have a `$TMP` that contains `*tk*` in it. This the reason why some tests are failing in the opened PRs. To fix it, we will instead use `--json` plus `jq` to find keys more robustly.

### Checklist - did you ...


goanpeca (approved):
Looks good @jaimergp 🚀

PR #582, opened 2022-11-10
Display extra pages in Windows installers

pseudoyim:
### Description

This PR enables the adding of extra pages to the Windows GUI installer. These extra pages can display text, links, and images. Such pages can be useful in providing additional instructions, guidance, or promotional materials for end-users before they begin using the application they just installed.

To use this feature, you will need to:
1. Create a `.txt` file that includes the block of NSIS code for your extra page(s).
2. In your `construct.yaml`, set the `extra_pages_file` key to point to this `.txt` file.
3. (Optional) In your `construct.yaml`, under the `extra_pages_image_files` key, list any image files that should be used in the extra pages.

Please see screenshots below for a demonstration of the new feature.

The example installer built from `examples/extra_pages` (named `ExtraPages 1.0.0`) includes two extra pages. The initial welcome page looks like this:
Screen Shot 2022-11-10 at 3 14 07 PM

The installation completes as expected.
Screen Shot 2022-11-10 at 3 19 15 PM

The first extra page displays text, a link, and an image.
Screen Shot 2022-11-10 at 3 19 28 PM

The second extra page displays text and checkboxes for two links that, when selected, will open in the user's browser after they click "Finish".
Screen Shot 2022-11-10 at 3 20 11 PM

### Checklist - did you ...


pseudoyim:
@jaimergp - following up on our offline discussion on how to handle `extra_pages` and `conclusion_text`. Do I have this right?
- If you specify an `extra_pages` file, you are also required to provide your own `MUI_FINISHPAGE` logic (and in here you can add things like checkboxes, etc). This is what I have in the [current example](https://github.com/conda/constructor/pull/582/files#diff-4f07a130f5a0b0535a6b722e8323b7fc56919877ffa38affff7a2c7360efdcd2R37-R67).
- If you do not specify an `extra_pages` file, the default `MUI_FINISHPAGE` logic will be applied.

jaimergp:
Yes, that's it, more or less!

I'd suggest we reuse the `conclusion_file` key. If the extension is `.nsi`, then we replace the placeholder currently named `EXTRA_PAGES`. If not, we use the standard `MUI_FINISH_PAGE` stuff. We need to document that the user can use this chunk to add as many pages as needed, as long as they use correct NSIS syntax. The examples directory would contain more details.

So, in the `main.nsi.tmpl`, we can have something like:

```
#if conclusion_file_is_nsi
@CONCLUSION_NSI@
#else 
!define MUI_FINISHPAGE_xxxx (this block is the current code)
...
#endif 
```

Does that make sense?

pseudoyim:
@jaime - I think so? Just to be clear, I've listed some scenario and what I believe should happen in each:

**Scenario 1**: User building a Windows installer wants `extra_pages` only.
Their `construct.yaml` looks like:

```
...
extra_pages: my_extra_page.nsi
...
```
RESULT: Extra pages are added, but the default `MUI_FINISH_PAGE` stuff is applied.

**Scenario 2**: User building a Windows installer wants `extra_pages` and a custom `conclusion_file`.
Their `construct.yaml` looks like:

```
...
extra_pages: my_extra_page.nsi
conclusion_file: my_conclusion.nsi
...
```
RESULT: Extra pages are added and the user's own `MUI_FINISH_PAGE` stuff is applied.

**Scenario 3**: User building a Windows installer wants to use a custom `conclusion_file` only.
Their `construct.yaml` looks like:

```
...
conclusion_file: my_conclusion.nsi
...
```
RESULT: The user's own `MUI_FINISH_PAGE` stuff is applied.

Do I have this correct?

jaimergp:
My suggestion involves dropping `extra_pages` in favour of `conclusion_file` (they are adjacent in the main.nsi file anyway). `conclusion_file` can add as many pages as needed before the conclusion page anyway.

From our Slack exchange, `welcome_file` could include a similar mechanism, in which the user is allowed to override `MUI_WELCOME_PAGE` with as much NSIS code as required, which can include several pages if needed.

Under this scenario, we wouldn't have `extra_pages` at all, just two places where you can replace the default welcome and conclusion pages with your own code blocks. This includes adding as many pages as you either before or after.

pseudoyim:
Closing this. Updates are in https://github.com/conda/constructor/pull/590.

PR #124, opened 2017-10-31
Cross conda interface master merge

nehaljwani:
Merging cross_conda_interface into master

DON'T merge this just yet!

PR #584, opened 2022-11-16
Windows - Disable option to add PATH during an AllUsers install

pseudoyim:
### Description

A critical privilege vulnerability was found in non-default installations of Anaconda installers for Windows users ([CVE 2022-26526](https://nvd.nist.gov/vuln/detail/CVE-2022-26526)). This also impacts all Windows installers built using the latest release of `constructor` (`v3.3.1`).

To mitigate this vulnerability, this PR disables the option to add the installation to the PATH environment variable during an "All Users" installation. Users can still add the installation to their PATH environment variable during a "Just Me" installation. Additionally, when installing with Administrator privileges, non-admin system Users will no longer have "Write" permissions.

Currently, the user sees these Advanced Installation Options after selecting `Just Me` or `All Users`:

Screen Shot 2022-11-16 at 2 58 39 PM
Screen Shot 2022-11-16 at 3 02 57 PM

With the changes in this PR, if the user selects `All Users`, the option to add to the PATH environment variable will show as disabled:

Screen Shot 2022-11-16 at 2 58 57 PM

### Checklist - did you ...

### DIFF
--- a/constructor/nsis/OptionsDialog.nsh
+++ b/constructor/nsis/OptionsDialog.nsh
@@ -78,36 +78,20 @@ Function mui_AnaCustomOptions_Show
     ${NSD_OnClick} $mui_AnaCustomOptions.CreateShortcuts CreateShortcuts_OnClick
 
     ${If} "${SHOW_ADD_TO_PATH}" == "yes"
+        # AddToPath is only an option for JustMe installations; it is disabled for AllUsers
+        # installations. (Addresses CVE-2022-26526)
         ${If} $InstMode = ${JUST_ME}
-            StrCpy $1 "my"
-        ${Else}
-            StrCpy $1 "the system"
+            ${NSD_CreateCheckbox} 0 "$5u" 100% 11u "Add ${NAME} to my &PATH environment variable"
+            IntOp $5 $5 + 11
+            Pop $mui_AnaCustomOptions.AddToPath
+            ${NSD_SetState} $mui_AnaCustomOptions.AddToPath $Ana_AddToPath_State
+            ${NSD_OnClick} $mui_AnaCustomOptions.AddToPath AddToPath_OnClick
+            ${NSD_CreateLabel} 5% "$5u" 90% 20u \
+                "NOT recommended. This can lead to conflicts with other applications. Instead, use \
+                the Commmand Prompt and Powershell menus added to the Windows Start Menu."
+            IntOp $5 $5 + 20
+            Pop $Ana_AddToPath_Label
         ${EndIf}
-        ${NSD_CreateCheckbox} 0 "$5u" 100% 11u "Add ${NAME} to $1 &PATH environment variable"
-        IntOp $5 $5 + 11
-        Pop $mui_AnaCustomOptions.AddToPath
-
-        # To address CVE-2022-26526.
-        # In AllUsers install mode, disable AddToPath as an option.
-        ${If} $InstMode = ${ALL_USERS}
-            StrCpy $Ana_AddToPath_State ${BST_UNCHECKED}
-            EnableWindow $mui_AnaCustomOptions.AddToPath 0
-        ${EndIf}
-
-        ${NSD_SetState} $mui_AnaCustomOptions.AddToPath $Ana_AddToPath_State
-        ${NSD_OnClick} $mui_AnaCustomOptions.AddToPath AddToPath_OnClick
-        ${NSD_CreateLabel} 5% "$5u" 90% 20u \
-            "NOT recommended. This can lead to conflicts with other applications. Instead, use \
-            the Commmand Prompt and Powershell menus added to the Windows Start Menu."
-        IntOp $5 $5 + 20
-        Pop $Ana_AddToPath_Label
-
-        # To address CVE-2022-26526.
-        # In AllUsers install mode, disable AddToPath label as well.
-        ${If} $InstMode = ${ALL_USERS}
-            EnableWindow $Ana_AddToPath_Label 0
-        ${EndIf}
-
     ${EndIf}
 
     ${If} "${SHOW_REGISTER_PYTHON}" == "yes"
--- a/constructor/nsis/OptionsDialog.nsh
+++ b/constructor/nsis/OptionsDialog.nsh
@@ -86,6 +86,14 @@ Function mui_AnaCustomOptions_Show
         ${NSD_CreateCheckbox} 0 "$5u" 100% 11u "Add ${NAME} to $1 &PATH environment variable"
         IntOp $5 $5 + 11
         Pop $mui_AnaCustomOptions.AddToPath
+
+        # To address CVE-2022-26526.
+        # In AllUsers install mode, disable AddToPath as an option.
+        ${If} $InstMode = ${ALL_USERS}
+            StrCpy $Ana_AddToPath_State ${BST_UNCHECKED}
+            EnableWindow $mui_AnaCustomOptions.AddToPath 0
+        ${EndIf}
+
         ${NSD_SetState} $mui_AnaCustomOptions.AddToPath $Ana_AddToPath_State
         ${NSD_OnClick} $mui_AnaCustomOptions.AddToPath AddToPath_OnClick
         ${NSD_CreateLabel} 5% "$5u" 90% 20u \
@@ -93,6 +101,13 @@ Function mui_AnaCustomOptions_Show
             the Commmand Prompt and Powershell menus added to the Windows Start Menu."
         IntOp $5 $5 + 20
         Pop $Ana_AddToPath_Label
+
+        # To address CVE-2022-26526.
+        # In AllUsers install mode, disable AddToPath label as well.
+        ${If} $InstMode = ${ALL_USERS}
+            EnableWindow $Ana_AddToPath_Label 0
+        ${EndIf}
+
     ${EndIf}
 
     ${If} "${SHOW_REGISTER_PYTHON}" == "yes"
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -229,7 +229,10 @@ FunctionEnd
                 Install for just me, add to PATH and register as system Python:$\n\
                     $EXEFILE /RegisterPython=1 /AddToPath=1$\n$\n\
                 Install for just me, with no registry modification (for CI):$\n\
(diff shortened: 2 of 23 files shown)


constructor/nsis/OptionsDialog.nsh
```diff
@@ -86,13 +86,28 @@ Function mui_AnaCustomOptions_Show
         ${NSD_CreateCheckbox} 0 "$5u" 100% 11u "Add ${NAME} to $1 &PATH environment variable"
         IntOp $5 $5 + 11
         Pop $mui_AnaCustomOptions.AddToPath
+
+        # To address CVE-2022-26526.
+        # In AllUsers install mode, disable AddToPath as an option.
+        ${If} $InstMode = ${ALL_USERS}
```
jaimergp, line 92: What if we don't show this option at all (and disable it?). All we have to do is make this condition wrap the block below. We should also update the docs in `construct.py` for the relevant keys.

constructor/nsis/main.nsi.tmpl
```diff
@@ -264,6 +264,12 @@ FunctionEnd
         ${EndIf}
     ${EndIf}
 
+    # To address CVE-2022-26526.
+    # In AllUsers install mode, do not allow AddToPath as an option.
```
jaimergp, line 268: Is there a CLI flag we need to disable as well?

pseudoyim: The `AllUsers`/`AddToPath` combo is currently disabled on the CLI as well, but I'm not sure how to best communicate to users that it is disabled. 

If the user installs from the CLI and adds flags for `/InstallationType=AllUsers` and  `/AddToPath=1`, then the PATH will **not** be changed at all (i.e. silently fails to add to PATH). That actually led to [this user raising an issue](https://github.com/ContinuumIO/anaconda-issues/issues/12995).

I tried viewing the CLI options by running
```
MinicondaX-X-Windows-x86_64.exe /?
```
and this window pops up:

Screen Shot 2022-11-21 at 11 31 18 AM

The content on that help page is set [here](https://github.com/conda/constructor/blob/main/constructor/nsis/main.nsi.tmpl#L211-L232). Shall we add a note like this?
```
NOTE: If ./InstallationType=AllUsers, then the AddToPath option is automatically disabled (i.e. if /AddToPath=1, it will be ignored).
```

constructor/nsis/OptionsDialog.nsh
pseudoyim: Thanks @jaimergp. Implemented your suggestion and built a test installer. Here's what happens:

Select "JustMe" installation.
Screen Shot 2022-12-06 at 7 55 26 PM

The user sees the option to "AddToPath".
Screen Shot 2022-12-06 at 7 57 07 PM

Select "AllUsers" installation.
Screen Shot 2022-12-06 at 7 57 14 PM

The option to "AddToPath" does not show up at all.
Screen Shot 2022-12-06 at 7 59 29 PM

I also tested this when running from the CLI and see the same results.

pseudoyim:
@jaimergp - is [this](https://github.com/pseudoyim/constructor/blob/416127b0f0d16d09e6b077a5371dc92fade1f144/constructor/construct.py#L337-L341) the only key for which the description needs to be updated (regarding how the `AddToPath` option is disabled for `AllUsers` installations)?

constructor/nsis/OptionsDialog.nsh
hoechenberger: This looks very good to me, great work!

dbast:
wouldn't the CVE be already be fixed in the "all users" cases by making the folder only admin writable but still allowing to have it in PATH?

news/584-windows-disable-path-for-allusers
```diff
@@ -0,0 +1,19 @@
...
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* (For Windows only) Fix for [CVE-2022-26526](https://nvd.nist.gov/vuln/detail/CVE-2022-26526). Installations for "All Users" will not be allowed the option to modify the system PATH environment variable during installation. Installations for "Just Me" will still be allowed the option to add the installation to their PATH environment variable. Additionally, when installing with Administrator privileges, non-admin system Users will no longer have “Write” permissions.
```
jaimergp, line 19: I'd move this to the `Bug fixes` section.

constructor/nsis/main.nsi.tmpl
jaimergp: Yes, a note would be appreciated. I also wonder if we can write something to the logs or even error out instead of silently ignoring the flag.

It looks like we can add some checks to the line 260:

https://github.com/conda/constructor/blob/ba1289e9faedfebc25ac86b325afbd59fd845a35/constructor/nsis/main.nsi.tmpl#L260

I guess we can do something like this:

```diff
    ${IfNot} ${Errors}
        ${If} $ARGV_AddToPath = "1"
+           ${If} $InstMode = ${ALL_USERS}
+               MessageBox MB_OK|MB_ICONEXCLAMATION "/AddToPath=1 is ignored in 'All Users' installations" /SD IDOK
+               StrCpy $Ana_AddToPath_State ${BST_UNCHECKED}
+           ${Else}
+               StrCpy $Ana_AddToPath_State ${BST_CHECKED}
+           ${EndIf}
-           StrCpy $Ana_AddToPath_State ${BST_CHECKED}
        ${ElseIf} $ARGV_AddToPath = "0"
            StrCpy $Ana_AddToPath_State ${BST_UNCHECKED}
        ${EndIf}
    ${EndIf}
```

jaimergp: We should also document this in `construct.py` (search the relevant key for this setting), and update the docs at `CONSTRUCT.md` via `scripts/make_docs.py`.

dbast (approved):

news/584-windows-disable-path-for-allusers
```diff
@@ -0,0 +1,17 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* (For Windows only) Fix for [CVE-2022-26526](https://nvd.nist.gov/vuln/detail/CVE-2022-26526). Installations for "All Users" will not be allowed the option to modify the system PATH environment variable during installation. Installations for "Just Me" will still be allowed the option to add the installation to their PATH environment variable. Additionally, when installing with Administrator privileges, non-admin system Users will no longer have “Write” permissions.
```
jaimergp, line 7:
```suggestion
* (For Windows only) Fix for [CVE-2022-26526](https://nvd.nist.gov/vuln/detail/CVE-2022-26526). Installations for "All Users" will not be allowed the option to modify the system PATH environment variable during installation. Installations for "Just Me" will still be allowed the option to add the installation to their PATH environment variable. Additionally, when installing with Administrator privileges, non-admin system Users will no longer have “Write” permissions. (#584)
```

pseudoyim:
> LGTM. Can you provide screenshots for the following items?
> 
> * Help message with the new description
> * Error message that appears if `/AddToPath=1` is passed in an "All Users" installation.

@jaimergp, per your request:

**Help message with the new description**
Screen Shot 2022-12-12 at 8 38 50 PM

**Error message that appears if `/AddToPath=1` is passed in an "All Users" installation.**
Screen Shot 2022-12-12 at 8 58 44 PM

It pops up again when the GUI gets to the installation type screen:
Screen Shot 2022-12-12 at 8 59 08 PM

constructor/nsis/main.nsi.tmpl
pseudoyim: Done. Thank you!

pseudoyim:
@jaimergp @hoechenberger - When I was testing the MessageBox (`/AddToPath=1 is disabled and ignored in 'All Users' installations`), it kept popping up exactly two times in a row (after hitting the first OK). 

I traced the problem to `!insertmacro ParseCommandLineArgs` being called twice in [main.nsi.tmpl](https://github.com/conda/constructor/blob/1ce73a3551bfdec04bbe39f64842207e64491725/constructor/nsis/main.nsi.tmpl#L323) ([line 323](https://github.com/conda/constructor/blob/1ce73a3551bfdec04bbe39f64842207e64491725/constructor/nsis/main.nsi.tmpl#L323) and [line 470](https://github.com/conda/constructor/blob/1ce73a3551bfdec04bbe39f64842207e64491725/constructor/nsis/main.nsi.tmpl#L470)). I _think_ it only needs to be called once and [this function](https://github.com/conda/constructor/blob/1ce73a3551bfdec04bbe39f64842207e64491725/constructor/nsis/main.nsi.tmpl#L321-L325) seemed unnecessary to me, so I removed it in [this commit](https://github.com/conda/constructor/pull/584/commits/da2dac81317f3aa5ecdba293f65ef59d0caff13d). After I removed it, this solved the problem of the MessageBox popping up twice.

Could you please review this and validate my thinking?

dbast (dismissed):

jaimergp:
Hm, I am not sure... the logic in that part is really opaque. It looks like they are first parsing the CLI arguments, then overriding some defaults, then parsing again.

Both calls were introduced in [the same PR](https://github.com/conda/constructor/pull/124/files#diff-af1b4db4a14299e7d6a5781c05afe756f88a088a81fbc74f97932ac467d3f821) (particularly, [this commit](https://github.com/nehaljwani/constructor/commit/5baba7f158cde311d54170a46ac2c29dcb772812)), but judging by the branch name that PR is part of a merge/reunification. After that, I can't find any history on why you need it twice.

Since we can't prove it was added by accident, and all we can say it *looks* like deliberate, I'd be inclined to say we shouldn't remove it.

Things we can do to prevent the double warning:
* Move the AllUsers vs AddToPatch check to the `OnInit_Release` function so it's only done once.
* Flip a global var to "remember" whether the user was warned already.

I think I prefer the 1st one.

This is not a pretty part of the codebase so at the very least we should create an issue to investigate the double argument parsing call.

pseudoyim:
Thanks for reviewing @jaimergp. Per your recommendation, I restored the `OnInit_Release` function and moved the AddToPath check under it. I built a test MinicondaX with these changes and the MessageBox shows up only once. Screenshots look the same as what I posted [here](https://github.com/conda/constructor/pull/584#issuecomment-1347681445).

dbast (approved):

dbast:
LGTM to me. all review comments are covered now and it locally tests correctly. mergin.

PR #590, opened 2022-12-07
Customized Welcome After Pages and Conclusion Pages

ryanskeith:
### Description
- This PR enables the adding of extra pages to the Windows GUI installer. These extra pages can display text, links, and images. Such pages can be useful in providing additional instructions, guidance, or promotional materials for end-users before they begin using the application they just installed.
- This addresses issues brought up in PR #582.

### Checklist - did you ...

- [ x ] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
- [ x ] Add / update outdated documentation?

### Screen shots

Screen Shot 2022-12-07 at 12 43 38 PM


jezdez:
@conda-bot check

dbast:
Also requires to locally run `scripts/make_docs.py` and commit the resulting updated CONSTRUCT.md file.

constructor/nsis/main.nsi.tmpl
```diff
@@ -137,6 +137,8 @@ BrandingText /TRIMLEFT "${COMPANY}"
 #!define MUI_PAGE_CUSTOMFUNCTION_SHOW OnStartup
 !define MUI_PAGE_CUSTOMFUNCTION_PRE SkipPageIfUACInnerInstance
 !insertmacro MUI_PAGE_WELCOME
+# Custom welcome file(s)
+@CUSTOM_WELCOME_FILE@
```
jaimergp, line 141: My proposal was that `custom_welcome` would _replace_ `MUI_PAGE_WELCOME`. The replacement code can include the macro again if you want to keep it, but this gives users a chance to actually remove the default Welcome page.

(See related comment in `winexe.py`)

```suggestion
#!define MUI_PAGE_CUSTOMFUNCTION_SHOW OnStartup
#if custom_welcome
# Custom welcome file(s)
@CUSTOM_WELCOME_FILE@
#else
!define MUI_PAGE_CUSTOMFUNCTION_PRE SkipPageIfUACInnerInstance
!insertmacro MUI_PAGE_WELCOME
#endif
```

constructor/winexe.py
```diff
@@ -255,6 +287,9 @@ def make_nsi(info, dir_path, extra_files=()):
                                       '${NAME} ${VERSION} (Python ${PYVERSION} ${ARCH})'
                                       )),
         ('@EXTRA_FILES@', '\n    '.join(extra_files_commands(extra_files, dir_path))),
+        ('@CUSTOM_WELCOME_FILE@', custom_nsi_insert_from_file(info.get('welcome_file', ''))),
```
jaimergp, line 290: Following the comment on the template, we will need a `custom_welcome` boolean added to the `ppd` dictionary.

constructor/nsis/main.nsi.tmpl
```diff
@@ -151,6 +153,10 @@ Page Custom mui_AnaCustomOptions_Show
 !define MUI_FINISHPAGE_TITLE_3LINES
 !define MUI_FINISHPAGE_TEXT __CONCLUSION_TEXT__
 #endif
+
+# Custom conclusion file(s)
+@CUSTOM_CONCLUSION_FILE@
+
 !insertmacro MUI_PAGE_FINISH
```
jaimergp, line 160: Same as with the welcome page, the conclusion is offered as a replacement:

```suggestion
#if custom_conclusion
# Custom conclusion file(s)
@CUSTOM_CONCLUSION_FILE@
#else
!insertmacro MUI_PAGE_FINISH
#endif
```

This is just a suggestion code-wise, feel free to implement it in a different way. If you do use the `#if` conditionals, you'll need to edit the `ppd` dictionary in `winexe.py`.

constructor/nsis/main.nsi.tmpl
```diff
@@ -456,7 +462,7 @@ Function .onInit
     Push $R2
 
     InitPluginsDir
-
+    @EXTRA_FILES_TO_PLUGINS_DIR@
```
jaimergp, line 465: Why do we need the files to reside in the plugins dir? Can they be at any other arbitrary location? We could define a "temporary" target directory for `extra_files` so they are not kept in the final installer artifact, but they are indeed located in a known location so the NSI code can locate them. Something like:

```yaml
extra_files:
  my_custom_welcome_image_1.png: "_nsis_temp/"  # or $PluginsDir if that's what we need.
```

jaimergp (changes requested):
Super cool to see this happening. One more step towards the fork re-integration! Thanks for the submission.

I have two main comments about the implementation:

* I'd rather see the custom welcome and conclusion pages as _replacement_ pages, not additional ones, by default. Of course, the custom page can define again the `MUI_*_PAGE` macros to re-insert the replaced pages, but this way users actually have a way to remove the default pages.
* I'd rather see how we reuse `extra_files` to handle the temporary files.

These are not strong opinions and I'll be happy to discuss them, but this is how I had envisioned the initial implementation.

constructor/winexe.py
```diff
@@ -47,7 +48,7 @@ def pkg_commands(download_dir, dists):
 
 
 def extra_files_commands(paths, common_parent):
-    paths = sorted([PureWindowsPath(p) for p in paths])
+    paths = sorted([Path(p) for p in paths])
```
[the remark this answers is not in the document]
ryanskeith: I was thinking since we are already running this in windows that we didn't need to be double explicit here.  That really was my only reason.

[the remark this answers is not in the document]
jaimergp: Turns out my assumptions were a bit wrong 😬 

According to [this old forum thread](https://nsis-dev.github.io/NSIS-Forums/html/t-183292.html), `makensis` on Linux will expect Unix-like paths (forward slashes). I guess it also auto-translates Windows-like paths to Unix-like paths after 15+ years, but I haven't checked.

Anyway, we can leave this change as is. It's cleaner and raises less questions. If you want to can leave a comment pointing to the forum thread, but not a big deal. Thanks!

constructor/osxpkg.py
```diff
@@ -117,7 +119,9 @@ def modify_xml(xml_path, info):
         root.append(welcome)
 
     ### CONCLUSION ###
-    if "conclusion_file" in info:
+    # The endswith .nsi is for windows specifically.  The nsi script will add in
+    # conclusion pages if added.
+    if "conclusion_file" in info and not info["conclusion_file"].endswith(".nsi"):
```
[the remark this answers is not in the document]
ryanskeith: yes

constructor/nsis/main.nsi.tmpl
ryanskeith: When the extra files are copied using the `copy_extra_files` in preconda.py, Python has no understanding of NSIS variables like $PluginsDir.  So I think using a temp dir as in your example may work.  The question is what happens in that directory after the install is created.  My hope is that it goes away, if not we need to find a way to designate temporary files.

jaimergp: I usually use some variation of `atexit.register(shutil.rmtree, temp_path)` for these cases. I think the only requirement is that the actual temporary path is templated somehow so it's replaced at build time.

constructor/winexe.py
```diff
@@ -259,6 +259,8 @@ def make_nsi(info, dir_path, extra_files=()):
     ppd['with_conclusion_text'] = bool(conclusion_text)
     ppd["enable_debugging"] = bool(os.environ.get("NSIS_USING_LOG_BUILD"))
     ppd["has_conda"] = info["_has_conda"]
+    ppd["custom_welcome"] = "welcome_file" in info and info["welcome_file"].endswith(".nsi")
+    ppd["custom_conclusion"] = "conclusion_file" in info and info["conclusion_file"].endswith(".nsi")
```
jaimergp, line 263: This is a bit more succinct if you like it better:

```suggestion
    ppd["custom_welcome"] = info.get("welcome_file", "").endswith(".nsi")
    ppd["custom_conclusion"] = info.get("conclusion_file", "").endswith(".nsi")
```

constructor/osxpkg.py
```diff
@@ -110,6 +110,8 @@ def modify_xml(xml_path, info):
             f.write(info["welcome_text"])
     else:
         welcome_path = None
+        if "welcome_file" in info and info["welcome_file"].endswith(".nsi"):
+            print(f"NSI welcome_file, {info['welcome_file'].endswith('.nsi')}, is ignored.")
```
jaimergp, line 114:
```suggestion
            print(f"Warning: NSI welcome_file, {info['welcome_file'].endswith('.nsi')}, is ignored.")
```

[Other parts of the code](https://github.com/conda/constructor/blob/02ff321671d791f05978f1ab12e82a0b4b90688d/constructor/utils.py#L145) prefix the message with `Warning: `, adding it here for consistency.

constructor/osxpkg.py
```diff
@@ -132,7 +134,8 @@ def modify_xml(xml_path, info):
                 f.write(info["conclusion_text"])
     else:
         conclusion_path = join(OSX_DIR, 'acloud.rtf')
-
+        if "conclusion_file" in info and info["conclusion_file"].endswith(".nsi"):
+            print(f"NSI conclusion_file, {info['conclusion_file'].endswith('.nsi')}, is ignored.")
```
jaimergp, line 138:
```suggestion
            print(f"Warning: NSI conclusion_file, {info['conclusion_file'].endswith('.nsi')}, is ignored.")
```

[Other parts of the code](https://github.com/conda/constructor/blob/02ff321671d791f05978f1ab12e82a0b4b90688d/constructor/utils.py#L145) prefix the message with `Warning: `, adding it here for consistency.

You can also apply the `dict.get(key, "")` trick here if you want!

constructor/nsis/main.nsi.tmpl
ryanskeith: If this were in Python, I agree with this but the problem we are trying to solve is to not have temporary files just used for installation left around after install. At this point, it is an NSIS problem.  This is why using the NSIS PlugsDir is nice because it is a builtin way to have files that are only used when the installer is run and installed with the files needed to run the application.

ryanskeith: I feel extra_files should be used for files that will be installed on the target host in the installation process and we want to implement a way label files only meant to be used during the process and not meant to be left on the host.  I am not seeing at clean way to do this with Windows as NSIS is in charge of creating the installer.  I think the cleaner solution would be to make a new keyword that encapsulates the idea of temporary files only meant for installation.

jaimergp: Yes, I see your point. The idea was to be "magic" about the target path, by detecting a specific suffix (e.g. `@build-time-only/some_file.txt`). These files wouldn't be copied to the installer, but to a temporary location that would be templated in the NSIS code.

For example, given this `extra_files` config:

```yaml
extra_files:
  - my_image.bmp: @build_time_only/my_image.bmp
```

The template code would have:

```
StrCpy $0 "__BUILD_TIME_ONLY_TEMP_DIR__\my_image.bmp"
System::Call 'user32::LoadImage(i 0, t r0, i ${IMAGE_BITMAP}, i 0, i 0, i ${LR_LOADFROMFILE}|${LR_LOADTRANSPARENT}|${LR_LOADMAP3DCOLORS}) i.s'
```

That, when rendered, would become:

```
StrCpy $0 C:\tmp\constructor_nsfjr7184_tmp\my_image.bmp
System::Call 'user32::LoadImage(i 0, t r0, i ${IMAGE_BITMAP}, i 0, i 0, i ${LR_LOADFROMFILE}|${LR_LOADTRANSPARENT}|${LR_LOADMAP3DCOLORS}) i.s'
```

So the only "contract" is the use of that templated placeholder instead of $PluginsDir.

Is reusing `extra_files` with "magic prefixes" too magic? I am willing to accept a `build_resources` or something like that if the team agrees (but the templated placeholder seems like a good way to go, regardless).

ryanskeith: @pseudoyim, @marcoesters and I were playing around with this today.  The crux of the problem is that we still need to package the bmp files with the NSIS installer.  The solution above works only locally as it is directly reference the image on the host machine.  We found that temp directory is removed it will not work.

We still need a way to designate to NSIS that the file is temporary.  So I think it is cleaner just to create new keyword.  An alternative approach would be to create code that puts these files in a temporary directory at install and then deletes them but to me this is cleaner:
```
temp_extra_files:
  - my_image.bmp
  ```
  This requires little need for the user to add things to their script as it gets handled.
  
I have a [PR](https://github.com/conda/constructor/pull/594), #594 that has a proposed solution to encapsulate this.

Comment (jaimergp) [reply constructor/nsis/main.nsi.tmpl]: Ok, let's go for the new key. About the name, do you think `build_time_files` reflects its purpose better? I am not sure if users would understand that "temporary" here means "temporary at build time" or "temporary at install time".

Comment (jaimergp) [reply constructor/nsis/main.nsi.tmpl]: Is 594 supposed to target this PR branch or `main` separately? It looks like it has way too many files in the diff (quick glance only).

Comment (ryanskeith) [reply constructor/nsis/main.nsi.tmpl]: Yes, @jaimergp , I branched off of this to make and test the changes on #594.  If we went ahead and merged this change the diff would be very small.

As far as the naming goes, I am flexible.  Currently our documentation for `temp_extra_files` is as follows:
```
Temporary files that could be referenced in the installation process (i.e. customized
 `welcome_file` and `conclusion_file` (see above)) . Should be a list of
filepaths, relative to the directory where `construct.yaml` is. In Windows, these
files will be copied into a temporary folder, the NSIS `$PLUGINSDIR`, during
install process (Windows only).
```

I think `build_time_files` may be too specific as there are files that really aren't for building but are used during the installation process.  This is one of our primary use cases at the moment.  I could see this word being expanded to include building resources or files so I think we should let it stand right now as `temp_extra_files`.

Comment (jaimergp) [reply constructor/nsis/main.nsi.tmpl]: > I think `build_time_files` may be too specific as there are files that really aren't for building but are used during the installation process. 

Yes, I see what you mean. The intent is that these files will not be part of the final installation, but only required while the installer is being generated. If the installer ends up embedding those resources somehow, that's ok, but the user won't see them in `$INSTDIR` once done, right?

In your current help message "the install process" is not when the user executes the distributed installer, but when the team is building the installer. Is that correct or am I misunderstanding? That's why I was calling them "build-time" files.

My main concern is that `temporary` in here is not a term that really describes how these type of files are being used.

File: news/590-windows-installer-enhancement [hunk 10]
```diff
@@ -0,0 +1,4 @@
+### Enhancements
+
+* Windows GUI installer enhancement to allow for additional custom pages.  These new pages can be added after the welcome page and before the conclusion. These extra pages can display text, links, and images. Such pages can be useful in providing additional instructions, guidance, or promotional materials for end-users before they begin using the application they just installed.
```
jaimergp, line 3:
```suggestion
* Windows GUI installer enhancement to allow for additional custom pages.  These new pages can be added after the welcome page and before the conclusion. These extra pages can display text, links, and images. Such pages can be useful in providing additional instructions, guidance, or promotional materials for end-users before they begin using the application they just installed. (#590)
```

constructor/osxpkg.py
```diff
@@ -128,7 +134,8 @@ def modify_xml(xml_path, info):
                 f.write(info["conclusion_text"])
     else:
         conclusion_path = join(OSX_DIR, 'acloud.rtf')
-
+        if info.get("conclusion_file","").endswith(".nsi"):
```
jaimergp, line 137:
```suggestion
        if info.get("conclusion_file", "").endswith(".nsi"):
```

constructor/nsis/main.nsi.tmpl
jaimergp: Also, can you add the commits of #594 to this PR directly? I'd rather merge this in a single step, and I'll need the whole logic in a single view for the final review :D 

Right now I am a bit confused with `$PluginsDir` and friends 😅

ryanskeith: Ok, Will do.

ryanskeith:
@jaimergp changes from PR #594 have been merged when you would like to take a gander.

constructor/nsis/main.nsi.tmpl
ryanskeith: @jaimergp Should we consider this resolved and close this issue?

jaimergp: Sure, I'll add more comments in the current commit if needed. Thanks for the discussion!

constructor/winexe.py
```diff
@@ -59,6 +59,38 @@ def extra_files_commands(paths, common_parent):
...
+        List[str]: Commands to be inserted into nsi template
+    """
+    if paths:
+        # Setting OutPath to PlugnsDir so NSIS File command copies the path into the PluginsDir
+        lines = ['SetOutPath $PLUGINSDIR']
+        for path in sorted([Path(p) for p in paths]):
+            lines.append(f"File {path}")
+        return lines
+    else:
+        return ['']
```
jaimergp, line 78:
```suggestion
    if not paths:
        return []
    # Setting OutPath to PluginsDir so NSIS File command copies the path into the PluginsDir
    lines = ['SetOutPath $PLUGINSDIR']
    for path in sorted([Path(p) for p in paths]):
        lines.append(f"File {path}")
    return lines
```

I usually deal with the least frequent cases first, then move to the general. Matter of taste, no strong opinion, but in this case I think it reads better.

Another question about this function:

- Why do we return `['']` instead of `[]`? I find it a bit unexpected and might deserve a comment if we really need to leave it as such.

examples/customized_welcome_conclusion/construct.yaml
```diff
@@ -0,0 +1,11 @@
...
+version: X
+installer_type: all
+channels:
+  - http://repo.anaconda.com/pkgs/main/
+specs:
+  - python
+conclusion_file: custom_conclusion.nsi  # [win]
+welcome_file: custom_welcome.nsi  # [win]
+temp_extra_files:
+  - ExtraPagesExampleImg.bmp
```
jaimergp, line 11:
```suggestion
  - ExtraPagesExampleImg.bmp

```

I think all examples have this extra newline at the end.

news/590-windows-installer-enhancement
```diff
@@ -0,0 +1,4 @@
+### Enhancements
+
+* Windows GUI installer enhancement to allow for additional custom pages.  These new pages can be added after the welcome page and before the conclusion. These extra pages can display text, links, and images. Such pages can be useful in providing additional instructions, guidance, or promotional materials for end-users before they begin using the application they just installed. (#590)
```
jaimergp, line 3:
```suggestion
* Windows GUI installer enhancement to allow for additional custom pages. These new pages can be added after the welcome page and before the conclusion. These extra pages can display text, links, and images. Such pages can be useful in providing additional instructions, guidance, or promotional materials for end-users before they begin using the application they just installed. (#590)
```

examples/customized_welcome_conclusion/custom_conclusion.nsi
```diff
@@ -0,0 +1,69 @@
+Page Custom muiExtraPages_Create
```
jaimergp, line 1: These example NSIS scripts will be part of the examples and probably consulted by people wanting to implement similar features. It might deserve some documentation on how it works and the overall logic followed. A couple comments at the beginning would be really welcome, at least.

jaimergp:
@ryanskeith Some more comments that need addressing, but we are getting very close to a mergeable state :D I would like to play with this PR a bit to get a feeling on how it could work for other projects before merging, but I think it all can happen within the week!

constructor/winexe.py
ryanskeith: Hey Jaime, I guess it actually doesn't make a difference!  They both produce the same thing at the end of the day.

ryanskeith:
@jaimergp Comments added to requested file.

jaimergp:
Thanks @ryanskeith! I fixed the conflicts and added one more (last?) question. Can you resync the docs (`python scripts/make_docs.py`)? I am going to merge soon and play with it once we are done. <3

examples/customized_welcome_conclusion/custom_welcome.nsi
```diff
@@ -0,0 +1,45 @@
...
+        "${PRODUCT_NAME}" \
+        "Extra Pages Example"
+
+    nsDialogs::Create /NOUNLOAD 1018
+    ${NSD_CreateLabel} 10u 10u 280u 40u "Extra Welcome Page 1.$\r$\n$\r$\n$\r$\nHere is a link to the conda Github organization:"
+    Pop $IntroAfterWelcomeText
+
+    ${NSD_CreateLink} 10u 55u 200u 10u "https://github.com/conda"
+    Pop $InstallationAfterWelcomeLink
+    ${NSD_OnClick} $InstallationAfterWelcomeLink LaunchLinkOne
```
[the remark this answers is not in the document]
ryanskeith: Yes, it seemed to work in our tests.

dbast:
docs should be resynced

examples/customized_welcome_conclusion/custom_welcome.nsi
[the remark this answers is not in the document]
jaimergp: Awesome. Thanks!

jaimergp:
Thanks! I'll merge as soon as everything's green.

dbast (approved):

jaimergp (approved):

PR #595, opened 2022-12-21
Add new 'build_outputs' key for non-installer artifacts

jaimergp:
### Checklist - did you ...

Supersedes https://github.com/conda/constructor/pull/508, with better control of what's produced, and without affecting conda's verbosity.

This introduces a new key, `build_outputs`, which takes a list of "plugin" names (and optionally their configurations). Each plugin (as defined in `build_outputs.py`) is allowed to generate one or more "non-installer" artifacts, like JSON files or text files.

This allows us to ask `constructor` for additional metadata usually not available:
- The internal `info` dictionary
- The solved environments
- Included licenses (super convenient for legal audits of the generated artifacts).

I've designed in a way that it can be extended with more outputs if needed. For example, one common need in corporate environments: CVE scans (although this possibly involves external tooling, which is a bit too much maybe).

I was a bit unsure about the `build_outputs` name (I also considered `build_artifacts` or `extra_outputs`), so I welcome suggestions if there are any.

### DIFF
--- a/CONSTRUCT.md
+++ b/CONSTRUCT.md
@@ -644,6 +644,20 @@ This setting can be passed as a list of:
 - `str`: each found file will be copied to the root prefix
 - `Mapping[str, str]`: map of path in disk to path in prefix.
 
+## `build_outputs`
+
+_required:_ no<br/>
+_type:_ list<br/>
+Additional artifacts to be produced after building the installer.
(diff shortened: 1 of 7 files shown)


goanpeca (approved):

larsoner:
Thanks @jaimergp !

PR #437, opened 2021-02-16
add doc build workflow

epassaro:
This PR addresses #398 

Documentation will be deployed to https://conda.github.io/constructor using the `gh-pages` branch.

An example on my fork: https://epassaro.github.io/constructor.

PR #598, opened 2022-12-28
Add docs workflow

jaimergp:
### Description

Copied and adapted from the `conda-libmamba-solver` docs.

Closes #437 (@epassaro, I kept your history here so you get your commits counted. I hope that's ok!).

### Checklist - did you ...


epassaro:
Nice!

jaimergp:
pre-commit.ci autofix

jaimergp:
pre-commit.ci autofix

.github/workflows/docs.yml
```diff
@@ -0,0 +1,64 @@
...
+    branches: ["main"]
+  pull_request:
+    branches: ["main"]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# Allow one concurrent deployment
+concurrency:
+  group: "pages"
```
dbast, line 15: won't pr builds in this case kill a running build on main branch? what about:
```suggestion
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
```

jaimergp:
That, and consistency with the rest of the documentation is being written in other projects within the organization.

.github/workflows/docs.yml
jaimergp: @jezdez - this comes from the conda-libmamba-solver docs. Any comments?

jaimergp: Ignore this, the workflow in conda-libmamba-solver does not run on PRs.

The idea is that if we merge too many PRs at once, subsequent pushes to main do not run into deployment run conditions and only the last one is pushed.

So we need a ternary conditional like:

`<your suggested expression> if on_pull_request else "pages"`

Which I _think_ can be done like this:

```suggestion
  group: ${{ github.event_name == 'pull_request' && join([github.workflow, (github.event.pull_request.number || github.sha)], '-') || 'pages' }}
 ```

jaimergp: @dbast - wdyt?

dbast: can be simplified to:
`pages-${{ github.workflow }}-${{ github.event.pull_request.number }}`  no && and || required as values will afaik be just empty in case of not a PR.

jaimergp:
pre-commit.ci autofix

dbast (approved):

PR #600, opened 2023-01-02
header.sh and osx scripts hardening via set -eu + shellcheck

dbast:
### Description

Shell scripts don't stop per default when commands finish with an error causing masked errors and undefined behaviours. `set -e` changes that behaviour by stoping in all cases where errors happen enabling better error reports on the actual error. If an error should be ignored then a command can be run via `$cmd || true`. 

This PR:
* adds `set -eu` to `/bin/sh` scripts and `set -euo pipefail` to `/bin/bash` scripts
* contains two test cases running shellcheck to ensure that the header.sh and osx scripts are shellcheck clean (this also found also the spaces issues in script handled in https://github.com/conda/constructor/pull/449). Those tests would have requested a lot more error handling if `set -e` wasn't set in all the scripts.
* fixes all shellcheck findings

### Checklist - did you ...


jaimergp:
Just a note about #449, which quoted many variables too. Maybe we should merge that first, now that support for spaces in conda seems to have been implemented robustly after some initial hiccups.

dbast:
yes, lets fix/finish and merge #449 first.

constructor/header.sh
```diff
@@ -450,7 +452,7 @@ for env_pkgs in "${PREFIX}"/pkgs/envs/*/; do
     if [ "$env_name" = "*" ]; then
         continue
     fi
-    printf "\nInstalling ${env_name} environment...\n\n"
+    printf "\nInstalling %s environment...\n\n" "${env_name}"
```
jaimergp, line 455: I am curious why do we need `printf` instead of a simple `echo`? The `\n` newlines?

constructor/header.sh
```diff
@@ -517,13 +519,13 @@ cat <<EOF
...
 
-if [ "$PYTHONPATH" != "" ]; then
+if [ "${PYTHONPATH:-}" != "" ]; then
     printf "WARNING:\\n"
     printf "    You currently have a PYTHONPATH environment variable set. This may cause\\n"
     printf "    unexpected behavior when running the Python interpreter in __NAME__.\\n"
     printf "    For best results, please verify that your PYTHONPATH only points to\\n"
     printf "    directories of packages that are compatible with the Python interpreter\\n"
-    printf "    in __NAME__: $PREFIX\\n"
+    printf "    in __NAME__: %s\\n" "$PREFIX"
```
jaimergp, line 528: How come we escape the `\n` newline here but not in the other instance I mentioned above?

constructor/header.sh
```diff
@@ -597,8 +601,7 @@ if [ "$TEST" = "1" ]; then
      fi
      conda index "$PREFIX"/conda-bld/__PLAT__/
      conda-build --override-channels --channel local --test --keep-going "$PREFIX"/conda-bld/__PLAT__/*.tar.bz2
-    )
-    NFAILS=$?
+    ) || NFAILS=$?
```
jaimergp, line 604: Why is this change needed? Shouldn't the other one behaev in the same way?

constructor/header.sh
```diff
@@ -5,10 +5,12 @@
 # PLAT:  __PLAT__
 # MD5:   __MD5__
 
+set -eu
```
jaimergp, line 8: The other scripts have `-o pipefail` too. Would that work here, or do we depend on that behavior?

constructor/osx/clean_cache.sh
```diff
@@ -4,7 +4,7 @@
 
 # $2 is the install location, which is ~ by default
 # but which the user can change.
-set -x
+set -eux
```
jaimergp, line 7: Same here, I'd rather use the same flags (with or without `x` depending on the use case, but same error signalling flags).

constructor/osx/prepare_installation.sh
```diff
@@ -31,8 +34,7 @@ touch "$PREFIX/conda-meta/history"
 # Extract the conda packages but avoiding the overwriting of the
 # custom metadata we have already put in place
 notify "Preparing packages..."
-"$CONDA_EXEC" constructor --prefix "$PREFIX" --extract-conda-pkgs
-if (( $? )); then
+if ! "$CONDA_EXEC" constructor --prefix "$PREFIX" --extract-conda-pkgs; then
```
jaimergp, line 37: Does this need grouping in the command or is this an idiomatic shell expression? Sorry about my ignorance, just trying to understand better :D

constructor/osx/run_installation.sh
```diff
@@ -24,12 +27,11 @@ CONDA_EXEC="$PREFIX/conda.exe"
 
 # Perform the conda install
 notify "Installing packages. This might take a few minutes."
-CONDA_SAFETY_CHECKS=disabled \
+if ! CONDA_SAFETY_CHECKS=disabled \
```
jaimergp, line 30: I am not sure this way of expressing the same thing is more readable or not. The other one is clearer in my opinion (checking the "booleanness" of the return code).

constructor/osx/update_path.sh
```diff
@@ -4,7 +4,8 @@
 
 # $2 is the install location, which is ~ by default, but which the user can
 # change.
-set -x
+set -eux
```
jaimergp, line 7: `-o pipefail`?

tests/test_header.py
```diff
@@ -53,21 +82,84 @@ def test_linux_template_processing(
...
+@pytest.mark.parametrize("batch_mode", [True])
+@pytest.mark.parametrize("keep_pkgs", [True])
+@pytest.mark.parametrize("has_conda", [False, True])
+@pytest.mark.parametrize("has_license", [True])
+@pytest.mark.parametrize("initialize_conda", [True])
+@pytest.mark.parametrize("initialize_by_default", [True])
+@pytest.mark.parametrize("has_post_install", [True])
+@pytest.mark.parametrize("has_pre_install", [False])
+@pytest.mark.parametrize("arch", ["x86_64", "aarch64"])
+@pytest.mark.parametrize("check_path_spaces", [True])
```
jaimergp, line 122: I know this is how we are checking other templated scripts but man is it getting out of hand 😂 Not a suggestion for now, but we need to come up with a different mechanism that doesn't suffer from combinatorial explosions so much. (Note to self!)

constructor/header.sh
dbast: There seems to be lots of reasons https://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo 

The reason this line was updated is due to https://www.shellcheck.net/wiki/SC2059

constructor/header.sh
dbast: that escaping is not required, but both leads to the same output see e.g. https://unix.stackexchange.com/questions/564998/why-doesnt-printf-escape-newlines

constructor/header.sh
dbast: With `set -e` being set, the shell would stop with an error before it evaluates `NFAILS=$?` when it is in a *newline*.

If the subshell expression is extended in the same line via `$( code ) || extra_code`, then the extra_code is executed in case of error allowing error code evaluation.

constructor/header.sh
dbast: what can be set depends on the shebang... `#!/bin/sh` in this case does not support `-o pipefail` (a very nice feature that requires bash). shellcheck also tells if shebang and set options are incompatible.

constructor/osx/clean_cache.sh
dbast: see above sh vs bash set options... x was given, only eu get added due to sh.

constructor/osx/prepare_installation.sh
dbast: see https://www.shellcheck.net/wiki/SC2181

constructor/osx/run_installation.sh
dbast: see https://www.shellcheck.net/wiki/SC2181

The previous code had to be changed due to `set -e` being enabled.

constructor/osx/update_path.sh
dbast: sh vs bash... so pipefail not supported.... and there are good reasons to stick with sh on osx as bash does not receive updates there due to GPL3 see https://www.reddit.com/r/bash/comments/393oqv/why_is_the_version_of_bash_included_in_os_x_so_old/

tests/test_header.py
```diff
@@ -1,10 +1,39 @@
...
+        cmd = [
+            "shellcheck",
+            # https://www.shellcheck.net/wiki/SC2034
+            "--exclude=SC2034",
+            f"{f.name}",
+        ]
+        p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+        sc_stdout, _ = p.communicate()
+        findings = (
+            sc_stdout.decode(sys.stdout.encoding).replace("\r\n", "\n").splitlines()
```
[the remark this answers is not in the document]
dbast: thanks ... will update.

tests/test_header.py
dbast: yep ... leads to >20k test cases if a parameter options are given... executing shellcheck would take forever due subshell and command startup. reduced the options to have only 16 shellcheck runs.

we could stop templating the shellcripts and instead have some parameters defined at the beginning and use real if/else shell conditionals instead of templating if/else conditionals... that would result in one shellcheck run to check everything :) ... every .sh. installer would have the same header.sh attached, except of the variable at the beginning.

tests/test_header.py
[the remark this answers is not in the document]
dbast: Done via text=True (universal_newlines=True is an alias to that)

dbast:
@jaimergp Rebased + split the commits and pushed them one after another to be able to see the shellcheck output after each commit.

constructor/header.sh
jaimergp: TIL! Nice, nice. Thanks for the details.

constructor/header.sh
jaimergp: Love it, thanks!

constructor/header.sh
jaimergp: Wow, what a rabbit hole. Thanks for the link. Yes, let's leave it like that 😂

constructor/header.sh
jaimergp: Got it.

Seems like single quotes are usually preferred for the `printf` format string, but we can do that in a separate PR.

jaimergp (approved):
Thanks for the work and detailed answers!

PR #602, opened 2023-01-10
Add `text_errors` option to `build_outputs.licenses`

jaimergp:
### Description

Continued work from #595

Some bundled license texts can be malformed, which leads to decoding errors. ([see examples on Windows here](https://github.com/napari/packaging/actions/runs/3884299816/jobs/6626915682#step:14:14906)). This allows users to specify how to handle that through the `open()` `errors` argument directly. Normally, you'd set this to `replace`.

### Checklist - did you ...

### DIFF
--- a/CONSTRUCT.md
+++ b/CONSTRUCT.md
@@ -655,8 +655,11 @@ Allowed keys are:
 - `pkgs_list`: The list of packages contained in a given environment. Options:
     - `env` (optional, default=`base`): Name of an environment in `extra_envs` to export.
 - `licenses`: Generate a JSON file with the licensing details of all included packages. Options:
-    - `include_text` (optional, default=`False`): Whether to dump the license text in the JSON.
+    - `include_text` (optional bool, default=`False`): Whether to dump the license text in the JSON.
       If false, only the path will be included.
+    - `text_errors` (optional str, default=`None`): How to handle decoding errors when reading the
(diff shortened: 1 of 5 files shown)


CONSTRUCT.md
```diff
@@ -658,8 +658,11 @@ Allowed keys are:
 - `pkgs_list`: The list of packages contained in a given environment. Options:
     - `env` (optional, default=`base`): Name of an environment in `extra_envs` to export.
 - `licenses`: Generate a JSON file with the licensing details of all included packages. Options:
-    - `include_text` (optional, default=`False`): Whether to dump the license text in the JSON.
+    - `include_text` (optional bool, default=`False`): Whether to dump the license text in the JSON.
       If false, only the path will be included.
+    - `text_errors` (optional str, default=`None`): How to handle decoding errors when reading the
```
larsoner, line 663: For a second I thought perhaps `'replace'` would be a better default, but in the end it's probably safest to not mess with people's included licenses, and make them change a var to do it themselves, so +1 on `None` as the default as it is now

larsoner (approved):

larsoner:
Thanks @jaimergp

CONSTRUCT.md
jaimergp: Yea I thought about it too. Worst case people can use the `path` attribute and check / collect the licenses themselves.

PR #605, opened 2023-01-11
Add tests for micromamba, transmutation and miniforge (continued)

jaimergp:
### Description

Continues and closes #503 

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](../blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [X] Add / update necessary tests?
- [x] Add / update outdated documentation?

PR #606, opened 2023-01-13
Set "python_requires='>=3.7'" and apply pyupgrade via pre-commit

dbast:
Python versions <3.7 are not supported and also got removed from the test matrix since a while. Stating this in the setup.py makes this more clear.

pyupgrade: Removes future imports, updates to f-strings, removes utf-8 pragmas (utf8 is default in py3), updates exceptions which are aliases and more.

pyupgrade was also applied without issues to conda/conda-build/conda-pack.

No logic changes.

### Checklist - did you ...

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,6 +16,12 @@ repos:
       - id: mixed-line-ending
       - id: end-of-file-fixer
       - id: trailing-whitespace
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v3.3.1
+    hooks:
+      - id: pyupgrade
+        args: ["--py37-plus", "--keep-percent-format"]
+        exclude: ^(constructor/nsis/.*py)
   - repo: https://github.com/PyCQA/flake8
     rev: 6.0.0
     hooks:
--- a/constructor/construct.py
+++ b/constructor/construct.py
@@ -326,21 +326,21 @@ Metadata about the installer can be found in the `%INSTALLER_NAME%`,
 `%INSTALLER_TYPE%` is set to `EXE`.
 '''),
 
-    ('default_prefix',         False, str, '''
+    ('default_prefix',         False, str, r'''
 Set default install prefix. On Linux, if not provided, the default prefix is
 `${HOME}/${NAME}`. On windows, this is used only for "Just Me" installation;
 for "All Users" installation, use the `default_prefix_all_users` key.
 If not provided, the default prefix is `${USERPROFILE}\${NAME}`.
 '''),  # noqa
 
-    ('default_prefix_domain_user', False, str, '''
+    ('default_prefix_domain_user', False, str, r'''
 Set default installation prefix for domain user. If not provided, the
 installation prefix for domain user will be `${LOCALAPPDATA}\${NAME}`.
 By default, it is different from the `default_prefix` value to avoid installing
 the distribution in the roaming profile. Windows only.
 '''),  # noqa
 
-    ('default_prefix_all_users', False, str, '''
+    ('default_prefix_all_users', False, str, r'''
 Set default installation prefix for All Users installation. If not provided,
 the installation prefix for all users installation will be
 `${ALLUSERSPROFILE}\${NAME}`. Windows only.
@@ -672,7 +672,7 @@ def parse(path, platform):
     try:
         with open(path) as fi:
             data = fi.read()
-    except IOError:
+    except OSError:
         sys.exit("Error: could not open '%s' for reading" % path)
     directory = dirname(path)
     content_filter = partial(select_lines, namespace=ns_platform(platform))
--- a/constructor/exceptions.py
+++ b/constructor/exceptions.py
@@ -13,7 +13,7 @@ class YamlParsingError(Exception):
 
 class UnableToParse(YamlParsingError):
     def __init__(self, original, *args, **kwargs):
-        super(UnableToParse, self).__init__(*args, **kwargs)
+        super().__init__(*args, **kwargs)
         self.original = original
 
     def error_msg(self):
@@ -31,13 +31,13 @@ class UnableToParse(YamlParsingError):
     def indented_exception(self):
         orig = str(self.original)
         def indent(s): return s.replace("\n", "\n--> ")
-        return "Error Message:\n--> {}\n\n".format(indent(orig))
+        return f"Error Message:\n--> {indent(orig)}\n\n"
 
 
 class UnableToParseMissingJinja2(UnableToParse):
     def error_body(self):
         return "\n".join([
-            super(UnableToParseMissingJinja2, self).error_body(),
+            super().error_body(),
             indent("""\
                 It appears you are missing jinja2.  Please install that
                 package, then attempt to build.
--- a/constructor/fcp.py
+++ b/constructor/fcp.py
@@ -6,7 +6,6 @@
 """
 fcp (fetch conda packages) module
 """
-from __future__ import absolute_import, division, print_function
 
 from collections import defaultdict
 import json
@@ -39,7 +38,7 @@ def getsize(filename):
 
 
 def warn_menu_packages_missing(precs, menu_packages):
-    all_names = set(prec.name for prec in precs)
+    all_names = {prec.name for prec in precs}
     for name in menu_packages:
         if name not in all_names:
             print("WARNING: no such package (in menu_packages): %s" % name)
@@ -183,9 +182,9 @@ def check_duplicates_files(pc_recs, platform, duplicate_files="error"):
             msg_str = "File '%s' found in multiple packages: %s" % (
                     member, ', '.join(fns))
             if duplicate_files == "warn":
-                print('Warning: {}'.format(msg_str))
+                print(f'Warning: {msg_str}')
             else:
-                sys.exit('Error: {}'.format(msg_str))
+                sys.exit(f'Error: {msg_str}')
 
     for member in map_members_icase:
         # Some filesystems are not case sensitive by default (e.g HFS)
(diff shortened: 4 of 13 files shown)


.pre-commit-config.yaml
```diff
@@ -16,6 +16,16 @@ repos:
...
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v3.3.1
+    hooks:
+      - id: pyupgrade
+        args: ["--py37-plus"]
+  - repo: https://github.com/akaihola/darker
+    rev: 1.6.1
+    hooks:
+      - id: darker
+        additional_dependencies: [black]
```
goanpeca, line 28: This is awesome @dbast !

Thanks for working on it.

How about using https://github.com/charliermarsh/ruff#usage ? whcih is faster and bundles a bunch of tools into a single one?

dbast: ruff would replace flake8? something we could try via another PR.

the idea here is to apply pyupgrade (don't see its functionality included in ruff) and to also style all the lines that are upgraded by it with black, which is done by darker (gradually styles lines with black that are anyway already changed). sorry if that wasn't clear.

constructor/construct.py
```diff
@@ -15,26 +15,42 @@
 
 # list of tuples (key name, required, type, description)
 KEYS = [
-    ('name',                   True,  str, '''
+    (
```
jaimergp, line 18: I was thinking of turning `KEYS` into a proper `pydantic` schema instead of this homemade pseudo-validation, in a future release. I am inclined to ignore changes in this file for now.


They are only formatting changes so that's fine but my thinking is that this file won't last as is for too long.

constructor/construct.py
```diff
@@ -701,53 +962,57 @@ def verify(info):
...
             obsolete_keys.add(key)
 
     for key in info:
         if key not in types_key:
             sys.exit("Error: unknown key '%s' in construct.yaml" % key)
         elt = info[key]
         if key in obsolete_keys:
-            sys.stderr.write("Warning: key '%s' is obsolete.\n"
-                             "  Its value '%s' is being ignored.\n" % (key, elt))
+            sys.stderr.write(
```
jaimergp, line 973: Unrelated but:

The whole logging business needs some love too. There's a mix of `print` and `sys.std*.write` logic to handle logging...

constructor/construct.py
```diff
@@ -701,53 +962,57 @@ def verify(info):
...
+                "Warning: key '%s' is obsolete.\n"
+                "  Its value '%s' is being ignored.\n" % (key, elt)
+            )
         types = types_key[key]
         if not isinstance(elt, types):
-            sys.exit("Error: key '%s' points to %s,\n"
-                     "       expected %s" % (key, type(elt), types))
+            sys.exit(
```
jaimergp, line 979: Another "unrelated but":

Exception handling and `sys.exit` are used interchangeably and should be revisited.

constructor/exceptions.py
```diff
@@ -31,15 +31,19 @@ def error_body(self):
...
 class UnableToParseMissingJinja2(UnableToParse):
     def error_body(self):
-        return "\n".join([
-            super(UnableToParseMissingJinja2, self).error_body(),
-            indent("""\
+        return "\n".join(
+            [
+                super().error_body(),
+                indent(
+                    """\
```
jaimergp, line 43: What about adding a `.lstrip()` call instead of escaping the newline here?

constructor/fcp.py
```diff
@@ -180,27 +179,26 @@ def check_duplicates_files(pc_recs, platform, duplicate_files="error"):
...
     for member in map_members_icase:
         # Some filesystems are not case sensitive by default (e.g HFS)
         # Throw warning on linux and error out on macOS/windows
-        fns = map_members_icase[member]['fns']
-        files = list(map_members_icase[member]['files'])
-        msg_str = "Files %s found in the package(s): %s" % (
-            str(files)[1:-1], ', '.join(fns))
+        fns = map_members_icase[member]["fns"]
+        files = list(map_members_icase[member]["files"])
+        msg_str = "Files {} found in the package(s): {}".format(str(files)[1:-1], ", ".join(fns))
```
jaimergp, line 193: f-string?

constructor/fcp.py
```diff
@@ -180,27 +179,26 @@ def check_duplicates_files(pc_recs, platform, duplicate_files="error"):
...
         if len(files) > 1:
-            msg_str = "Files %s found in the package(s): %s" % (
-                str(files)[1:-1], ', '.join(fns))
-            if duplicate_files == "warn" or platform.startswith('linux'):
-                print('Warning: {}'.format(msg_str))
+            msg_str = "Files {} found in the package(s): {}".format(
```
jaimergp, line 195: f-string?

constructor/main.py
```diff
@@ -52,14 +51,16 @@ def get_output_filename(info):
...
-    ext = info['installer_type']
-    return '%s-%s-%s.%s' % ('%(name)s-%(version)s' % info,
-                            os_map.get(osname, osname),
-                            arch_name_map.get(arch, arch),
-                            ext)
+    osname, arch = info["_platform"].split("-")
+    os_map = {"linux": "Linux", "osx": "MacOSX", "win": "Windows"}
+    arch_name_map = {"64": "x86_64", "32": "x86"}
+    ext = info["installer_type"]
+    return "{}-{}-{}.{}".format(
```
jaimergp, line 58: Wow, this one is a mouthful 😂

constructor/main.py
```diff
@@ -52,14 +51,16 @@ def get_output_filename(info):
...
+        "%(name)s-%(version)s" % info,
```
jaimergp, line 59: Would an f-string make it too verbose?

constructor/nsis/_nsis.py
```diff
@@ -47,8 +47,8 @@ def gui_excepthook(exctype, value, tb):
         import ctypes
         import traceback
         MB_ICONERROR = 0x00000010
-        title = u'Installation Error'
-        msg = u''.join(traceback.format_exception(exctype, value, tb))
+        title = "Installation Error"
```
jaimergp, line 50: We should ignore this file.

It's executed at runtime by the Python interpreter included in the `base` env that the installer provides. If we pyupgrade this, we would prevent people from using `constructor` to deploy Py2k installations.

Not ideal, and the installer logic should be uncoupled from the installer contents (working on it via conda-standalone), but for now no modifications in this file.

constructor/nsis/_system_path.py
```diff
@@ -73,7 +73,7 @@ def remove_from_system_path(pathname, allusers=True, path_env_var='PATH'):
         reg_value = None
         try:
             reg_value = reg.QueryValueEx(key, path_env_var)
-        except WindowsError:
```
jaimergp, line 76: Same thing as `_nsis.py`. Can't touch this.

constructor/preconda.py
```diff
@@ -171,7 +171,7 @@ def write_conda_meta(info, dst_dir, final_urls_md5s, user_requested_specs=None):
 
     cmd = path_split(sys.argv[0])[-1]
     if len(sys.argv) > 1:
-        cmd = "%s %s" % (cmd, " ".join(sys.argv[1:]))
+        cmd = "{} {}".format(cmd, " ".join(sys.argv[1:]))
```
jaimergp, line 174: f-string?

jaimergp (changes requested):
LGTM but we need to revert the changes in `_nsis.py` and same-dir friends.

dbast:
de-scoped this now a bit:
* excluded the nsis/*.py files
* disabled percent-format to .format/f-string conversions
* removed darker for less noise

much smaller easier to review diff.

constructor/preconda.py
dbast: disabled... requires manual conversion

constructor/nsis/_system_path.py
dbast: excluded

constructor/nsis/_nsis.py
dbast: excluded

constructor/main.py
dbast: disabled... requires manual conversion

constructor/main.py
dbast: disabled... requires manual conversion

constructor/fcp.py
dbast: disabled... requires manual conversion

constructor/fcp.py
dbast: disabled... requires manual conversion

constructor/construct.py
dbast: agreed ... the exception handler can then also convert different exceptions into different exit codes

constructor/construct.py
dbast: yep

.pre-commit-config.yaml
```diff
@@ -16,6 +16,12 @@ repos:
       - id: mixed-line-ending
       - id: end-of-file-fixer
       - id: trailing-whitespace
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v3.3.1
+    hooks:
+      - id: pyupgrade
+        args: ["--py37-plus", "--keep-percent-format"]
+        exclude: ^(constructor/nsis/.*py)
```
jaimergp, line 24:
```suggestion
        exclude: ^(constructor/nsis/*.py)
```

Did you mean to type this? :D

dbast: yes... this also confused me... it is not a glob pattern (then *.sh would be good), but a regex (where "." matches a single character 0 or any times "*").

jaimergp: Ah, wow, hahaha ok ok then this is ok

jaimergp (dismissed):

jaimergp:
LGTM. Let's merge #605 before though in case there are conflicts.

jaimergp:
Add / edit the relevant `news` item and we are good to go!

dbast:
news entry is there now.

news/606-require-py37
```diff
@@ -0,0 +1,19 @@
...
+
+* <news item>
+
+### Bug fixes
+
+* <news item>
+
+### Deprecations
+
+* Officially require Python>=3.7 via setup.py. Older Python versions are EOL and not part of the test matrix since #479.
```
jaimergp, line 11:
```suggestion
* Officially require Python>=3.7 via setup.py. Older Python versions are EOL and not part of the test matrix since #479. (#606)
```

jaimergp (approved):

PR #636, opened 2023-01-28
Fix undefined variable causing failure

### LINKED ISSUE #635: Fix bugs in interactive mode due to unbound variables
### Checklist

### What happened?

This https://github.com/conda/constructor/pull/600 enabled `set -u` which leads to unbound variables in the interactive mode of the installers, e.g. the recent released https://github.com/conda-forge/miniforge/releases/download/22.11.1-0/Miniforge3-22.11.1-0-MacOSX-arm64.sh ... This was not caught during automated testing as automated testing is done in batch mode.

Errors are like:
```
installation finished.
Do you wish the installer to initialize Miniforge3
by running conda init? [yes|no]
Miniforge3-22.11.1-0-MacOSX-arm64.sh: line 394: DEFAULT: unbound variable
```

and 

```
Please, press ENTER to continue
>>> 
Miniconda3-py39_XXXXXX.sh: line 126: $1: unbound variable
```

### Conda Info

_No response_

### Conda Config

_No response_

### Conda list

_No response_

### Additional Context

_No response_

hmaarrfk:
Closes #635

### Checklist - did you ...

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -531,7 +531,6 @@ if [ "${PYTHONPATH:-}" != "" ]; then
 fi
 
 if [ "$BATCH" = "0" ]; then
-    DEFAULT=no
 #if initialize_conda is True and initialize_by_default is True
     DEFAULT=yes
 #endif
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -537,6 +537,9 @@ if [ "$BATCH" = "0" ]; then
 #if initialize_conda is True and initialize_by_default is False
     DEFAULT=no
 #endif
+#if initialize_conda is False
+    DEFAULT=no
+#endif
 
 #if has_conda and initialize_conda is True
     # Interactive mode.
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -531,13 +531,11 @@ if [ "${PYTHONPATH:-}" != "" ]; then
 fi
 
 if [ "$BATCH" = "0" ]; then
+    DEFAULT=no
 #if initialize_conda is True and initialize_by_default is True
     DEFAULT=yes
 #endif
 #if initialize_conda is True and initialize_by_default is False
-    DEFAULT=no
-#endif
-#if initialize_conda is False
     DEFAULT=no
 #endif
 
--- a/news/636-fix-init-conda
+++ b/news/636-fix-init-conda
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fix interactive installation in Linux and MacOX by ensuring user default is initialized.
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* <news item>
--- a/news/636-fix-init-conda
+++ b/news/636-fix-init-conda
@@ -4,7 +4,7 @@
 
 ### Bug fixes
 
-* Fix interactive installation in Linux and MacOX by ensuring user default is initialized.
+* Fix interactive installation in Linux and MacOX by ensuring user default is initialized. (fixes #635 via #636)
 
 ### Deprecations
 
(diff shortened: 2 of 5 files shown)


hmaarrfk:
Why is the default value None?

hmaarrfk:
https://github.com/conda/constructor/blob/main/constructor/shar.py#L69

dbast:
@hmaarrfk Thanks for debugging this.  `initialize_by_default` should be a boolean defaulting to False as stated here https://github.com/conda/constructor/blob/main/constructor/construct.py#L427 .. so defaulting to None should be a bug.

constructor/header.sh
```diff
@@ -537,6 +537,9 @@ if [ "$BATCH" = "0" ]; then
 #if initialize_conda is True and initialize_by_default is False
     DEFAULT=no
 #endif
+#if initialize_conda is False
+    DEFAULT=no
+#endif
```
dbast, line 542: This is not fixing it as initialize_by_default has the None value and `initialize_conda is True`... see this debug build https://github.com/dbast/constructor/actions/runs/4030769478/jobs/6929654326#step:12:167

dbast:
@hmaarrfk The code should be correct now with if/else independent DEFAULT=no for protection.

Also have seen your interactive test in the miniforge repo ... thinking about doing that also here.

news/636-fix-init-conda
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fix interactive installation in Linux and MacOX by ensuring user default is initialized.
```
dbast, line 7:
```suggestion
* Fix interactive installation in Linux and MacOX by ensuring user default is initialized. (fixes #635 via #636)
```

constructor/header.sh
```diff
@@ -531,6 +531,7 @@ if [ "${PYTHONPATH:-}" != "" ]; then
 fi
 
 if [ "$BATCH" = "0" ]; then
+    DEFAULT=no
```
hmaarrfk, line 534: I guess this makes writing these "if" statements easier, but I think that the "generated code" looks uglier.

I do personally prefer the addition of the full if statement:

```
#if initialize_conda is False
   DEFAULT=no
#endif
````

Comment (dbast) [reply constructor/header.sh:534]: The DEFAULT variable is not used in the case of `initialize_conda is False`, see if statement in line 541/542... so also "ugly" :/ ... ` ... anyways no change at all required for the header.sh due to `initialize_by_default` now being initialised to False instead of None. 

A general pattern has to be enabled in the code so that the values which are read via `info.get("name", "default")` actually have a default and type defined in the construct.py.

Comment (dbast) [inline constructor/header.sh:534]:
```suggestion
```

hmaarrfk: understood. then lets not modify the header.sh at all :D.

dbast:
Intend to merge this soon, opened https://github.com/conda/constructor/issues/637 to improve testing like done in the miniforge repo.

dbast:
oh. there is a potential other solution (instead of setting the parameter default value from None to False):
```
#if initialize_conda is True and initialize_by_default is None
    DEFAULT=""
#endif
```

I like the current solution of setting the parameter default value from None to False more as that is then more in line with the documentation of the initialize_by_default parameter (bool defaulting to False). What speaks for the None solution is that the nsis template also has a no behaviour https://github.com/conda/constructor/blob/main/constructor/nsis/main.nsi.tmpl#L622

@jaimergp What do you think?

dbast (approved):

proppy:
will this eventually land in https://anaconda.org/main/constructor/files ?

jaimergp:
3.4.3 is out now I think, up to the Anaconda team to package it! Check https://github.com/AnacondaRecipes/constructor-feedstock/pulls

PR #639, opened 2023-01-31
Report example runs as integration test coverage

dbast:
### Description

run_examples.py automatically runs constructor with coverage if the coverage Python module is importable (so installing codecov -> coverage is enough). Also moving the codecov upload from inside the recipe to the action as there is much more  meta-data that gets reported that way (PR numbers etc).

This pushes the coverage from currently ~20% to >85%, see https://app.codecov.io/gh/conda/constructor/pull/639 and allows to either see what is not covered yet by tests or what is potentially dead code.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -71,7 +71,6 @@ jobs:
           conda create -p ../conda conda-build conda-verify
       - name: Build the package
         env:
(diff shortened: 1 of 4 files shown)


dbast:
Something is strange when looking at the report ... put PR in draft mode... maybe that recovers after CI finishes.

PR #664, opened 2023-04-26
Fix yes no parsing

marcoesters:
### Description

Improve yes/no parsing for interactive installation with sh installers.

* Make answers case insensitive
* Accept default value for license agreement question

### Checklist - did you ...

### DIFF
--- a/.github/workflows/issues.yml
+++ b/.github/workflows/issues.yml
@@ -28,7 +28,7 @@ jobs:
           labels: ${{ env.FEEDBACK_LBL }}
           github_token: ${{ secrets.PROJECT_TOKEN }}
       # add [pending::support], if still open
-      - uses: actions-ecosystem/action-add-labels@v1.1.0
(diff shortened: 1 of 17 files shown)


pseudoyim (commented):
I built an example MinicondaX using @marcoesters's dev branch, and the default `[no]` now gets applied when I hit Enter, and the installer exits as expected.

Screen Shot 2023-05-03 at 4 28 26 PM copy

I noticed two of the CI builds are failing on the example builds, but it's not clear what the causes are ([here](https://github.com/conda/constructor/actions/runs/4811442769/jobs/8565451687?pr=664#step:13:186) and [here](https://github.com/conda/constructor/actions/runs/4811442769/jobs/8565451863?pr=664#step:13:69)). Perhaps they just need a re-punch?

jaimergp (approved):
Seeing this code appear so often in the installer makes me wonder whether we should put it in a function, but that can be done in a different PR.

Thanks!

Issue #612, opened 2023-01-17
Migrate from shared ctools anaconda.org channel to own constructor channel


dbast:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

.

PR #675, opened 2023-05-12
Run examples using `pytest` + optimize CI setup

### LINKED ISSUE #641: Convert the test suite to 100% pytest-based
### Checklist

### What is the idea?

Our current test infrastructure uses some pytest logic, but the main logic happens in the `run_examples.py` script, which depends on the content of `examples/`. This is a homemade, ad-hoc script that has grown a bit too much, with many features and cases. Every time we add a new variable to check it's shoehorned in and it is not getting any better.

I suggest we use a `pytest`-only test infra, powered by fixtures and other pytest goodies. The good folks at MNE (@larsoner more specifically) did suggest something along these lines a while ago ([see comment here](https://github.com/conda/constructor/pull/525#issuecomment-1177568197) for more details and ideas).

This would allow us to test better, clearer and more efficiently. Debugging should also get easier thanks to the `-k` flag.

### Why is this needed?

* Better development experience
* Long-term maintainability
* Readability of the test infra

### What should happen?

* All the examples should be covered by a pytest case, with a docstring explaining what we are aiming for in that test.
* Post-installation checks wouldn't need the `post_install.sh` hacks, and could be done with pytest idioms. Testing with Bash makes me feel bad :D

### Additional Context

_No response_

jaimergp:
### Description

Closes #641 

### Checklist - did you ...


.github/workflows/main.yml
```diff
@@ -20,201 +20,143 @@ concurrency:
...
-      - name: Build docs
-        run: |
-          source $CONDA/bin/activate
-          conda activate constructor
-          set -ex
-          python scripts/make_docs.py
-          git diff --exit-code
+          package-name: ${{ github.event.repository.name }}
+          subdir: ${{ matrix.subdir }}
+          anaconda-org-channel: ctools
```
jaimergp, line 159: Should we change to `conda-canary` here? xref https://github.com/conda/constructor/issues/612

constructor/osxpkg.py
```diff
@@ -20,6 +21,17 @@
 logger = logging.getLogger(__name__)
 
 
+def calculate_install_dir(yaml_file, subdir=None):
```
jaimergp, line 24: Utility function needed in tests.

.github/workflows/main.yml
larsoner: Still TODO?

tests/test_examples.py
```diff
@@ -0,0 +1,386 @@
...
+            "Possible causes: permissions, non-supported characters, long paths...\n"
+            "Consider setting 'check_path_spaces' and 'check_path_length' to 'False'."
+        )
+    if error_lines:
+        raise AssertionError("\n".join(error_lines))
+
+
+def _run_uninstaller_exe(install_dir):
+    # Now test the uninstallers
+    if " " in str(install_dir):
```
larsoner, line 121: Better to raise an exception here? Send safer than silently allowing the test to pass (assuming that's the outcome of short circuiting here)

tests/test_examples.py
```diff
@@ -0,0 +1,386 @@
...
+        _run_installer(input_path, installer, install_dir, request=request)
+
+
+def test_example_scripts(tmp_path, request):
+    input_path = _example_path("scripts")
+    for installer, install_dir in create_installer(input_path, tmp_path, with_spaces=True):
+        _run_installer(input_path, installer, install_dir, request=request)
+
+
+def test_example_shortcuts(tmp_path, request):
```
larsoner, line 357: For these I would have expected a `parametrize` over the across the `_example_path`s. Assuming that function returns filenames quickly it seems fine to refactor this way and would give more complete failure reporting for example if a change breaks multiple examples. It also would allow selecting tests by the example filename because that would show up in the resulting test name

larsoner (commented):
Already looks like a great improvement! Just a few comments

tests/test_examples.py
jaimergp: That's a known issue we couldn't fix in CI (works locally), but I agree we could at least add a warning!

tests/test_examples.py
```diff
@@ -0,0 +1,390 @@
...
+    _execute(cmd)
+
+    # Windows installers won't raise exit codes so we need to check the log file
+    error_lines = []
+    try:
+        log_is_empty = True
+        with open(os.path.join(install_dir, "install.log"), encoding="utf-16-le") as f:
+            for line in f:
+                log_is_empty = False
+                if ":error:" in line:
```
marcoesters, line 104: Do we want to use `line.lower()` to make ourselves case-independent?

marcoesters (dismissed):
I've been going back and forth on whether to change the boiler-plate code into paramatrized tests.

However, I think Jaime has a point that there is potential that the tests become more involved, especially if there are plans to replace the `test_install.*` scripts.

It's a good first step. However, I think my comment should be implemented before merging.

tests/test_examples.py
jaimergp:
```suggestion
                if ":error:" in line.lower():
```

jaimergp: Done, thanks! Unfortunately I'll need the approval again, as it gets automatically dismissed with new changes 😬

jezdez (approved):

Issue #588, opened 2022-12-05
Investigate the usage of `conda init --reverse` in uninstallers


jaimergp:
Why not use the same method as conda init --reverse? That might be more robust to upstream pattern change.
Moreover, it seems that current method will break AutoRun with value like `<Conda AutoRun> & <Other App AutoRun>`, by substituting it to `& <Other App AutoRun>`. But it is invalid to have beginning `&` in CMD.exe

_Originally posted by @YouJiacheng in https://github.com/conda/constructor/issues/521#issuecomment-1246166684_

jaimergp:
not stale

PR #684, opened 2023-06-15
Compare registry values to `$INSTDIR` before removing keys

marcoesters:
### Description

The current uinstaller relies on values set during compile time to delete registry keys. Changes to the registry after the installation will not be taken into account when performing the uninstallation.

This PR scans the registry during uninstallation and deletes the entries that can be directly tied to the installation through `$INSTDIR`.

### Checklist - did you ...


jaimergp:
I like this better than what we have right now, but I wonder if we should pursue #588 after this. That'd mean replacing the logic in NSIS with the logic in `conda-standalone` (aka `conda init`), for both installation and uninstallation.

This way we make `constructor` more conda-agnostic. Of course, outside the scope of this PR.

news/684-uninstaller-check-registry
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Uninstaller checks the registry for INSTDIR before deleting registry keys (#684)
```
jaimergp, line 3:
```suggestion
* The Windows uninstaller will check the registry for `$INSTDIR` before deleting hardcoded registry keys (#684)
```

constructor/nsis/main.nsi.tmpl
```diff
@@ -1185,10 +1187,36 @@ Section "Uninstall"
     #   carefully.  More info at https://docs.conda.io/projects/conda/en/latest/user-guide/troubleshooting.html#solution
     System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_DLL_SEARCH_MODIFICATION_ENABLE", "1").r0'
 
+    # Read variables the uninstaller needs from the registry
```
jaimergp, line 1190: If I understand this correctly, this block will search the registry for two variables:

- INSTALLER_NAME_FULL
- INSTALLER_VERSION

These were previously assumed to be constant and would have used the hardcoded values at install time (`${NAME}` [used in [line 65](https://github.com/conda/constructor/blob/fa2040b199eafe4fb8488df30eeac43f26a951da/constructor/nsis/main.nsi.tmpl#L65)] and `${VERSION}`, respectively).

However we are now saying that this might not be always the case. Could you elaborate under what circumstances this is true? e.g. updating Anaconda in place but not updating the registry keys (or not deleting the old ones before adding the new ones)?

constructor/nsis/main.nsi.tmpl
```diff
@@ -1202,13 +1230,26 @@ Section "Uninstall"
...
+        DeleteRegKey SHCTX "SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\$INSTALLER_NAME_FULL"
+    ${EndIf}
+
     # If Anaconda was registered as the official Python for this version,
     # remove it from the registry
-    ReadRegStr $0 SHCTX "Software\Python\PythonCore\${PY_VER}\InstallPath" ""
-    ${If} $0 == "$INSTDIR"
-        DeleteRegKey SHCTX "Software\Python\PythonCore\${PY_VER}"
-    ${EndIf}
+    StrCpy $R0 "SOFTWARE\Python\PythonCore"
```
jaimergp, line 1239: This block is easier to understand, I think. Basically we don't want to assume that the `base` Python is still the same version. However, is this updated when `conda` updates the `python` package in `base`? I don't remember seeing code for this 🤔 It would still be the same path, so it would work, but maybe the metadata is outdated when it comes to the reported version in the registry.

constructor/nsis/main.nsi.tmpl
jaimergp: I guess this logic allows the uninstaller to run for _other_ conda installations if needed, I don't know if that's a bug or a feature 😬

marcoesters: > I guess this logic allows the uninstaller to run for _other_ conda installations if needed, I don't know if that's a bug or a feature 😬

That is already possible. The changes in this PR will, however, conduct checks to make sure the deletion is performed consistently on the file system and the registry.

There is another, admittedly nice, application: if you install the same Miniconda/Miniforge version twice, installation 2 will overwrite the registry keys of number 1. If you then run the uninstaller of number 1, it will delete the keys of number 2.

constructor/nsis/main.nsi.tmpl
marcoesters: This also prevents issues in cross-installer usage, as described in my other comment.

I do not think that `conda` updates the registry when python is updated, which it should probably do.

jaimergp (approved):

Issue #674, opened 2023-05-12
`miniforge` example failing for `micromamba` x `macos`


jaimergp:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

The CI has been failing for a while on the same example:

```
--- Testing Miniforge3-4.10.1-0-MacOSX-x86_64.sh
/bin/sh /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmpcef2verr/miniforge-xho1us3v/Miniforge3-4.10.1-0-MacOSX-x86_64.sh
--- RETURNCODE: 1 ---
--- STDOUT ---

Welcome to Miniforge3 4.10.1-0

In order to continue the installation process, please review the license
agreement.
Please, press ENTER to continue
>>> Copyright (c) 2016, Example, Inc.
All rights reserved.

Redistribution and use in source and binary forms, with or without
modification, are permitted provided that the following conditions are met:
    * Redistributions of source code must retain the above copyright
      notice, this list of conditions and the following disclaimer.
    * Redistributions in binary form must reproduce the above copyright
      notice, this list of conditions and the following disclaimer in the
      documentation and/or other materials provided with the distribution.
    * Neither the name of Example, Inc. nor the
      names of its contributors may be used to endorse or promote products
      derived from this software without specific prior written permission.

THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
DISCLAIMED. IN NO EVENT SHALL EXAMPLE, INC. BE LIABLE FOR ANY
DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
[...]

PR #692, opened 2023-07-07
Fix transmutation with micromamba

jaimergp:
### Description

Closes https://github.com/conda/constructor/issues/674.

### Checklist - did you ...

### DIFF
--- a/constructor/preconda.py
+++ b/constructor/preconda.py
@@ -133,8 +133,11 @@ def write_files(info, dst_dir):
 
     with open(join(dst_dir, 'urls'), 'w') as fo:
         for url, md5 in all_final_urls_md5s:
-            url = ensure_transmuted_ext(info, url)
-            fo.write('%s#%s\n' % (url, md5))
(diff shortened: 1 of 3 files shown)


jezdez (approved):

jaimergp:
I looked a bit into this and fixed, I think:

- Problems started somewhere between Apr 3rd and 21st (see [this PR](https://github.com/conda/constructor/pull/660) for the first occurrence). It used to work before, but there were no significant releases in that time that explain the problem.
- The problem is not about extraction this time, but hash verification. Basically the extracted archives will have a different hash in their `info/repodata_record.json` if they have been transmuted, and micromamba doesn't like that, so it thinks that the cache is corrupted and tries to download the artifact again. The solution is to remove the md5 for those files so micromamba doesn't check. 
- To summarize, the URL and md5 in the explicit input file must match the ones present in the repodata_record.json on the corresponding info/ directory. Otherwise it constitutes a cache mismatch.

jezdez (approved):

PR #702, opened 2023-07-20
Skip test_write_images on Linux

marcoesters:
### Description

Writing images is only supported on Windows and MacOS, but pytest will currently run it on any system.

* Skip `test_write_images` if the platform is not Windows or MacOS

### Checklist - did you ...

### DIFF
--- a/tests/test_imaging.py
+++ b/tests/test_imaging.py
@@ -10,7 +10,7 @@ if sys.platform == "win32" or sys.platform == "darwin":
 
 @pytest.mark.skipif(
     sys.platform != 'win32' and sys.platform != 'darwin',
-    reason='imaging not available on Linux'
+    reason='imaging only available on Windows and MacOS'
 )
 def test_write_images():
     tmp_dir = tempfile.mkdtemp()
--- a/tests/test_imaging.py
+++ b/tests/test_imaging.py
@@ -1,9 +1,15 @@
+import pytest
 import shutil
+import sys
 import tempfile
 
 from constructor.imaging import write_images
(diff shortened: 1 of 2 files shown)


tests/test_imaging.py
```diff
@@ -1,9 +1,16 @@
...
+import sys
 import tempfile
 
+import pytest
+
 from constructor.imaging import write_images
 
 
+@pytest.mark.skipif(
+    sys.platform != 'win32' and sys.platform != 'darwin',
```
jaimergp, line 11: Why not?

```suggestion
    sys.platform == "linux",
```

tests/test_imaging.py
```diff
@@ -1,9 +1,16 @@
 import shutil
+import sys
 import tempfile
 
+import pytest
+
 from constructor.imaging import write_images
```
jaimergp, line 7: Wouldn't this trigger an import error for pillow? We might want to guard this import.

jaimergp (commented):
LGTM, but see comments and please add the news file. Thanks!

tests/test_imaging.py
marcoesters: You are correct, I added a guard.

tests/test_imaging.py
marcoesters: I did it because `sys.platform` can have more values than "win32", "darwin", and "linux". I don't mind changing it if you think this is better.

I did adjust the wording for the reason though.

jaimergp (approved):

PR #711, opened 2023-08-14
Add CI tests for conda-standalone canaries

jaimergp:
### Description

We've been doing some work over at `conda/conda-standalone` so it can be maintained like a proper open source project. This means we have now canaries we can use to test new versions with constructor.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -40,41 +40,41 @@ jobs:
         include:
           # UBUNTU
           - os: ubuntu
-            python-version: 3.8
+            python-version: "3.8"
             conda-standalone: conda-standalone
             check-docs: true
           - os: ubuntu
-            python-version: 3.9
+            python-version: "3.9"
             conda-standalone: conda-standalone
             check-docs: true
           - os: ubuntu
-            python-version: 3.10
+            python-version: "3.10"
             conda-standalone: conda-standalone-nightly
           - os: ubuntu
-            python-version: 3.11
+            python-version: "3.11"
             conda-standalone: micromamba
           # MACOS
           - os: macos
-            python-version: 3.9
+            python-version: "3.9"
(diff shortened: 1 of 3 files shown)


jaimergp:
Everything seems to work. Let's add it to the CI.

.github/workflows/main.yml
```diff
@@ -31,28 +32,55 @@ defaults:
...
-            micromamba: true
+            conda-standalone: conda-standalone
             check-docs: true
+          - os: ubuntu
+            python-version: "3.10"
+            conda-standalone: conda-standalone-nightly
+          - os: ubuntu
+            python-version: "3.11"
+            conda-standalone: micromamba
+          # MACOS
```
kenodegard, line 56: are you intentionally skipping Python 3.8 for macOS?

kenodegard (dismissed):

.github/workflows/main.yml
jaimergp: Accidental at first, then I thought, hm why not? macOS runners are scarce after all. Then I asked in the call whether we can drop or not for `conda/conda`. I guess I'll add it back?

kenodegard: In conda-build we opted to only test the lower and upper versions on macOS. My thought behind that was we'd catch stdlib deprecations with the upper version and we'd catch usage of newer stdlib features that the lower version doesn't support.

Dunno how relevant that would be for `constructor`, just wanted to be sure the exclusion was intentional. 👍🏻

jaimergp: I like that. Will apply changes and ping you again. Thanks!

jaimergp: Done @kenodegard. Can you reapprove? Thanks 🙏

kenodegard (approved):

PR #713, opened 2023-08-23
Clarify wording for running conda init

marcoesters:
### Description

When asking whether to execute `conda init`, there is no explanation about what the consequences of the decision are. This can surprise new users to conda when their prompt is changed after restarting their shell.

Update the prompt to describe what is being changed and how it can be reversed before making a decision.

### Checklist - did you ...

### DIFF
--- a/constructor/osxpkg.py
+++ b/constructor/osxpkg.py
@@ -224,10 +224,13 @@ def modify_xml(xml_path, info):
                 'initialize_by_default', True) else 'false')
             path_choice.set('title', "Add conda initialization to the shell")
             path_description = """
-            If this box is checked, "conda init" will be executed to ensure that
-            conda is available in your preferred shell upon startup. If unchecked,
-            you must this initialization yourself or activate the environment
-            manually for each shell in which you wish to use it."""
+            If this box is checked, conda will be automatically activated in your
+            preferred shell on startup. This will change the command prompt when
+            activated. If your prefer that conda's base environment not be activated
+            on startup, run `conda config --set auto_activate_base false`. You can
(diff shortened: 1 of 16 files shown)


jaimergp (commented):
This is awesome, thanks! We should probably add it to PKG and EXEs UIs too, but we don't have too many characters there 😬

marcoesters:
> This is awesome, thanks! We should probably add it to the [PKG UI too](https://github.com/marcoesters/constructor/blob/51485ead6a825a553abcd2d9a4e3b5c7b1130a4d/constructor/osxpkg.py#L220-L231).

Good call! I expanded that text, too

constructor/osxpkg.py
```diff
@@ -224,10 +224,13 @@ def modify_xml(xml_path, info):
...
             path_description = """
-            If this box is checked, "conda init" will be executed to ensure that
-            conda is available in your preferred shell upon startup. If unchecked,
-            you must this initialization yourself or activate the environment
-            manually for each shell in which you wish to use it."""
+            If this box is checked, conda will be automatically activated in your
+            preferred shell on startup. This will change the command prompt when
+            activated. If your prefer that conda's base environment not be activated
+            on startup, run `conda config --set auto_activate_base false`. You can
+            undo this by running `conda init --reverse ${SHELL}`.
```
jaimergp, line 231: Is `${SHELL}` expected to be defined here or is it just a placeholder? I am assuming that it will be rendered in the SH installers but I am not sure about the PKGs. If it's meant to be a placeholder, that's ok too, just wanted to make sure it was not an oversight.

jaimergp (approved):

PR #715, opened 2023-08-27
Debug system calls on osx

m-kuhn:
This will log any external calls invoked on osx (with level "debug").
I found it quite handy to have a look at what `pkgbuild` commands are actually executed.

### DIFF
--- a/constructor/osxpkg.py
+++ b/constructor/osxpkg.py
@@ -366,7 +366,7 @@ def pkgbuild_prepare_installation(info):
         payload.set("installKBytes", str(approx_pkgs_size_kb))
         tree.write(payload_xml)
         # repack
-        check_call(["pkgutil", "--flatten", f"{pkg}.expanded", pkg])
+        explained_check_call(["pkgutil", "--flatten", f"{pkg}.expanded", pkg])
         return pkg
     finally:
         shutil.rmtree(f"{pkg}.expanded")
@@ -459,7 +459,7 @@ def create(info, verbose=False):
                 "com.apple.security.cs.allow-dyld-environment-variables": True,
             }
             plist_dump(plist, f)
-        check_call(
+        explained_check_call(
             [
                 # hardcode to system location to avoid accidental clobber in PATH
                 "/usr/bin/codesign",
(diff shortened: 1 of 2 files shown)


constructor/osxpkg.py
```diff
@@ -21,6 +21,13 @@
 logger = logging.getLogger(__name__)
 
 
+def check_call(args):
+    """
+    Execute a system process and debug the invocation
+    """
+    logger.debug(f"Executing: {' '.join(args)}")
```
jaimergp, line 28:
```suggestion
    logger.debug("Executing: %s", " ".join(args))
```

constructor/osxpkg.py
```diff
@@ -21,6 +21,13 @@
 logger = logging.getLogger(__name__)
 
 
+def check_call(args):
```
jaimergp, line 24:
```suggestion
def explained_check_call(args):
```

jaimergp (commented):
LGTM. Can you move the helper function to `constructor.utils` and import from there? Thanks!

constructor/osxpkg.py
m-kuhn: Just for my understanding, is this for consistency with the existing codebase or is there a reason to stay python <3.6 compatible?

jaimergp:
Looks like there's a missing import on `osxpkg`

constructor/utils.py
```diff
@@ -12,12 +12,20 @@
...
 from ruamel import yaml
 
 logger = logging.getLogger(__name__)
 
 
+def explained_check_call(args):
+    """
+    Execute a system process and debug the invocation
+    """
+    logger.debug(f"Executing: {' '.join(args)}")
```
jaimergp, line 26:
```suggestion
    logger.debug("Executing: %s", " ".join(args))
```

jaimergp:
pre-commit.ci autofix

constructor/osxpkg.py
jaimergp: Consistency and best practices with `logging`.

jaimergp:
Your fork doesn't allow us to commit directly to the branch so pre-commit can't autofix, sorry. You'll have to address the requested changes locally.

jaimergp (approved):
Thank you!

PR #718, opened 2023-09-13
Added support for `extra_env_variables`

jlstevens:
Adds support for an `extra_env_variables` key in the `construct.yaml` e.g.:

```yaml
extra_env_variables:
  - CUSTOM_VARIABLE_1=CUSTOM1
  - CUSTOM_VARIABLE_2=CUSTOM2
```

This makes these environment variables available to the `pre_install` and `post_install` scripts.

***Motivation***: This mechanism makes it much easier to template installers that use `pre_install` and `post_install` scripts, allowing those scripts to stay generic and unchanged while making the templated changes clear in the `construct.yaml` in an easy-to-read, declarative form.

Right now only Unix .sh installers are supported in this PR. I wanted to see whether to constructor maintainers are open to this addition before implementing support for other platforms.

The only open question I have right now is whether there is a use-case to allow custom overrides of the pre-defined variables (i.e. `$INSTALLER_NAME`, `$INSTALLER_VER`, `$INSTALLER_PLAT`). My instinct is that there is no good reason to allow this so this PR ensures that custom extra variable definitions cannot override these definitions.

### DIFF
--- a/constructor/shar.py
+++ b/constructor/shar.py
@@ -92,8 +92,10 @@ def get_header(conda_exec, tarball, info):
 
     data = read_header_template()
     data = preprocess(data, ppd)
+    custom_variables = info.pop('extra_env_variables', [])
     data = fill_template(data, replace)
 
+    data = data.replace("_EXTRA_ENV_VARIABLES_=''", '\n'.join([f'export {var}' for var in custom_variables]))
     return data
 
 
--- a/examples/scripts/construct.yaml
+++ b/examples/scripts/construct.yaml
@@ -5,6 +5,11 @@ channels:
   - http://repo.anaconda.com/pkgs/main/
 specs:
   - python
+
+extra_env_variables:
+  - CUSTOM_VARIABLE_1=CUSTOM1
+  - CUSTOM_VARIABLE_2=CUSTOM2
+
 pre_install: pre_install.sh   # [unix]
 pre_install: pre_install.bat  # [win]
 pre_install_desc: "Adding this description makes the script selectable in the UI"
--- a/CONSTRUCT.md
+++ b/CONSTRUCT.md
@@ -442,6 +442,18 @@ is compulsory and the option to disable it will not be offered.
 
 This option has no effect on `SH` installers.
 
+
+### `extra_env_variables`
+
+_required:_ no<br/>
+_type:_ list<br/>
+
+List of additional environment variables to be made available to the
+pre_install and post_install scripts, in the form of VAR=VALUE
+pairs. These environment variables are in addition to those in the
+`post_install` section above and take precedence in the case of name
+collisions.  Unix only.
+
 ### `pre_uninstall`
 
 _required:_ no<br/>
--- a/constructor/construct.py
+++ b/constructor/construct.py
@@ -270,6 +270,13 @@ provided, it will default to `io.continuum`. (MacOS only)
 Application name in the Windows "Programs and Features" control panel.
 Defaults to `${NAME} ${VERSION} (Python ${PYVERSION} ${ARCH})`.
 '''),
+('extra_env_variables',           False, list, '''
+List of additional environment variables to be made available to the
+pre_install and post_install scripts, in the form of VAR=VALUE
+pairs. These environment variables are in addition to those in the
+`post_install` section above and take precedence in the case of name
+collisions.  Unix only.
+'''),
 
     ('pre_install',            False, str, '''
 Path to a pre-install script, run after the package cache has been set, but
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -23,6 +23,8 @@ fi
 
 # Export variables to make installer metadata available to pre/post install scripts
 # NOTE: If more vars are added, make sure to update the examples/scripts tests too
+
+_EXTRA_ENV_VARIABLES_=''  # Templated extra environment variable(s) 
 export INSTALLER_NAME='__NAME__'
 export INSTALLER_VER='__VERSION__'
 export INSTALLER_PLAT='__PLAT__'
--- a/examples/scripts/post_install.sh
+++ b/examples/scripts/post_install.sh
@@ -9,10 +9,15 @@ echo "INSTALLER_NAME=${INSTALLER_NAME}"
 echo "INSTALLER_VER=${INSTALLER_VER}"
 echo "INSTALLER_PLAT=${INSTALLER_PLAT}"
 echo "INSTALLER_TYPE=${INSTALLER_TYPE}"
+echo "CUSTOM_VARIABLE_1=${CUSTOM_VARIABLE_1}"
+echo "CUSTOM_VARIABLE_2=${CUSTOM_VARIABLE_2}"
 echo "PREFIX=${PREFIX}"
 
 test "${INSTALLER_NAME}" = "Scripts"
 test "${INSTALLER_VER}" = "X"
+test "${CUSTOM_VARIABLE_1}" = "CUSTOM1"
+test "${CUSTOM_VARIABLE_2}" = "CUSTOM2"
+
 if [[ $(uname -s) == Linux ]]; then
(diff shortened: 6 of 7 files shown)


jaimergp:
Thanks for bringing this up in such an accommodating way!

I am not opposed to the idea, but:

- It indeed needs to be ported to other platforms
- We need to have a well defined syntax and be clear about what's supported and what's not (e.g. referencing other variables, escaping characters, etc)
- We need to be very careful about spaces, quotes, special characters like `$`, which might have different rules in different platforms.
- The field name should be clear about only affecting the pre and post install scripts, not the resulting installation.

In general I wonder if this is better solved with an example that documents how to run a pre-processing script to template your pre and post scripts on your own, without direct support in `constructor`. Given the amount of edge cases that can arise, it's tricky to come up with a generalized approach that works in all cases.

jlstevens:
> I am not opposed to the idea

Glad to hear it!

> It indeed needs to be ported to other platforms

Agreed. My last commit adds support to Windows.

> We need to have a well defined syntax and be clear about what's supported and what's not (e.g. referencing other variables, escaping characters, etc). We need to be very careful about spaces, quotes, special characters like $, which might have different rules in different platforms.

Yes, we will need clear documentation. I think it would be fine to only support simple literals initially and apply strict validation of the construct.yaml.

> The field name should be clear about only affecting the pre and post install scripts, not the resulting installation.

Got any suggestions? Would `script_extra_env_variables` be a better name?

> In general I wonder if this is better solved with an example that documents how to run a pre-processing script to template your pre and post scripts on your own, without direct support in constructor.  Given the amount of edge cases that can arise, it's tricky to come up with a generalized approach that works in all cases.

While I agree that would be useful for even more complex use cases, I think the idea of passing in additional environment variables into the scripts improves on a fairly simple and clearly defined interface, especially as constructor is already defining environment variables for consumption by these scripts. I think your suggestion is a good one but orthogonal to what is proposed in this PR.

examples/scripts/construct.yaml
```diff
@@ -5,6 +5,11 @@ channels:
   - http://repo.anaconda.com/pkgs/main/
 specs:
   - python
+
+extra_env_variables:
+  - CUSTOM_VARIABLE_1=CUSTOM1
+  - CUSTOM_VARIABLE_2=CUSTOM2
```
jaimergp, line 11: What about using the YAML dict syntax for this?

```suggestion
  - CUSTOM_VARIABLE_1: CUSTOM1
  - CUSTOM_VARIABLE_2: CUSTOM2
```

constructor/winexe.py
```diff
@@ -86,6 +86,23 @@ def insert_tempfiles_commands(paths: os.PathLike) -> List[str]:
...
+        info: Dictionary of information parsed from construct.yaml
+
+    Returns:
+        List[str]: Commands to be inserted into nsi template
+    """
+    lines = []
+    for variable in info.get('extra_env_variables',[]):
+        split = variable.split('=')
+        if len(split)==2:
+            (name, val) = split
```
jaimergp, line 102:
```suggestion
    for name, value in info.get('extra_env_variables', {}).items():
```

constructor/winexe.py
```diff
@@ -86,6 +86,23 @@ def insert_tempfiles_commands(paths: os.PathLike) -> List[str]:
...
+        lines.append(f"""System::Call 'kernel32::SetEnvironmentVariable(t,t)i("{name}", "{val}").r0'""")
```
jaimergp, line 103:
```suggestion
        lines.append(f"""System::Call 'kernel32::SetEnvironmentVariable(t,t)i("{name}", {str_esc(value)}).r0'""")
```

We need to be careful with values that have special characters. `str_esc` helps a bit, and also adds the quotes, so this is a bit safer.

constructor/shar.py
```diff
@@ -92,8 +92,11 @@ def get_header(conda_exec, tarball, info):
 
     data = read_header_template()
     data = preprocess(data, ppd)
+    custom_variables = info.get('extra_env_variables', [])
```
jaimergp, line 95:
```suggestion
    custom_variables = info.get('extra_env_variables', {})
```

constructor/shar.py
```diff
@@ -92,8 +92,11 @@ def get_header(conda_exec, tarball, info):
...
     data = fill_template(data, replace)
 
+    data = data.replace("_EXTRA_ENV_VARIABLES_=''", '\n'.join(
+        [f'export {var}' for var in custom_variables]))
```
jaimergp, line 99:
```suggestion
        [f'export {key}={value}' for key, value in custom_variables.items()]))
```

We need to decide whether we handle quote escaping or delegate that to the user with proper docs and examples.

constructor/osxpkg.py
```diff
@@ -308,8 +308,12 @@ def move_script(src, dst, info, ensure_shebang=False, user_script_type=None):
         'CONSTRUCTOR_VERSION': info['CONSTRUCTOR_VERSION'],
     }
     data = preprocess(data, ppd)
+    custom_variables = info.get('extra_env_variables', [])
```
jaimergp, line 311:
```suggestion
    custom_variables = info.get('extra_env_variables', {})
```

examples/scripts/construct.yaml
jlstevens: Sure! That helps validate thing at the yaml level so I am all for it.

constructor/winexe.py
jlstevens: I was planning to think about validation for the case of a missing `=` but that is moot now we are using yaml dicts instead. Happy to apply this suggestion.

constructor/shar.py
jlstevens: My own inclination is to delegate this to the user by documenting it properly.

jlstevens:
I went ahead and applied all your suggestions.

If this seems good to you, I'll go ahead and start tackling the docs.

jlstevens:
> Let's go with script_env_variables. I think extra is implied.

Agreed. Implemented in [7d43220](https://github.com/conda/constructor/pull/718/commits/7d4322003c901d1ff80ba28bbf7d995877f8ef6b)

> This makes me prefer single quoting, but that means no string interpolation. I think this is fine as long as it's documented. 

I agree. Single quoting seems safer and I think will be appropriate for the vast majority of use-cases. Once you get your values into your pre-/post-install scripts you can do your string interpolation in there.

On Unix I guess this just means wrapping the value from the dictionary in `'` which I can do easily enough. Unfortunately, I don't know anything about quoting on Windows (at least not off-hand) so I would have look into it. Do you know what kind of quoting would work best for Windows?

jlstevens:
I've updated the `scripts` example with a more challenging string. You can use whatever characters you want now except for single quote themselves which you can escape using `'\''` - I have documented this in CONSTRUCT.md.

Now I will test this out on Windows with the `.bat` files.

jaimergp:
> I have documented this in CONSTRUCT.md

In case you are doing this manually, the workflow is to edit `construct.py` and then run `scripts/make_docs.py` to sync all the generated copies.

jlstevens:
Getting special characters to work in the Windows environment variable seems to be trickier:

* Spaces are handled fine.
* `str_esc` handles `$` by turning it into `$$`
* Surprisingly `#` seems to be treated as a yaml comment and gets dropped (at least in my windows environments)
* Any quotes (single or double) truncate the string

So until I figure it out, my Windows test string is the same as the Unix one but without `#!'"` which I haven't manage to handle.

I am not sure how much effort should be spent on allowing these characters: I would be tempted to simply document that they are not supported (or I could apply some validation in the code...except for `#` which vanishes along with everything after it!)

dbast:
@conda-bot check

jlstevens:
@jaimergp All checks are now green!

Ready for review.

dbast (approved):

jlstevens:
Thanks for reviewing @dbast!

Issue #679, opened 2023-05-30
(RFC) How to update existing installations


jaimergp:
The update logic in the installers is a bit convoluted right now, and I don't think it's working at all.

The background is that previous iterations of constructor would consider the target environment (if not empty) as part of the fresh installation. However, at some point, the installer was instructed to error if the target directory was not empty (by default). Adding the `-f` or `-u` switches in the shell installer would somehow allow that, but there were no guarantees it would succeed. I am not aware of similar switches on Windows installers.

These days, the installer does not solve the target environment, it simply runs an explicit install, overwriting as needed if `-f` was passed.

I think that's still the right approach (happy to discuss that, though), but some users are confused by [how to update the installation](https://github.com/conda-forge/miniforge/issues/459#issuecomment-1568111984). There are two options:

- In installers providing a `conda` deployment, `conda update -n base conda` should suffice assuming the `base` environment is tractable.
- If `conda` is not involved, then it's a matter of uninstalling / removing and installing the new one.

I am thinking that maybe we could provide either a custom field to change the default message, or at least parameterize the message depending on whether `conda` is included in `base` or not. However, some apps might not want to mention `conda` at all to keep their delivery story simpler for their users (e.g. Spyder or napari).

marcoesters:
> The update logic in the installers is a bit convoluted right now, and I don't think it's working at all.

I would second that. There are a lot of intricacies involved. It looks like `-u` and `-f` are actually identical flags. So, updating really just installs whatever is in the installer into the `conda` environment.

That can, of course, cause all kinds of conflicts with existing packages, even in the best case scenario where the `base` environment has not been changed at all.

jaimergp:
Yes, to add some more context to other contributors:

* `constructor` runs an explicit install on the target location. This means that the solver does _not_ run at all. It's mostly extracting packages to the location. This is awesome when the location is empty, but it's a source of problems when we are "updating".
* What to expect out of an update? This depends on the lifetime of the application.
  *  If the application is not expected to change during its lifetime, an update might just be a standardized way of running the uninstallation before running the new installation on the same location. In other words, the "update" code would just delete everything in the target location, and then run the installation on top, usually keeping configuration files.
  * If the application does change during its lifetime, then it's a bit more complex. In installations like `Miniconda`, what are we supposed to do? Run `conda update conda`? If we blindly extract things there, it will most likely result in a broken installation, as Marco said above.

[...]

PR #736, opened 2023-12-15
Update installation for windows .exe

huche6:
### Description

- No possibility to update an installation with windows .exe installer. It would be nice to have something which would look like the -u flag available for linux .sh installer #679. 
- I changed the MessageBox which would poped when a path already exists to a YESNO MessageBox. The message is now "The directory '$INSTDIR' is not empty, do you want to update the installation ?". 
  - If the user choose "yes", the installer update the $INSTDIR
  - If the user choose "no", them are send back to select the installation path.
- I added a function in main.nsi.tmpl which is called CheckInstdirValidity. It checks the validity of the selected directory during the installation. It avoids to install something in a windows critical directory or a non conda directory. 
- This function was made just by externalising some code from the function un.onInit. I had in mind to do so to be able to call it when necessary. However, for some reasons, I can not achieve to call CheckInstdirValidity within un.onInit. If anyone has some idea about how to do it, it would avoid code duplication.  

### Checklist - did you ...

### DIFF
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -518,7 +518,7 @@ Function CheckInstdirValidity
         IfFileExists $INSTDIR\$2 0 invalid_dir
         IntOp $1 $1 + 1
         goto loop_conda
-    endloop_conda: 
+    endloop_conda:
 
     # All checks have passed
     goto valid_dir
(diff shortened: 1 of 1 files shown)


marcoesters:
> However, for some reasons, I can not achieve to call CheckInstdirValidity within un.onInit.

NSIS requires that all functions called during uninstallation start with `un.`. The way to implement this is to create a macro like it's done here: https://github.com/conda/constructor/blob/main/constructor/nsis/main.nsi.tmpl#L1020-L1059

huche6:
Thank you for your reply. I will have a look at it !

jaimergp:
Hi @huche6, before we proceed any further, we should discuss what to expect out an installer-driven update. Best done at #679. I'll mark this as draft til we have agreed on a plan forward.

Issue #731, opened 2023-11-17
pkg installers skip "Destination Select" page


marcoesters:
`pkg` installers created by `constructor` skip the "Destination Select" page, which causes the installer to try and install into `/`, which MacOS luckily prohibits. The resulting error message is misleading: it says that the package is not compatible with the MacOS version.

This is a known problem with pkg installers. See, e.g., these stackoverflow posts:

* https://stackoverflow.com/questions/4647416/mac-packagemaker-destination-select-step-a-skipping-boption-permanently-disa
* https://stackoverflow.com/questions/11487596/making-macos-installer-packages-which-are-developer-id-ready/11487658#11487658

The problem is that this isn't reproducible. So far, this issue appeared on MacOS 12 and 13 (osx-arm64), but not all. It's not even consistent with the same version. I have even seen that the same installer works at first and then stopped working on the same machine (I have never seen it suddenly start working though).

We cannot solve the bug in `pkgutil`, but maybe we can find a workaround.

PR #745, opened 2024-01-08
Allow customization of the domains attribute in pkg installers

marcoesters:
### Description

Allow customization of the domains attribute in the [Distribution](https://developer.apple.com/library/archive/documentation/DeveloperTools/Reference/DistributionDefinitionRef/Chapters/Distribution_XML_Ref.html) file in pkg installers.

This allows tailoring the "Destination Select" page, e.g., to only allow installing into specific locations (and thus circumvent the bug in #731).

### Checklist - did you ...


constructor/osxpkg.py
```diff
@@ -276,16 +276,17 @@ def modify_xml(xml_path, info):
     # - Install on a specific disk: /<custom_root>/<name>
     # On modern Mac systems, installing in root is not allowed. So
     # we remove this option by not supplying enable_localSystem
-    # below. Alternatively, we could have chosen not to set the
+    # in the default value for pkg_domains. enable_localSystem
+    # should only be set to true if default_location_pkg is supplied
```
jaimergp, line 280: Should we add a check in `main.py` testing for this?

news/745-customize-pkg-domains
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Allow customization of the domains attribute in pkg installers (#744)
```
jaimergp, line 3:
```suggestion
* Add `pkg_domains` setting to allow customization of the `domains` attribute in `PKG` installers (#731 via #744)
```

constructor/osxpkg.py
marcoesters: I added a warning

jaimergp: Thanks! Since the incompatibility is already documented, it should be enough (maybe someone reaaally wants to do the wrong thing 🤷).

jaimergp (approved):

tests/test_examples.py
```diff
@@ -524,6 +525,32 @@ def test_register_envs(tmp_path, request):
         assert str(install_dir) not in environments_txt
 
 
+@pytest.mark.skipif(sys.platform != 'darwin', reason='MacOS only')
+@pytest.mark.parametrize('domains', ({}, {'enable_anywhere': 'false', 'enable_localSystem': True}))
+def test_pkg_distribution_domains(tmp_path, domains):
```
jaimergp, line 530: Do you think we should still try to install the PKG in this test to make sure `pkg_domains` work?

marcoesters: The domains only affect the GUI installer, not the CLI. You can select `CurrentUserHomeDirectory` and `LocalSystem` as the target, no matter what `Distributions` says.

PR #758, opened 2024-02-27
Add cli option to supply path to the construct yaml file (supersedes #728)

### LINKED ISSUE #727: A cli option to specify the path to the construct yaml
### Checklist

### What is the idea?

By default constructor requires a construct.yaml to be in the directory where it is running.  We have had scenarios where we had multiple versions of construct.yaml and wanted to test between the two.  We have always had to rename or create symlinks to the file that we wanted to be processed.  This option would allow the files to exist separately and allow us to build the installer using the options in each of the files.

### Why is this needed?

_No response_

### What should happen?

_No response_

### Additional Context

_No response_

jaimergp:
### Description

Supersedes #728 and closes #727

### Checklist - did you ...


news/758-yaml-filename
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Add a new `--construct-yaml-fn` argument to specify an input file not named `construct.yaml`. (#727 via #758)
```
jezdez, line 3:
```suggestion
* Add a new `-c`/`--config-file` argument to specify an input file not named `construct.yaml`. (#727 via #758)
```

tests/test_examples.py
```diff
@@ -265,6 +266,8 @@ def create_installer(
         str(input_dir),
         "--output-dir",
         str(output_dir),
+        "--construct-yaml-fn",
```
jezdez, line 269:
```suggestion
        "--config-file",
```

tests/test_examples.py
```diff
@@ -383,7 +386,9 @@ def test_example_miniforge(tmp_path, request):
 
 def test_example_noconda(tmp_path, request):
     input_path = _example_path("noconda")
-    for installer, install_dir in create_installer(input_path, tmp_path, with_spaces=True):
+    for installer, install_dir in create_installer(
+        input_path, tmp_path, construct_yaml_filename="constructor_input.yaml", with_spaces=True
```
jezdez, line 390:
```suggestion
        input_path, tmp_path, config_file="constructor_input.yaml", with_spaces=True
```

constructor/main.py
```diff
@@ -68,7 +68,8 @@ def get_output_filename(info):
 
 def main_build(dir_path, output_dir='.', platform=cc_platform,
                verbose=True, cache_dir=DEFAULT_CACHE_DIR,
-               dry_run=False, conda_exe="conda.exe"):
+               dry_run=False, conda_exe="conda.exe",
+               construct_yaml_filename="construct.yaml"):
```
jezdez, line 72:
```suggestion
               config_file="construct.yaml"):
```

constructor/main.py
```diff
@@ -78,7 +79,7 @@ def main_build(dir_path, output_dir='.', platform=cc_platform,
     except ValueError:
         sys.exit("Error: invalid platform string '%s'" % platform)
 
-    construct_path = join(dir_path, 'construct.yaml')
+    construct_path = join(dir_path, construct_yaml_filename)
```
jezdez, line 82:
```suggestion
    construct_path = join(dir_path, config_file)
```

But! If the input can be relative or absolute paths, this should more carefully join with `os.path.relpath` or similar.

jezdez (changes requested):

constructor/main.py
jaimergp: This argument is meant to _only_ contain a filename, not a path. It should be combined with the main input directory.

constructor/main.py
```diff
@@ -357,6 +358,13 @@ def main():
                    action="store",
                    metavar="CONDA_EXE")
 
+    p.add_argument('--construct-yaml-fn',
```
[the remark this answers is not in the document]
jaimergp: Can we make it `config-filename` then?

constructor/main.py
jezdez: But you add a new option which is not validated and doesn't prevent from pass in a path, `--config-file=/some/other/path/construct.yaml` is possible (if I read the code correctly), so you'll have to account for that possibility by constructing the path more defensively. Previously that wasn't needed since the filename was hard-coded.

constructor/main.py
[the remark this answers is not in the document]
jezdez: Unless you're validating the data that it's just a name, I don't think that's useful IMO. It's easier to validate the value (e.g. if it's a relative or an absolute path, treat it as the default, and fallback to joining with the CWD)

constructor/main.py
jaimergp: Let's validate then 🚀

jaimergp:
@jezdez PTAL. Once this goes in, I think we can release a new constructor version.

constructor/main.py
```diff
@@ -381,6 +389,11 @@ def main():
     dir_path = args.dir_path
     if not isdir(dir_path):
         p.error("no such directory: %s" % dir_path)
+    if os.sep in args.config_filename:
+        p.error("--config-filename can only be a filename, not a path")
```
jezdez, line 393: 👍🏻

jezdez (approved):

PR #771, opened 2024-04-09
Support azure signtool

### LINKED ISSUE #767: Support AzureSignTool
### Checklist

### What is the idea?

[AzureSignTool](https://github.com/vcsjones/AzureSignTool/) is another tool that can be used to sign binaries. It uses a vault to obtain the certificate instead of using a certificate file.

`constructor` should extend its support to different signing tools and also support key-vault signature processes.

### Why is this needed?

Using vaults to sign binaries is going to be more and more important. 
`AzureSignTool` uses a vault without files, so signing installers is not possible with this

As of now, `constructor` uses `signtool.exe` and local files to sign binaries, so `AzureSignTool` or other file-less signature tools will not work.

### What should happen?

* Expanding `construct.yaml` to allow for vaults to sign packages.
* Allow for other tools to sign binaries, starting with `AzureSignTool`.

### Additional Context

_No response_

marcoesters:
### Description

Add support for AzureSignTool to sign Windows installers. `AzureSignTool` uses Azure key vaults, which do not use certificate files on disk. To achieve this, this PR:

* Adds a `windows_signing_tool` keyword that selects the tool to sign. This key determines whether to sign the installer instead of `signing_certificate`, which makes signing more extensible.
* Creates classes for Windows signing tools.
* Adds documentation for all environment variables for both `SignTool` and `AzureSignTool`.
* Allows `signtool.exe` to use a different digest algorithms.

Since this requires access to Azure, tests cannot be added for this feature.

Closes #767.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -95,7 +95,7 @@ jobs:
           environment-file: dev/environment.yml
           python-version: ${{ matrix.python-version }}
       - name: Install AzureSignTool
-        if: ${{ matrix.os == 'windows' }}
+        if: matrix.os == 'windows'
         run: dotnet.exe tool install --global AzureSignTool
         shell: pwsh
       - name: Supply extra dependencies and install constructor
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -94,6 +94,10 @@ jobs:
           activate-environment: constructor-dev
           environment-file: dev/environment.yml
           python-version: ${{ matrix.python-version }}
+      - name: Install AzureSignTool
+        if: ${{ matrix.os == 'windows' }}
+        run: dotnet.exe tool install --global AzureSignTool
+        shell: pwsh
       - name: Supply extra dependencies and install constructor
         run: |
           files=(--file "tests/requirements.txt")
@@ -129,6 +133,11 @@ jobs:
           flags: unit
       - name: Run examples
         env:
+          AZURE_SIGNTOOL_KEY_VAULT_CERTIFICATE: ${{ secrets.AZURE_SIGNTOOL_KEY_VAULT_CERTIFICATE }}
+          AZURE_SIGNTOOL_KEY_VAULT_CLIENT_ID: ${{ secrets.AZURE_SIGNTOOL_KEY_VAULT_CLIENT_ID }}
+          AZURE_SIGNTOOL_KEY_VAULT_SECRET: ${{ secrets.AZURE_SIGNTOOL_KEY_VAULT_SECRET }}
+          AZURE_SIGNTOOL_KEY_VAULT_TENANT_ID: ${{ secrets.AZURE_SIGNTOOL_KEY_VAULT_TENANT_ID }}
+          AZURE_SIGNTOOL_KEY_VAULT_URL: ${{ secrets.AZURE_SIGNTOOL_KEY_VAULT_URL }}
           CONSTRUCTOR_EXAMPLES_KEEP_ARTIFACTS: "${{ runner.temp }}/examples_artifacts"
           CONSTRUCTOR_SIGNTOOL_PATH: "C:/Program Files (x86)/Windows Kits/10/bin/10.0.17763.0/x86/signtool.exe"
         run: |
--- a/constructor/construct.py
+++ b/constructor/construct.py
@@ -16,6 +16,7 @@ from constructor.exceptions import UnableToParse, UnableToParseMissingJinja2, Ya
 from constructor.utils import yaml
 
 WIN_SIGNTOOLS = [
+    "azuresigntool",
     "signtool",
 ]
 
--- a/constructor/construct.py
+++ b/constructor/construct.py
@@ -251,17 +251,14 @@ to sign `conda.exe`. For this, you need an "Application certificate" (different 
 The tool used to sign Windows installers. Must be one of: {", ".join(WIN_SIGNTOOLS)}.
 Some tools require `signing_certificate` to be set.
 Defaults to `signtool` if `signing_certificate` is set.
+Additional environment variables may need to be used to configure signing.
+See the documentation for details:
+https://conda.github.io/constructor/howto/#signing-exe-installers
 '''),
 
     ('signing_certificate',  False, str, '''
 On Windows only, set this key to the path of the certificate file to be used
-with the `windows_signing_tool`. Additional environment variables may need to
-be used to configure signing, namely:
-
-- `CONSTRUCTOR_PFX_CERTIFICATE_PASSWORD` (password to unlock the certificate, if needed)
-- `CONSTRUCTOR_SIGNTOOL_PATH` (absolute path to `signtool.exe`, in case is not in `PATH`)
-- `CONSTRUCTOR_SIGNTOOL_TIMESTAMP_SERVER_URL` (custom RFC 3161 timestamping server, default is
-http://timestamp.sectigo.com)
+with the `windows_signing_tool`.
 '''),
 
     ('attempt_hardlinks',          False, (bool, str), '''
--- a/constructor/construct.py
+++ b/constructor/construct.py
@@ -805,8 +805,9 @@ def verify(info):
                          f"must be an instance of {types_str}")
     if signtool := info.get("windows_signing_tool"):
         if signtool not in WIN_SIGNTOOLS:
-            sys.exit("Value for 'windows_signing_tool' must be one of: "
-                f"{', '.join(WIN_SIGNTOOL)}. You tried to use: {signtool}."
+            sys.exit(
+                "Value for 'windows_signing_tool' must be one of: "
+                f"{', '.join(WIN_SIGNTOOLS)}. You tried to use: {signtool}."
             )
         need_cert_file = ["signtool"]
         if signtool in need_cert_file and not info.get("signing_certificate"):
--- a/constructor/construct.py
+++ b/constructor/construct.py
@@ -15,6 +15,10 @@ from ruamel.yaml import YAMLError
 from constructor.exceptions import UnableToParse, UnableToParseMissingJinja2, YamlParsingError
 from constructor.utils import yaml
 
+WIN_SIGNTOOLS = [
+    "signtool",
+]
+
 # list of tuples (key name, required, type, description)
 KEYS = [
     ('name',                   True,  str, '''
@@ -242,9 +246,16 @@ to sign `conda.exe`. For this, you need an "Application certificate" (different 
 `Developer ID Application: Name of the owner (XXXXXX)`.
 '''),
 
+    ('windows_signing_tool', False, str, f'''
+The tool used to sign Windows installers. Must be one of: {", ".join(WIN_SIGNTOOLS)}.
+Some tools require `signing_certificate` to be set.
+Defaults to `signtool` if `signing_certificate` is set.
+'''),
+
     ('signing_certificate',  False, str, '''
-On Windows only, set this key to the path of a PFX certificate to be used with `signtool`.
-Additional environment variables can be used to configure this step, namely:
+On Windows only, set this key to the path of the certificate file to be used
+with the `windows_signing_tool`. Additional environment variables may need to
+be used to configure signing, namely:
 
 - `CONSTRUCTOR_PFX_CERTIFICATE_PASSWORD` (password to unlock the certificate, if needed)
 - `CONSTRUCTOR_SIGNTOOL_PATH` (absolute path to `signtool.exe`, in case is not in `PATH`)
@@ -792,6 +803,14 @@ def verify(info):
                 types_str = " or ".join([type_.__name__ for type_ in types])
                 sys.exit(f"Value for 'extra_envs.{env_name}.{key}' "
                          f"must be an instance of {types_str}")
+    if signtool := info.get("windows_signing_tool"):
+        if signtool not in WIN_SIGNTOOLS:
+            sys.exit("Value for 'windows_signing_tool' must be one of: "
+                f"{', '.join(WIN_SIGNTOOL)}. You tried to use: {signtool}."
+            )
+        need_cert_file = ["signtool"]
+        if signtool in need_cert_file and not info.get("signing_certificate"):
+            sys.exit(f"The signing tool {signtool} requires 'signing_certificate' to be set.")
 
 
 def generate_doc():
--- a/constructor/construct.py
+++ b/constructor/construct.py
@@ -802,7 +802,7 @@ def verify(info):
                 sys.exit(f"Value for 'extra_envs.{env_name}.{key}' "
                          f"must be an instance of {types_str}")
     if signtool := info.get("windows_signing_tool"):
-        if signtool not in WIN_SIGNTOOLS:
+        if signtool.lower().replace(".exe", "") not in WIN_SIGNTOOLS:
             sys.exit(
                 "Value for 'windows_signing_tool' must be one of: "
                 f"{', '.join(WIN_SIGNTOOLS)}. You tried to use: {signtool}."
--- a/constructor/main.py
+++ b/constructor/main.py
@@ -109,7 +109,10 @@ def main_build(dir_path, output_dir='.', platform=cc_platform,
         if info.get(key):  # only join if there's a truthy value set
             info[key] = abspath(join(dir_path, info[key]))
 
-    if info.get("signing_certificate") and not info.get("windows_signing_tool"):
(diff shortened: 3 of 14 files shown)


docs/source/howto.md
```diff
@@ -38,21 +38,63 @@ On Windows, you can also add extra pages to the installer. This is an advanced o
...
+| `AZURE_SIGNTOOL_KEY_VAULT_URL`              | The URL of the key vault with the certificate.                                              | `-kvu`   | Empty                        |
+| `CONSTRUCTOR_SIGNTOOL_PATH`                 | Path to `AzureSignTool.exe`. Needed if `azuresigntool` is not in `PATH`.                    | N/A      | `azuresigntool`              |
+| `CONSTRUCTOR_SIGNTOOL_FILE_DIGEST`          | Digest algorithm for creating the file signatures.                                          | `-fd`    | `sha256`                     |
+| `CONSTRUCTOR_SIGNTOOL_TIMESTAMP_SERVER_URL` | URL to the RFC 3161 timestamp server.                                                       | `-tr`    | http://timestamp.sectigo.com |
+| `CONSTRUCTOR_SIGNTOOL_TIMESTAMP_DIGEST`     | Digest algorithm for the RFC 3161 time stamp.                                               | `-td`    | `sha256`                     |
+
+:::{note}
+
+If neither `AZURE_SIGNTOOL_KEY_VAULT_ACCESSTOKEN` nor `AZURE_SIGNTOOL_KEY_VAULT_SECRET` are set, `constructor` will use a Managed Identity (`-kvm` CLI option).
+
```
jaimergp, line 89:
```suggestion
:::
```

docs/source/howto.md
```diff
@@ -38,21 +38,63 @@ On Windows, you can also add extra pages to the installer. This is an advanced o
...
+### Signing and notarizing pkg installers
```
jaimergp, line 90:
```suggestion
### Signing and notarizing PKG installers
```

news/771-add-azure-signtool-support
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Add support for AzureSignTool to sign Windows installers.
```
jaimergp, line 3:
```suggestion
* Add support for AzureSignTool to sign Windows installers. (#767 via #771)
```

constructor/construct.py
```diff
@@ -792,6 +801,15 @@ def verify(info):
                 types_str = " or ".join([type_.__name__ for type_ in types])
                 sys.exit(f"Value for 'extra_envs.{env_name}.{key}' "
                          f"must be an instance of {types_str}")
+    if signtool := info.get("windows_signing_tool"):
+        if signtool not in WIN_SIGNTOOLS:
+            sys.exit(
```
jaimergp, line 806: This reminds me we need to start raising exceptions instead of sys.exits, but that's for another PR.

constructor/main.py
```diff
@@ -109,6 +109,12 @@ def main_build(dir_path, output_dir='.', platform=cc_platform,
         if info.get(key):  # only join if there's a truthy value set
             info[key] = abspath(join(dir_path, info[key]))
 
+    # Normalize name and set default value
+    if info.get("windows_signing_tool"):
+        info["windows_signing_tool"] = info["windows_signing_tool"].lower().replace(".exe", "")
```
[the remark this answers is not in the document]
jaimergp: At the same time, if we are validating the input in construct.py:805, why do we need normalization here? 🤔 For library usage? In that case it should be the same logic (or move the logic here).

constructor/main.py
```diff
@@ -109,6 +109,12 @@ def main_build(dir_path, output_dir='.', platform=cc_platform,
...
+    elif info.get("signing_certificate"):
+        info["windows_signing_tool"] = "signtool"
```
jaimergp, line 116: We should check if we are overriding something here, and warn appropriately.

constructor/signing.py
```diff
@@ -0,0 +1,222 @@
...
+            f"{win_str_esc(self.executable)} sign /f {win_str_esc(self.certificate_file)} "
+            f"/tr {win_str_esc(timestamp_server)} /td {timestamp_digest} /fd {file_digest}"
+        )
+        if "CONSTRUCTOR_PFX_CERTIFICATE_PASSWORD" in os.environ:
+            # signtool can get the password from the env var on its own
+            command += ' /p "%CONSTRUCTOR_PFX_CERTIFICATE_PASSWORD%"'
+        return command
+
+    def verify_signing_tool(self):
+        self._verify_tool_is_available()
```
jaimergp, line 93:
```suggestion
        super()._verify_tool_is_available()
```

constructor/signing.py
```diff
@@ -0,0 +1,222 @@
...
+    def __init__(self):
+        super().__init__(os.environ.get("CONSTRUCTOR_SIGNTOOL_PATH", "AzureSignTool"))
+
+    def get_signing_command(self) -> str:
+        def _check_required_env_vars(env_vars):
+            missing_vars = {var for var in env_vars if var not in os.environ}
+            if missing_vars:
+                raise RuntimeError(
+                    f"Missing required environment variables {', '.join(missing_vars)}."
+                )
```
jaimergp, line 128: I think this can be easily moved to the base class as its own (static?) method.

constructor/signing.py
```diff
@@ -0,0 +1,222 @@
...
+            ' -kvu %AZURE_SIGNTOOL_KEY_VAULT_URL%'
+            ' -kvc %AZURE_SIGNTOOL_KEY_VAULT_CERTIFICATE%'
+            f" -tr {timestamp_server}"
+            f" -td {timestamp_digest}"
+            f" -fd {file_digest}"
+        )
+        # There are three ways to sign:
+        #   1. Access token
+        #   2. Secret (requires tenant ID)
+        #   3. Managed identity (requires prior login to Azure)
```
jaimergp, line 159: Can we `log.info` which method was detected?

constructor/signing.py
```diff
@@ -0,0 +1,222 @@
...
+            command += " -kvm"
+        return command
+
+    def verify_signing_tool(self):
+        self._verify_tool_is_available()
+        check_call([self.executable, "--help"], stdout=PIPE, stderr=PIPE)
+
+    def verify_signature(self, installer_file: Union[str, Path]):
+        """
+   `    https://learn.microsoft.com/en-us/dotnet/api/system.management.automation.signaturestatus
```
jaimergp, line 185:
```suggestion
        https://learn.microsoft.com/en-us/dotnet/api/system.management.automation.signaturestatus
```

constructor/signing.py
```diff
@@ -0,0 +1,222 @@
...
+                command,
+            ],
+            capture_output=True,
+            text=True,
+        )
+        # The return code will always be 0,
+        # but stderr will be non-empty on errors
+        if proc.stderr:
+            logger.error(proc.stderr)
+            return
```
jaimergp, line 207: Should we raise here? 🤔

constructor/winexe.py
```diff
@@ -217,27 +210,7 @@ def uninstall_menus_commands(info):
...
-        )
-        if "CONSTRUCTOR_PFX_CERTIFICATE_PASSWORD" in os.environ:
-            # signtool can get the password from the env var on its own
-            command += ' /p "%CONSTRUCTOR_PFX_CERTIFICATE_PASSWORD%"'
-        return command
-    return ""
-
-
-def make_nsi(info, dir_path, extra_files=None, temp_extra_files=None):
+def make_nsi(info, dir_path, extra_files=None, temp_extra_files=None, signing_tool=None):
```
jaimergp, line 213: Can we add some typing? There are a few things called `signtool` or similar now and it can get confusing. I think this expects an instance of the new class in this PR?

jaimergp (commented):
Very nice work. I'm a little concerned about the lack of tests, so I wonder if it's feasible to have a some secrets added to a dummy vault (or something that implements the same API). However, that might be too complicated and this feature will be used routinely by Anaconda anyway, so I guess we'll know when it breaks.

Double check that the docs render correctly locally, I found a little typo with the admonition syntax. Other than that, this is good to go after addressing the comments.

constructor/main.py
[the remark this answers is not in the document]
marcoesters: > What happens if a user tries to set this to a full path? 

`constructor` will and should fail. `windows_signing_tool` is the name. Otherwise, I would have to infer from the binary what the appropriate signing too is.

>  In that case it should be the same logic

Yes, that's true, good catch!

> (or move the logic here).

I'm not sure about this. Setting default values and manipulating input is done in `main_build`. `validate()` performs validation here. If we move this line into `validate()`, we are mixing scopes, which makes things harder to track.

constructor/main.py
marcoesters: That check is done in line 113 - if `windows_signing_tool` is set in any way, we won't get to `elif`.

constructor/signing.py
marcoesters: Since this is a very general helper function, I moved it to utils

marcoesters:
I added integration tests and secrets were added to the repository. However, the secrets will not be available for anybody who submits a PR from the fork, which will cause the tests to fail. The tests do execute the correct command though: https://github.com/conda/constructor/actions/runs/8760110611/job/24044501617#step:9:2133

Those tests will have to be skipped until merged into `main`. So, we will have to add 
As discussed with @jaimergp offline, the best way to go here is to add test failure reporting as they do for `conda-libmambda-solver`: https://github.com/conda/conda-libmamba-solver/blob/efa3b84141e4ff777928aaae952fac417b059091/.github/workflows/tests.yml#L680-L690

This requires a token, so I will convert this PR into draft until that infrastructure is in place.

Additional things I found:

* The environment is not copied into the subprocess on Windows when no extra environment variables are set. On POSIX, this is handled by the subprocess module, but not for Windows. So, I copy the environment explicitly for all to get conistent behavior.
* The debugging section was not in any TOC tree for the documentations, so I added it.

.github/workflows/main.yml
```diff
@@ -94,6 +94,10 @@ jobs:
           activate-environment: constructor-dev
           environment-file: dev/environment.yml
           python-version: ${{ matrix.python-version }}
+      - name: Install AzureSignTool
+        if: ${{ matrix.os == 'windows' }}
```
jaimergp, line 98:
```suggestion
        if: matrix.os == 'windows'
```
These can be omitted in `if` clauses, I think.

constructor/signing.py
```diff
@@ -0,0 +1,220 @@
...
+            "CONSTRUCTOR_SIGNTOOL_TIMESTAMP_DIGEST",
+            "sha256"
+        )
+        file_digest = os.environ.get(
+            "CONSTRUCTOR_SIGNTOOL_FILE_DIGEST",
+            "sha256"
+        )
+        command = (
+            f"{win_str_esc(self.executable)} sign /f {win_str_esc(self.certificate_file)} "
+            f"/tr {win_str_esc(timestamp_server)} /td {timestamp_digest} /fd {file_digest}"
```
jaimergp, line 85: Just noting here that no quotes are needed because we don't expect spaces here.

constructor/signing.py
```diff
@@ -0,0 +1,220 @@
...
+        )
+        file_digest = os.environ.get(
+            "AZURE_SIGNTOOL_FILE_DIGEST",
+            "sha256"
+        )
+
+        command = (
+            f"{self.executable} sign -v"
+            ' -kvu %AZURE_SIGNTOOL_KEY_VAULT_URL%'
+            ' -kvc %AZURE_SIGNTOOL_KEY_VAULT_CERTIFICATE%'
```
jaimergp, line 145: Should we quote this?

constructor/signing.py
```diff
@@ -0,0 +1,220 @@
...
+            logger.info("AzureSignTool: signing binary using secret.")
+            required_env_vars = (
+                "AZURE_SIGNTOOL_KEY_VAULT_CLIENT_ID",
+                "AZURE_SIGNTOOL_KEY_VAULT_TENANT_ID",
+            )
+            check_required_env_vars(required_env_vars)
+            command += (
+                " -kvi %AZURE_SIGNTOOL_KEY_VAULT_CLIENT_ID%"
+                " -kvt %AZURE_SIGNTOOL_KEY_VAULT_TENANT_ID%"
+                " -kvs %AZURE_SIGNTOOL_KEY_VAULT_SECRET%"
```
jaimergp, line 168: Same, do we need to quote any of these?

constructor/signing.py
```diff
@@ -0,0 +1,220 @@
...
+            capture_output=True,
+            text=True,
+        )
+        # The return code will always be 0,
+        # but stderr will be non-empty on errors
+        if proc.stderr:
+            raise RuntimeError(f"Signature verification failed.\n{proc.stderr}")
+        try:
+            status, status_message = proc.stdout.strip().split("\n")
+            status = int(status)
```
jaimergp, line 208: Can this fail if status is empty or None?

tests/test_examples.py
```diff
@@ -484,6 +483,43 @@ def test_example_signing(tmp_path, request):
...
 
 
+@pytest.mark.skipif(sys.platform != "win32", reason="Windows only")
+@pytest.mark.skipif(
+        not shutil.which("azuresigntool") and not os.environ.get("AZURE_SIGNTOOL_PATH"),
+        reason="AzureSignTool not available"
+)
+@pytest.mark.parametrize(
+        "auth_method",
+        os.environ.get("AZURE_SIGNTOOL_TEST_AUTH_METHODS", "token,secret").split(","),
```
jaimergp, line 493: Why do we need this "parametrized" parametrization? Are you using it locally while debugging? I don't see it in the CI setup 🤔

constructor/signing.py
marcoesters: I'm not sure how this could be `None` unless `proc.stdout` can somehow be `None`.

If `status` is and empty string, `int(status)` raises a `ValueError`, which is caught further down.

tests/test_examples.py
marcoesters: I wanted to make sure that all authentication methods could potentially be covered. For tokens and secrets, I can just check the environment variables and skip if needed. For managed identities, however, this does not work because there isn't an environment variable to check for (logging into the vault must happen before AzureSignTool is executed).

So, the only "safe" methods to add into the parametrization are `token` and `secret`. However, I still wanted to allow testing with a managed identity.

Does that make sense? I tried to explain this briefly in the docstring.

constructor/signing.py
marcoesters: We can. I just tested it locally and the test worked fine with quotes, too.

tests/test_examples.py
```diff
@@ -484,6 +483,43 @@ def test_example_signing(tmp_path, request):
...
+)
+def test_azure_signtool(tmp_path, request, monkeypatch, auth_method):
+    """Test signing installers with AzureSignTool.
+
+    There are three ways to authenticate with Azure: tokens, secrets, and managed identities.
+    There is no good sentinel environment for manged identities, so an environment variable
```
jaimergp, line 499:
```suggestion
    There is no good sentinel environment for managed identities, so an environment variable
```

I think?

docs/source/howto.md
```diff
@@ -38,21 +38,63 @@ On Windows, you can also add extra pages to the installer. This is an advanced o
...
+| `AZURE_SIGNTOOL_KEY_VAULT_URL`         | The URL of the key vault with the certificate.                                              | `-kvu`   | Empty                        |
+| `AZURE_SIGNTOOL_PATH`                  | Path to `AzureSignTool.exe`. Needed if `azuresigntool` is not in `PATH`.                    | N/A      | `azuresigntool`              |
+| `AZURE_SIGNTOOL_TIMESTAMP_SERVER_URL`  | URL to the RFC 3161 timestamp server.                                                       | `-tr`    | http://timestamp.sectigo.com |
+| `AZURE_SIGNTOOL_TIMESTAMP_DIGEST`      | Digest algorithm for the RFC 3161 time stamp.                                               | `-td`    | `sha256`                     |
+
+:::{note}
+
+If neither `AZURE_SIGNTOOL_KEY_VAULT_ACCESSTOKEN` nor `AZURE_SIGNTOOL_KEY_VAULT_SECRET` are set, `constructor` will use a Managed Identity (`-kvm` CLI option).
+:::
+### Signing and notarizing PKG installers
```
jaimergp, line 90:
```suggestion

### Signing and notarizing PKG installers
```

jaimergp (approved):

PR #776, opened 2024-04-22
Add step to file issues if tests fails after merging to main

### LINKED ISSUE #775: Automatically create issue on test failure after merging to `main`
### Checklist

### What is the idea?

Add automatic issue creation when tests fail after pushing to `main`, similar to what is being done for `conda-libmamba-solver`: https://github.com/conda/conda-libmamba-solver/blob/efa3b84141e4ff777928aaae952fac417b059091/.github/workflows/tests.yml#L680-L690

### Why is this needed?

Integration tests for key-vault certificate systems such as Azure require repository secrets (see #771). These secrets are not available for anybody using a fork. This will cause integration tests to be fail, so they have to be skipped.

This means that test failures will not be visible until after merging into `main`. This can cause errors to go unnoticed. Automatically creating an issue on test failures will provide visibility for when requests fail.

### What should happen?

An issue should be created when tests fail after merging to `main`.

### Additional Context

_No response_

marcoesters:
### Description

Adding integration tests for AzureSignTool (#771) requires access to repository secrets. These are not available when submitting a PR from a fork. Tests must then be executed after merging to `main`. This risks that errors are not caught.

To increase visibility of these problems, create an issue when tests fail after being merged into `main`.

Closes #775.

Token has been added to the repo: https://github.com/conda/infrastructure/issues/920

### Checklist - did you ...

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -145,6 +145,16 @@ jobs:
         run: |
           python scripts/make_docs.py
           git diff --exit-code
+      - name: Report failures
+        if: always() && github.event_name == 'push'
+        uses: JasonEtco/create-an-issue@v2
+        env:
+          GITHUB_TOKEN: ${{ secrets.CONSTRUCTOR_ISSUES }}
+          RUN_ID: ${{ github.run_id }}
+          TITLE: "Constructor tests failed"
+        with:
+          filename: .github/TEST_FAILURE_REPORT_TEMPLATE.md
+          update_existing: true
       - name: Upload the example installers as artifacts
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -145,6 +145,16 @@ jobs:
         run: |
           python scripts/make_docs.py
           git diff --exit-code
+      - name: Report failures
+        if: always() && github.event_name == 'push'
```
jaimergp, line 149: This should only run on `main`, in case we have `push` events to a branch for a PR. But we are only allowing this event on `main` and tags, so we are fine.

jaimergp (approved):

PR #782, opened 2024-04-23
Use separate job to create issues on test failure

marcoesters:
### Description

Submitting issues on test failures in the test job can result in duplicate issues since the tests run in parallel. This introduces a race condition.

Instead, use a separate job that gets called when all tests have concluded.

Also, add a pre-commit hook to check GitHub workflows, which has been missing so far.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -145,8 +145,21 @@ jobs:
         run: |
           python scripts/make_docs.py
           git diff --exit-code
+      - name: Upload the example installers as artifacts
+        if: github.event_name == 'pull_request' && matrix.python-version == '3.9'
+        uses: actions/upload-artifact@v3
+        with:
+          name: installers-${{ runner.os }}-${{ github.sha }}-${{ github.run_id }}-${{ github.run_number }}-${{ github.run_attempt }}
+          path: "${{ runner.temp }}/examples_artifacts"
+          retention-days: 7
+
+  report:
+    name: Report failures
+    needs: [tests]
+    if: always() && github.event_name == 'push' && needs.tests.result == 'failure'
+    runs-on: ubuntu-latest
+    steps:
       - name: Report failures
-        if: always() && github.event_name == 'push'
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -145,8 +145,21 @@ jobs:
...
+        uses: actions/upload-artifact@v3
+        with:
+          name: installers-${{ runner.os }}-${{ github.sha }}-${{ github.run_id }}-${{ github.run_number }}-${{ github.run_attempt }}
+          path: "${{ runner.temp }}/examples_artifacts"
+          retention-days: 7
+
+  report:
+    name: Report failures
+    needs: [tests]
+    if: always() && github.event_name == 'push' && needs.tests.result == 'failure'
```
jaimergp, line 159: Should we also check if we are pushing to `main`, or is that guaranteed? Better to add it even if the `on` triggers are configured safely now, because we might miss it in a future updates and end up with issues for noisy signals.

jaimergp:
LGTM. Let's fix the conflicts and address the branch == main issue and that's all.

.github/workflows/main.yml
```diff
@@ -145,8 +145,21 @@ jobs:
...
+          path: "${{ runner.temp }}/examples_artifacts"
+          retention-days: 7
+
+  report:
+    name: Report failures
+    needs: [tests]
+    if: always() && github.event_name == 'push' && github.ref_name == 'main' && needs.tests.result == 'failure'
+    runs-on: ubuntu-latest
+    steps:
       - name: Report failures
```
jaimergp, line 162: I think we need to checkout the repo before this action so it can access the template MD.

jaimergp:
```suggestion
      - name: Retrieve the source code
        uses: actions/checkout@44c2b7a8a4ea60a981eaca3cf939b5f4305c123b # v4.1.5
      - name: Report failures
```

jaimergp (approved):

PR #784, opened 2024-05-01
pkg installer: Fix ownership of files installed into $HOME

marcoesters:
### Description

When installing outside `$HOME`, the pkg installer installs as `root`. Constructor changes the ownership of the `$PREFIX` directory to `$USER`: https://github.com/conda/constructor/blob/58385ee6d58a2228ddbf8b40f2fe179a3e77b5ab/constructor/osx/run_installation.sh#L115

However, the installer creates additional files and directories outside of `$PREFIX`:

* `${HOME}/.conda`
* All files and directories created by `conda init`, e.g., `${HOME}/.bash_profile` or `${HOME}/.config/powershell`

This PR fixes the ownership of those files and directories as well.

### Checklist - did you ...


constructor/osx/run_installation.sh
```diff
@@ -110,9 +110,10 @@ if ! "$PREFIX/bin/python" -V; then
     exit 1
 fi
 
-# This is unneeded for the default install to ~, but if the user changes the
+# This is not needed for the default install to ~, but if the user changes the
 # install location, the permissions will default to root unless this is done.
 chown -R "$USER" "$PREFIX"
+chown -R "$USER" "${HOME}/.conda"
```
jaimergp, line 116: I think it might also leave a `.condarc` somewhere if the config file asked for that option. Should we add it here too? e.g.

```suggestion
test -f "${HOME}/.condarc" && chown "$USER" "${HOME}/.condarc"
```

constructor/osx/update_path.sh
```diff
@@ -8,4 +8,35 @@ set -eux
...
+            if [[ -f "${file}" ]] || [[ -d "${file}" ]]; then
+                OWNER=$(stat -f "%u" "${file}" | id -un)
+                if [[ "${OWNER}" == "root" ]]; then
+                    chown "${USER}" "${file}"
+                fi
+            fi
+            file="${file%/*}"
+        done
+    done
+fi
```
jaimergp, line 42: Awesome shell skills here! This script is called with `sh` though, so I think we need to either:

- Change the shebang to bash
- Make sure we only use strict POSIX (e.g. no `[[ ]]` ifs).

news/784-pkg-installation-file-ownership
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fix ownership of files created by the pkg installer outside of $PREFIX. (#784)
```
jaimergp, line 7:
```suggestion
* Fix ownership of files created by the PKG installer outside of `$PREFIX`. (#784)
```

constructor/osx/update_path.sh
```diff
@@ -8,4 +8,35 @@ set -eux
...
+# files modified by conda init must be changed to belong to $USER.
+# For shells like fish or powershell, conda init creates subdirectories
+# inside $HOME and/or $PREFIX, so ensure that parent directories also
+# have the correct owner.
+if [[ "${USER}" != "root" ]]; then
+    echo "Fixing permissions..."
+    read -r -a MODIFIED_FILES <<< "$(\
+      echo "${INIT_FILES}" |\
+      grep modified |\
+      awk '{print $2}' |\
```
jaimergp, line 24:
```suggestion
      awk '/modified/{print $2}' |\
```

I think this is equivalent. You can probably add the `grep` call below here as well, in between the slashes.

jaimergp:
Let's double check SH vs BASH vs POSIX for that nice chunk of shell code :)

constructor/osx/update_path.sh
marcoesters: Doesn't `sh` link to `bash` on MacOS?

constructor/osx/run_installation.sh
marcoesters: When I tested it, `.condarc` had the correct ownership, but better safe than sorry.

constructor/osx/update_path.sh
marcoesters: Even if it does, I think it's better to be explicit. I changed the shell to bash since that's what we're using in the other scripts as well.

jaimergp (approved):

Issue #814, opened 2024-06-28
Output files are overwritten for `installer_type: all`


marcoesters:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

The output files have generic names like `info.json`. When creating both `pkg` and `sh` installers, `constructor` does not distinguish between the two when creating output files so that `info.json` is overwritten.

It is of course possible to just call them `info.sh.json` and `info.pkg.json`, but I propose a different naming scheme: `<installer name>_info.json`. While this will make file names longer, it makes file names unique and directly tied to the installer and simplifies the work of users: if multiple installers are built into one directory or uploaded to a GitHub release, the user will not have to do any sort of renaming to make sure files are not overwritten.

### Conda list

```shell
# packages in environment at /Users/mesters/opt/miniconda3/envs/constructor:
#
# Name                    Version                   Build  Channel
archspec                  0.2.3              pyhd3eb1b0_0    defaults
boltons                   23.0.0          py312hca03da5_0    defaults
brotli-python             1.0.9           py312h313beb8_8    defaults
bzip2                     1.0.8                h80987f9_6    defaults
c-ares                    1.19.1               h80987f9_0    defaults
ca-certificates           2024.3.11            hca03da5_0    defaults
certifi                   2024.6.2        py312hca03da5_0    defaults
cffi                      1.16.0          py312h80987f9_1    defaults
charset-normalizer        2.0.4              pyhd3eb1b0_0    defaults
conda                     24.5.0          py312hca03da5_0    defaults
conda-libmamba-solver     24.1.0             pyhd3eb1b0_0    defaults
conda-package-handling    2.3.0           py312hca03da5_0    defaults
conda-package-streaming   0.10.0          py312hca03da5_0    defaults
conda-standalone          24.5.0               hca03da5_0    defaults
constructor               3.8.0           py312hca03da5_0    defaults
distro                    1.9.0           py312hca03da5_0    defaults
expat                     2.6.2                h313beb8_0    defaults
fmt                       9.1.0                h48ca7d4_1    defaults
freetype                  2.12.1               h1192e45_0    defaults
frozendict                2.4.2           py312hca03da5_0    defaults
icu                       73.1                 h313beb8_0    defaults
idna                      3.7             py312hca03da5_0    defaults
jinja2                    3.1.4           py312hca03da5_0    defaults
jpeg                      9e                   h80987f9_1    defaults
[...]

PR #816, opened 2024-07-01
Add option to output file with installer hashes

marcoesters:
### Description

Add option to output hashes of installer file. The file format is written in a way that it can be verified with `shasum -c` and similar UNIX tools.

In order to work with `installer_type: all` (see issue #814), I merged the info objects before creating outputs. I'm not sure if that closes the issue, but it's a temporary workaround.

### Checklist - did you ...

### DIFF
--- a/constructor/build_outputs.py
+++ b/constructor/build_outputs.py
@@ -3,6 +3,7 @@ Additional artifacts to be produced after building the installer.
 
 Update documentation in `construct.py` if any changes are made.
 """
+import hashlib
 import json
 import logging
 import os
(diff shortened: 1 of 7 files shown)


jaimergp:
LGTM but I wonder if the output files should be one per installer, named as `{installer-name}-{version}.{extension}.{algorithm}`; e.g. for `Miniconda-24.5.0_py310.sh` we would find SHA256 in `Miniconda-24.5.0_py310.sh.sha256`. I'm thinking of these files published next to the [installers in Miniforge](https://github.com/conda-forge/miniforge/releases/tag/24.3.0-0), for example.

marcoesters:
> LGTM but I wonder if the output files should be one per installer, named as `{installer-name}-{version}.{extension}.{algorithm}`; e.g. for `Miniconda-24.5.0_py310.sh` we would find SHA256 in `Miniconda-24.5.0_py310.sh.sha256`. I'm thinking of these files published next to the [installers in Miniforge](https://github.com/conda-forge/miniforge/releases/tag/24.3.0-0), for example.

I think that makes sense since users will download only one installer and then `shasum -c` would fail. I used the entire filename for simplicity though, which Miniforge does as well.

constructor/build_outputs.py
```diff
@@ -33,14 +34,34 @@ def process_build_outputs(info):
                 f"Available keys: {tuple(OUTPUT_HANDLERS.keys())}"
             )
         outpath = handler(info, **config)
-        logger.info("build_outputs: '%s' created '%s'.", name, os.path.abspath(outpath))
+        logger.info("build_outputs: '%s' created '%s'.", name, outpath)
+
+
+def dump_hash(info, algorithm=""):
+    if algorithm not in hashlib.algorithms_available:
+        raise ValueError(f"Invalid algorithm: {', '.join(algorithm)}")
```
jaimergp, line 42:
```suggestion
        raise ValueError(f"Invalid algorithm: {algorithm}")
```

I think you mean this?

jaimergp (dismissed):

constructor/build_outputs.py
marcoesters: I originally planned to allow for multiple algorithms at once for one file. That didn't quite fit in the structure of how the outputs were handled (it only allowed for one file at a time). Now that I re-arranged it to output one hash file per installer, I added it back in.

constructor/build_outputs.py
```diff
@@ -33,14 +34,40 @@ def process_build_outputs(info):
                 f"Available keys: {tuple(OUTPUT_HANDLERS.keys())}"
             )
         outpath = handler(info, **config)
-        logger.info("build_outputs: '%s' created '%s'.", name, os.path.abspath(outpath))
+        logger.info("build_outputs: '%s' created '%s'.", name, outpath)
+
+
+def dump_hash(info, algorithm=[]):
```
jaimergp, line 40:
```suggestion
def dump_hash(info, algorithm=None):
```

jaimergp (approved):

PR #829, opened 2024-07-31
Merge configuration files into pyproject.toml

marcoesters:
### Description

Convert config files into `pyproject.toml` to use more modern python packaging practices.

The lower bounds for `setuptools` and `setuptools_scm[toml]` were taken from `menuinst`. I'm not sure if they need to be adjusted or can be removed altogether.

Also add setuptools to host dependencies since it will be removed from pip's run dependencies on conda-forge eventually.

### Checklist - did you ...


pyproject.toml
```diff
@@ -1,3 +1,44 @@
+[build-system]
+requires = ["setuptools>=45", "setuptools_scm[toml]>=6.2"]
```
jaimergp, line 2: I'd say we need 
```suggestion
requires = ["setuptools>=70.1", "setuptools_scm[toml]>=6.2"]
```

To make sure `wheel` is around. xref https://github.com/conda-forge/pip-feedstock/issues/124#issuecomment-2261091498

.flake8
jaimergp, line 1: Cannot we have this in pyproject.toml too?

constructor/_version.py
jaimergp, line 1: We might need to add this path to .gitignore.

constructor/__init__.py
```diff
@@ -1,3 +1,5 @@
-from . import _version
-
-__version__ = _version.get_versions()['version']
+try:
+    from . import _version
+    __version__ = _version.__version__
+except ImportError:
+    __version__ = "dev"
```
jaimergp, line 5: I like this [idiom](https://github.com/conda/conda-libmamba-solver/blob/c25621abd4e4b0d9726223c617eaef5fcfa22807/conda_libmamba_solver/__init__.py#L4-L13) better:

```suggestion
try:
    from ._version import version as __version__
except ImportError:
    try:
        from importlib.metadata import version

        __version__ = version("constructor")
        del version
    except ImportError:
        __version__ = "0.0.0.unknown"
```

.flake8
marcoesters: `flake8` unfortunately doesn't support `pyproject.toml`

jaimergp (approved):

.flake8
jaimergp: Got it. We could move to `ruff` in a subsequent PR and get rid of this too.

PR #847, opened 2024-08-30
Windows: Report installation messages to console

### LINKED ISSUE #764: Windows: output to stdout during command line installation
### Checklist

### What is the idea?

While a decision has not been made about a replacement for `NSIS` (#646), it might be worth exploring adding `stdout` output to the Windows command line installer.

This could be a good start: https://forums.winamp.com/forum/developer-center/nsis-discussion/261756-write-to-command-line-when-silent#post3290436

### Why is this needed?

There is currently no way to debug failures in command line installers. Using log files causes other issues (e.g., writing outside the installation directory with GUI installers when the "Browse..." button is clicked).

### What should happen?

There should be an additional flag that allows users to output to stdout (or suppress output to stdout).

### Additional Context

_No response_

### LINKED ISSUE #812: Windows installer CLI option `/S` flag doesn't include any feedback
### Checklist

### What happened?

When running the windows installer with `/S`, as expected, no installation GUI occurs. Instead, I would have expected that there is some feedback, for example the exit code.

Instead, the installer directly exits with exit code 0 and you have no way of knowing apart from opening Task Manager and searching for `python` or `_conda` or your installer running whether the installer is still running or has already finished.

For example, when providing an invalid path with `/D`, it doesn't fail but instead defaults to the default location.

I'm using conda-constructor via `pixi global install constructor` in an isolated environment.

### Conda Info

_No response_

### Conda Config

_No response_

### Conda list

_No response_

### Additional Context

_No response_

jaimergp:
### Description

- Closes #764 
- Closes #812 

Taking inspiration from the examples collected in https://github.com/conda/constructor/issues/764#issuecomment-2211160821, we can print directly to the attached console (which might _not_ be the stdout stream!) and report some information. This mostly works but we can't rely on stream redirection, piping or capture (that's why we still need the log files for testing) unless we execute the installer directly without `START /WAIT`. For most end users, this won't be an issue, I hope. The output can be silenced with `/Q` (quiet).

This only applies to `/S` mode (S for silent, but let's call it headlesS now).

This is how it looks more or less:

```
> cmd /C START /WAIT Miniforge3-4.10.1-0-Windows-x86_64.exe /S /D=C:\Users\Jaime\Miniforge-again
Welcome to Miniforge3 4.10.1-0

By continuing this installation you are accepting this license agreement:
C:\Users\Jaime\Miniforge-again\EULA.txt
Please run the installer in GUI mode to read the details.

Miniforge3 will now be installed into this location:
C:\Users\Jaime\Miniforge-again

Unpacking payload...
Setting up the package cache...
Setting up the base environment...
Installing packages for base, creating shortcuts if necessary...
Initializing conda directories...
Running post install...
Done!
```

```
> cmd /C START /WAIT Miniforge3-4.10.1-0-Windows-x86_64.exe /?
Installs Miniforge3 4.10.1-0

USAGE
-----

Miniforge3-4.10.1-0-Windows-x86_64.exe [options]

OPTIONS
-------

/InstallationType=AllUsers [default: JustMe]
/AddToPath=[0|1] [default: 0]
/KeepPkgCache=[0|1] [default: 1]
/RegisterPython=[0|1] [default: AllUsers: 1, JustMe: 0]
/NoRegistry=[0|1] [default: AllUsers: 0, JustMe: 0]
/NoScripts=[0|1] [default: 0]
/NoShortcuts=[0|1] [default: 0]
/CheckPathLength=[0|1] [default: 1]
/? (show this help message)
/S (run in CLI/headless mode)
/D=[installation directory] (must be last parameter)

EXAMPLES
--------

Install for all users, but don't add to PATH env var:
> Miniforge3-4.10.1-0-Windows-x86_64.exe /InstallationType=AllUsers

Install for just me, add to PATH and register as system Python:
> Miniforge3-4.10.1-0-Windows-x86_64.exe /RegisterPython=1 /AddToPath=1

Install for just me, with no registry modification (for CI):
> Miniforge3-4.10.1-0-Windows-x86_64.exe /NoRegistry=1

Install via CLI (no GUI) into C:\Miniforge3
> cmd /C START /WAIT Miniforge3-4.10.1-0-Windows-x86_64.exe /S /D=C:\Miniforge3

NOTE: If you install for AllUsers, then the option to AddToPath
is disabled (i.e. if /InstallationType=AllUsers, then
/AddToPath=1 will be ignored).
```

### Checklist - did you ...


marcoesters (commented):
I like this a lot. However, I think it there should be an option to suppress the output to still allow for old behavior, especially since stream redirecting doesn't work. 

Alternatively, we can make this an opt-in option, but I think outputting into the console should be the default.

Do you know how this would look like inside a GitHub Action?

jaimergp:
> I think it there should be an option to suppress the output to still allow for old behavior, especially since stream redirecting doesn't work.

Good point. You mean a CLI flag (e.g. `/Q`?) or a `constructor` option?

> Do you know how this would look like inside a GitHub Action?

Haven't checked yet, but will do.

marcoesters:
> > I think it there should be an option to suppress the output to still allow for old behavior, especially since stream redirecting doesn't work.
> 
> Good point. You mean a CLI flag (e.g. `/Q`?) or a `constructor` option?

A CLI flag

jaimergp:
> Do you know how this would look like inside a GitHub Action?

[The output doesn't show up in GHA logs](https://github.com/napari/packaging/actions/runs/10798679836/job/29952979416?pr=182#step:27:1). I guess their capture mechanism doesn't involve low level Windows APIs and "just" the stdout streams 😂 

Or am I missing something in how I set this up?

jaimergp:
Ok, it looks like stream redirection does work if invoked directly like:

```
my-installer.exe /S > out.txt
```

However that call is not blocking, so the user will have to check the logs manually via polling or something. Added this gotcha to the docs.

jaimergp:
You can see that the [direct call](https://github.com/napari/packaging/pull/182/files#diff-afc0068c04c2f82b5ba745d52a251795a27940607f0257cb7861a9d687ab8909R587-R588) does work on [GHA output now](https://github.com/napari/packaging/actions/runs/10799611276/job/29955994779?pr=182#step:27:1), but we have to block with some sleep-like statements.

docs/source/cli-options.md
```diff
@@ -104,6 +104,31 @@ Run the uninstaller in silent mode from its original location:
...
 ```
 
+:::{admonition} Stream redirection in EXE installers
+:class: info
+
+When run in `/S` mode, EXE installers will output some minimal info to the console when invoked
+like this:
+
+```batch
+> cmd.exe /c start /wait my_installer.exe /S
```
marcoesters, line 114: `cmd.exe /c` is not necessary. You could also consider adding PowerShell since this is the default profile in the Windows Terminal

```
Start-Process -FilePath my_installer.exe -ArgumentList "/S" -NoNewWindow -Wait
```

constructor/nsis/main.nsi.tmpl
```diff
@@ -284,6 +282,7 @@ FunctionEnd
                 /CheckPathLength=[0|1] [default: 1]$\n\
                 /? (show this help message)$\n\
                 /S (run in CLI/headless mode)$\n\
+                /Q (omit CLI logging)$\n\
```
marcoesters, line 285: Maybe add "quiet mode"? I don't know if I would have understood "omit CLI logging" without context.

marcoesters (commented):
I discovered another gotcha: calling the uninstaller opens a new Window for some reason.

docs/source/cli-options.md
jaimergp: Nice, you can even use `-RedirectStandardOutput out.txt` to correctly redirect things!

jaimergp:
> calling the uninstaller opens a new Window for some reason.

Yep, can reproduce. I think this is due to the fact that [the uninstaller copies itself to a temporary location](https://nsis-dev.github.io/NSIS-Forums/html/t-227826.html) so it can delete itself during the "Removing files and folders..." step. That would cause the new process to lose the parent console and hence why we get a new allocated console for that. I guess that's better than nothing so I'm inclined to leave it like that. I'm not concerned about CI usage because in those cases users don't really uninstall, just spin a new one up, right?

docs/source/cli-options.md
jaimergp: Added this to the documentation.

marcoesters: It looks like you replaced the documentation. I don't see the harm showing both `cmd.exe` and PowerShell.

marcoesters:
> > calling the uninstaller opens a new Window for some reason.
> 
> Yep, can reproduce. I think this is due to the fact that [the uninstaller copies itself to a temporary location](https://nsis-dev.github.io/NSIS-Forums/html/t-227826.html) so it can delete itself during the "Removing files and folders..." step. That would cause the new process to lose the parent console and hence why we get a new allocated console for that. I guess that's better than nothing so I'm inclined to leave it like that. I'm not concerned about CI usage because in those cases users don't really uninstall, just spin a new one up, right?

I use the uninstallation procedure in a CI, but I don't see how this would break it. And even if it does, `/Q` never attaches to the console.

However, it might be worth pointing out the new window in the documentation.

docs/source/cli-options.md
jaimergp: Added tabs with both now.

docs/source/cli-options.md
```diff
@@ -69,65 +69,83 @@ Windows installers have the following CLI options available:
...
+### Examples
 
-Run the installer in silent mode:
+Run the installer in headless mode:
 
-```batch
-> cmd.exe /c start /wait my_installer.exe /S
+`````{tab-set}
+````{tab-item} CMD
+```pwsh
```
marcoesters, line 91:
```suggestion
```batch
```

docs/source/cli-options.md
```diff
@@ -69,65 +69,83 @@ Windows installers have the following CLI options available:
...
 
-Run the uninstaller in silent mode from its original location:
+Run the installer in headless mode, for all users, adding to PATH and installing to a custom path:
 
-```batch
-> cmd.exe /c start /wait "C:\Program Files\my_app\uninstaller.exe" /S _?=C:\Program Files\my_app
+
+`````{tab-set}
+````{tab-item} CMD
+```pwsh
```
marcoesters, line 107:
```suggestion
```batch
```

docs/source/cli-options.md
```diff
@@ -69,65 +69,83 @@ Windows installers have the following CLI options available:
...
+Redirect the console output to a file named `log.txt`:
 
-When run in `/S` mode, EXE installers will output some minimal info to the console when invoked
-like this:
 
-```batch
-> cmd.exe /c start /wait my_installer.exe /S
+`````{tab-set}
+````{tab-item} CMD
+```pwsh
```
marcoesters, line 123:
```suggestion
```batch
```

docs/source/cli-options.md
```diff
@@ -69,65 +69,83 @@ Windows installers have the following CLI options available:
...
+`````
 
-Stream redirection does work as expected if run directly:
+Run the uninstaller in headless mode from its original location:
 
-```batch
-> my_installer.exe /S > logs.txt
+`````{tab-set}
+````{tab-item} CMD
+```pwsh
```
marcoesters, line 139:
```suggestion
```batch
```

docs/source/cli-options.md
jaimergp: I used pwsh on purpose here because the batch syntax highlighting didn't shown any colours at all.

marcoesters (approved):

PR #852, opened 2024-09-06
Add extra pages to pkg installers

marcoesters:
#### Summary of changes

* Add feature to create extra pages to pkg installers after the installation step, analogous to the capability for Windows installers.
* Refactor code signing for pkg installers into a class.
* Add testing for application signing.
* Expand the post-install page features to Windows.
* Expand documentation about signing pkg installers.

#### Detailed explanation

`constructor` has the capability to add additional pages for exe installer by overloading the `custom_conclusion` property. This is not possible for `pkg` installer because custom pages have a different format from conclusion pages: extra pages follow the Apple plugins format whereas conclusions are RTF or HTML.

So, a new keyword for `construct.yaml` is required. For pkg installer, this keyword can take multiple extra pages and add them to a pkg installer. These pages can either be compiled projects or simple, single-target projects. The latter is helpful for repositories since it would not require committing binaries.

The plugins must be signed to pass notarization just like the `_conda` binary. To make the code DRYer, the `codesign` commands are now a class just like the Windows signing tools. Caveat: `get_signing_command()` has a different return type because they are used differently. However, I'm happy to change the Windows return type to `list` and do the join operation elsewhere.

Signing has not been tested on MacOS yet. The challenge is to create self-signed certificates that must be added to a keychain. This is limited to application signing. For installers, a certificate has to be trusted to be found in the keychain. Trusting a certificate requires authentication and interaction with keychains, causing the job to be stuck. I have not found a way to circumvent this.

This concept has been expanded to Windows as well. While this can still be done via `custom_conclusion`, using the extra pages keyword provides a cleaner solution, especially for those who use the default conclusion page.

### Checklist - did you ...


constructor/construct.py
```diff
@@ -569,6 +569,17 @@
 If this key is missing, it defaults to a message about Anaconda Cloud.
 You can disable it altogether if you set this key to `""` (empty string).
 (MacOS only).
+'''),
+
+    ('post_install_pages', False, (list, str), '''
+Adds extra panels to the installers to be shown after installation.
```
jaimergp, line 575:
```suggestion
Adds extra pages to the installers to be shown after installation.
```

constructor/construct.py
```diff
@@ -569,6 +569,17 @@
...
+
+For pkg installers, these can be compiled installer plug-ins or
```
jaimergp, line 577:
```suggestion
For PKG installers, these can be compiled `installer` plug-ins or
```

constructor/construct.py
```diff
@@ -569,6 +569,17 @@
...
+a directories containing an Xcode project. In the latter case,
```
jaimergp, line 578:
```suggestion
directories containing an Xcode project. In the latter case,
```

constructor/construct.py
```diff
@@ -569,6 +569,17 @@
...
+constructor will try and compile the project file using `xcodebuild`.
+
+For Windows, the extra pages must be nsi files.
```
jaimergp, line 581:
```suggestion
For Windows, the extra pages must be `.nsi` files.
```

constructor/osxpkg.py
```diff
@@ -423,6 +425,79 @@ def pkgbuild_prepare_installation(info):
...
+                        "Plugin directory contains an uncompiled project,"
+                        " but xcodebuild is not available."
+                    )
+                try:
+                    subprocess.run([xcodebuild, "--help"], check=True, capture_output=True)
+                except subprocess.CalledSubprocessError:
+                    raise RuntimeError(
+                        "Plugin directory contains an uncomplied project,"
+                        " but xcodebuild requires XCode to compile plugins."
+                    )
```
jaimergp, line 457: This could be a utility function to reduce the size of the loop.

Also:

```suggestion
            if not xcodebuild:
                xcodebuild = shutil.which("xcodebuild")
                if not xcodebuild:
                    raise RuntimeError(
                        "Plugin directory contains an uncompiled project,"
                        " but xcodebuild is not available."
                    )
                try:
                    subprocess.run([xcodebuild, "--help"], check=True, capture_output=True)
                except subprocess.CalledSubprocessError:
                    raise RuntimeError(
                        "Plugin directory contains an uncompiled project, "
                        "but xcodebuild requires XCode to compile plugins."
                    )
```

examples/osxpkg_extra_pages/construct.yaml
jaimergp, line 1: Can we add a README or something in this example to quickly state how to write a plugin like the one provided? A link to documentation or a Github repo out there with some info. Right now I wouldn't even know where to start.

constructor/osxpkg.py
```diff
@@ -423,6 +425,79 @@ def pkgbuild_prepare_installation(info):
...
+            page_in_plugins = join(PLUGINS_DIR, plugin_name)
+            shutil.copytree(page, page_in_plugins)
+
+    if codesigner:
+        with NamedTemporaryFile(suffix=".plist", delete=False) as entitlements:
+            plist = {
+                "com.apple.security.cs.allow-unsigned-executable-memory": True,
+                "com.apple.security.cs.disable-library-validation": True,
+            }
+            plist_dump(plist, entitlements)
```
jaimergp, line 479: Since you implemented the plist dumper in the codesign class already, can we just pass the dict directly to the `.sign_bundle()` call below?

marcoesters: This would create a new plist file for each plugin even though they are the same entitlements file.

An added benefit here is that both branches of the `sign_bundle()` call are covered with the same example test, but that's incidental.

jaimergp (approved):
Thanks!

Issue #851, opened 2024-09-05
constructor does not respect ssl_verify set manually or in condarc


millsks:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

When executing constructor I receive SSL verify errors because of a self-signed certificate chain.  This is normally taken care of with specifying the CA cert in **ssl_verify** or by using `truststore` to let it default to the CA cert stores on the server.  No matter what I set the value of ssl_verify the value in the conda_context.ssl_verify referenced in fcp.py is always set to True.  The only way I can get it to override is to use the environment variables `CONDA_SSL_VERIFY=False` or the REQUESTS_CA_BUNDLE with the path to a specific CA cert.

Is this by design or should the **ssl_verify** set in condarc be honored?

### Additional Context

I am using micromamba 1.5.7-0 on Linux with the latest version of constructor.  The issue is happening on a closed system that I do not have the ability to share screenshots.

jaimergp:
Yea, this block is problematic:

https://github.com/conda/constructor/blob/00d82305994fb4cee0c64949dedb0e8bc4f81ccb/constructor/fcp.py#L478-L490

We'll need to improve the logic there a bit. Thanks for the report.

millsks:
Looks good from my local run with your `ssl-verify` branch. I put in `print(f"conda_context.ssl_verify: {conda_context.ssl_verify}")`  after line 463 on constructor/fcp.py to verify in the output.

``` bash
$ conda config --set ssl_verify true
$ grep ^ssl_verify ~/.condarc
ssl_verify: true
$ constructor Miniforge3 --dry-run --debug 2>&1|grep ssl_verify
conda_context.ssl_verify: True
^C
$ conda config --set ssl_verify false
$ grep ^ssl_verify ~/.condarc
ssl_verify: false
$ constructor Miniforge3 --dry-run --debug 2>&1|grep ssl_verify
conda_context.ssl_verify: False
^C
$ conda config --set ssl_verify truststore
$ grep ^ssl_verify ~/.condarc
ssl_verify: truststore
$ constructor Miniforge3 --dry-run --debug 2>&1|grep ssl_verify
conda_context.ssl_verify: truststore
^C
```

millsks:
This is the version that is installed on my local system using your branch.
``` bash
$ constructor --version
constructor 2.0.3.dev689+gbb1c92a
$ conda list|grep ^constructor
[...]

PR #860, opened 2024-09-13
Respect original ssl_verify values

jaimergp:
### Description

Closes https://github.com/conda/constructor/issues/851

### Checklist - did you ...


jaimergp:
Fix confirmed at https://github.com/conda/constructor/issues/851#issuecomment-2360118576

marcoesters:
While the fix makes sense to me, is there a way you could write a test to show it?

jaimergp:
Adding tests for this turned out to be a bit tricky (would need to patch CondaSession), so instead I went for simple and added a couple of assertions in the code. What do you think?

marcoesters (approved):

Issue #642, opened 2023-02-01
Revisit uninstallers


jaimergp:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

Research the best way to implement uninstallers. Currently, only Windows provides some, and they do not do all tasks in the conda-canonical way:

- Files are batch removed without running pre-unlink logic
- Shortcuts are removed by manually calling some menuinst logic
- Windows Registry does not use conda init --reverse (#572)
- I am not sure if the .condarc files are cleared up (related to #568)
- Environments are not removed from `~/.conda/environments.txt`

The uninstallation logic is not present on Linux or macOS either.

### Why is this needed?

Better uninstallation processes.

### What should happen?

The system should be left in a functional state, with no lingering files or configuration.

### Additional Context

Uninstallation times have been previously reported as slow, so we might need to balance speed and correctness. See https://github.com/conda/constructor/issues/168. It's mostly UI related, so as long as the output is kept in check, we should be ok. Nevertheless, timings should be analyzed.

marcoesters:
A viable path for Linux and MacOS uninstallers could be to add that feature to `conda-standalone`. `menuinst` has gotten pretty good about removing its shortcuts and the standalone binary can be used to run `conda init --reverse`.

Essentially, it could try and remove the base environment using default `conda` commands and clean up whatever needs to be cleaned up afterwards. The user would then just have to remove the standalone binary in the end.

marcoesters:
I just found that the uninstaller won't remove shortcuts for packages created in a new environment on Windows since the uninstaller doesn't know about theses prefixes. The solution here is to remove them for each environment it can find, but recursive searches in NSIS aren't ideal.

Python/conda-standalone using pathlib's glob functionality would be a much better tool.

jaimergp:
Aren't we doing `conda remove --all` for those?

marcoesters:
We only do `conda remove --all` for environments that were shipped with installers: https://github.com/conda/constructor/blob/main/constructor/winexe.py#L195-L210

This does not account for user-created environments.

Issue #568, opened 2022-09-11
Tracking issues relating to previous condarc files hampering installation


hmaarrfk:
### Summary

The goal here is to collect user stories of how previous condarc files are hampering the installation for users. I think generally, when you get an installer, you expect it to "work". But this is problematic since some commands are affected by the current .condarc in the global space.

I think many of these issues are resolved by: https://github.com/conda/constructor/pull/559

But there may still be something we can do help users get a working setup after they've installed their package. Maybe ask them if they want to delete the old condarc file and start new?

### Linked Issues & PRs

- [ ] https://github.com/conda-forge/miniforge/issues/347
- [ ] https://github.com/conda-forge/miniforge/issues/255
- [ ] https://github.com/conda-forge/miniforge/issues/298

PR #863, opened 2024-09-19
Ignore .condarc files during installation

### LINKED ISSUE #542: Bundled `conda-standalone` will inherit settings from existing .condarc files
### Checklist

### What happened?

Our bundled conda-standalone.exe will read and load existing `.condarc` configurations, which might break the solver behavior and possibly others. For example, we built the installer with specs from different channels with `channel_priority: flexible` (default) but the user running the installer has a .condarc file with `channel_priority: strict` -- potential problems.

The solution is to add a flag or environment variable to upstream `conda` to disable the search path for `.condarc` files.

### Conda Info

```shell
N/A
```

### Conda Config

```shell
N/A
```

### Conda list

```shell
N/A
```

### Additional Context

This could go to `conda/conda-standalone` too, but I'll leave it here as a reminder to enable once available, since we are the primary users of that tool.

marcoesters:
### Description

`conda-standalone` and `micromamba` load existing `.condarc` files during the installation process, which can have unintended consequences and even break the installation. These issues have been frequently reported (see #542, #568, and the issues reported therein). The behavior of the installer should be set by the installer itself and not by residual configuration files, especially since they are not cleaned up at all (see #642).

`micromamba` has a `--no-rc` option, which is also proposed to be added to `conda-standalone` (https://github.com/conda/conda-standalone/pull/99). This feature will add the `--no-rc` option with compatible standalone executables.

Closes #542 and #568.

### Checklist - did you ...

### DIFF
--- a/constructor/main.py
+++ b/constructor/main.py
@@ -201,9 +201,9 @@ def main_build(dir_path, output_dir='.', platform=cc_platform,
 
     # Add --no-rc option to CONDA_EXE command so that existing
     # .condarc files do not pollute the installation process.
-    if exe_name == "conda-standalone" and Version(exe_version) >= Version("24.9.0"):
+    if exe_type == StandaloneExe.CONDA and exe_version and exe_version >= Version("24.9.0"):
         info["_ignore_condarcs_arg"] = "--no-rc"
-    elif exe_name == "micromamba":
+    elif exe_type == StandaloneExe.MAMBA:
         info["_ignore_condarcs_arg"] = "--no-rc"
     else:
         info["_ignore_condarcs_arg"] = ""
--- a/constructor/main.py
+++ b/constructor/main.py
@@ -197,6 +197,15 @@ def main_build(dir_path, output_dir='.', platform=cc_platform,
             )
         )
 
+    # Add --no-rc option to CONDA_EXE command so that existing
+    # .condarc files do not pollute the installation process.
+    if exe_name == "conda-standalone" and Version(exe_version) >= Version("24.9.0"):
+        info["_ignore_condarcs_arg"] = "--no-rc"
+    elif exe_name == "micromamba":
+        info["_ignore_condarcs_arg"] = "--no-rc"
+    else:
+        info["_ignore_condarcs_arg"] = ""
+
     if 'pkg' in itypes:
         if (domains := info.get('pkg_domains')) is not None:
             domains = {key: str(val).lower() for key, val in domains.items()}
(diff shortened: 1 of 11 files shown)


constructor/main.py
```diff
@@ -197,6 +197,15 @@ def main_build(dir_path, output_dir='.', platform=cc_platform,
...
 
+    # Add --no-rc option to CONDA_EXE command so that existing
+    # .condarc files do not pollute the installation process.
+    if exe_name == "conda-standalone" and Version(exe_version) >= Version("24.9.0"):
+        info["_ignore_condarcs_arg"] = "--no-rc"
+    elif exe_name == "micromamba":
+        info["_ignore_condarcs_arg"] = "--no-rc"
+    else:
+        info["_ignore_condarcs_arg"] = ""
+
```
jezdez, line 208: That hardcoding of the names of binaries that constructor builds is a dealbreaker, let's find a different way

jezdez (changes requested):

constructor/main.py
marcoesters: Is your objection here that I use an if-else chain or that I use strings? The latter is the result of the `identify_conda_exe` function, so changing that will widen the scope of this PR.

Identifying whether the standalone function is `conda-standalone` is done inconsistently (see, e.g., here: https://github.com/conda/constructor/blob/49b80727bbdabce4d150e3fba4d03c953574ea1f/constructor/main.py#L95), but I wonder if a separate PR would be a better place to address the issue.

marcoesters: Here is a PR that addresses the issue: https://github.com/conda/constructor/pull/864

That would get rid of string comparisons at least and removes inconsistencies.

constructor/main.py
```diff
@@ -197,6 +197,15 @@ def main_build(dir_path, output_dir='.', platform=cc_platform,
             )
         )
 
+    # Add --no-rc option to CONDA_EXE command so that existing
+    # .condarc files do not pollute the installation process.
+    if exe_name == "conda-standalone" and Version(exe_version) >= Version("24.9.0"):
```
jaimergp, line 202:
```suggestion
    if exe_name == "conda-standalone" and exe_version and Version(exe_version) >= Version("24.9.0"):
```

Protect against `exe_version = None`.

marcoesters:
Even if approved, we should probably wait with merging this until `conda-standalone 24.9.x` is released.

jaimergp (approved):

jezdez (approved):

PR #873, opened 2024-10-04
Justme permissions

### LINKED ISSUE #872: Running Anaconda 2024.06-1 installer on Windows for "Just Me" messes with installation folder permissions
### Checklist

### What happened?

I'm trying to install Anaconda 2024.06-1 on Windows using the official installer, in graphical mode, for "Just Me", but towards the end of the installation the installer shows `Setting installation folder permissions...`, and after a very long time it finishes, having revoked my write permissions to the Anaconda installation folder. I am thus unable to install or update any packages on the `base` environment, which is annoying. Checking the `constructor` source code for this string shows that this step _shouldn't even run_ if the installer was not running with UAC elevation. At no other point in the installation does this conditional check for UAC elevation succeed (i.e., it proceeds as if it was a "Just Me" installation). A very strange situation.

Note that this appears to be a different issue from #828, where the installer was custom-built rather than an official release, was running as `SYSTEM`, and was run in the "All Users" installation mode.

The code in question that changes file permissions, found in `constructor/constructor/nsis/main.nsi.tmpl`:
```
    # To address CVE-2022-26526.
    # Revoke the write permission on directory "$INSTDIR" for Users if this is
    # being run with administrative privileges. Users are:
    #   AU - authenticated users
    #   BU - built-in (local) users
    #   DU - domain users
    ${If} ${UAC_IsAdmin}
        ${Print} "Setting installation directory permissions..."
        AccessControl::DisableFileInheritance "$INSTDIR"
        AccessControl::RevokeOnFile "$INSTDIR" "(AU)" "GenericWrite"
        AccessControl::RevokeOnFile "$INSTDIR" "(DU)" "GenericWrite"
        AccessControl::RevokeOnFile "$INSTDIR" "(BU)" "GenericWrite"
        AccessControl::SetOnFile "$INSTDIR" "(BU)" "GenericRead + GenericExecute"
        AccessControl::SetOnFile "$INSTDIR" "(DU)" "GenericRead + GenericExecute"
    ${EndIf}
```

### Additional Context

_No response_

marcoesters:
### Description

The Windows installer restricts permissions for users if the installation was done with administrator privileges. It was reported in #872 that this can create a situation where a `JustMe` installation restricts write access to the installation directory when `UAC_IsAdmin` mistakenly thinks the user is an administrator.

Instead of checking for administrative privileges, check whether the installation is an `AllUsers` installation. Privileges should be elevated at this point anyway and reflects the intent of the user. A `JustMe` installation should not be able to write into a sensitive directory in the first place since it never requires admin privileges.

Closes #872.

### Checklist - did you ...

### DIFF
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -1372,11 +1372,11 @@ Section "Install"
 
     # To address CVE-2022-26526.
     # Revoke the write permission on directory "$INSTDIR" for Users if this is
-    # being run with administrative privileges. Users are:
+    # an all-users installation. Users are:
     #   AU - authenticated users
     #   BU - built-in (local) users
     #   DU - domain users
-    ${If} ${UAC_IsAdmin}
+    ${If} $InstMode == ${ALL_USERS}
         ${Print} "Setting installation directory permissions..."
         AccessControl::DisableFileInheritance "$INSTDIR"
         AccessControl::RevokeOnFile "$INSTDIR" "(AU)" "GenericWrite"
--- a/news/873-justme-permission-bug
+++ b/news/873-justme-permission-bug
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Restrict Windows directory permission based on installation mode, not privilege level. (#872 via #873)
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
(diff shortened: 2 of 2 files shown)


marcoesters:
> (even though you could always run the installer as administrator to circumvent that).

This is exactly the situation that could re-introduce the CVE since we cannot prevent the installer from being executed with admin privileges. I am closing this PR - we need to find a better solution.

PR #892, opened 2024-11-09
Use Jinja for templating

jaimergp:
### Description

It was about time.

### Checklist - did you ...

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -1,28 +1,28 @@
 #!/bin/sh
 #
-# Created by constructor __CONSTRUCTOR_VERSION__
+# Created by constructor {{ constructor_version }}
 #
-# NAME:  __NAME__
-# VER:   __VERSION__
-# PLAT:  __PLAT__
-# MD5:   __MD5__
+# NAME:  {{ installer_name }}
+# VER:   {{ installer_version }}
+# PLAT:  {{ installer_platform }}
+# MD5:   {{ installer_md5 }}
 
 set -eu
 
-#if osx
+{%- if osx %}
 unset DYLD_LIBRARY_PATH DYLD_FALLBACK_LIBRARY_PATH
-#else
+{%- else %}
 export OLD_LD_LIBRARY_PATH="${LD_LIBRARY_PATH:-}"
 unset LD_LIBRARY_PATH
-#endif
+{%- endif %}
 
 if ! echo "$0" | grep '\.sh$' > /dev/null; then
     printf 'Please run using "bash"/"dash"/"sh"/"zsh", but not "." or "source".\n' >&2
-    return 1
+    exit 1
 fi
 
-#if osx and min_osx_version
-min_osx_version="__MIN_OSX_VERSION__"
+{%- if osx and min_osx_version %}
+min_osx_version="{{ min_osx_version }}"
 system_osx_version=$(SYSTEM_VERSION_COMPAT=0 sw_vers -productVersion)
 # shellcheck disable=SC2183 disable=SC2046
 int_min_osx_version="$(printf "%02d%02d%02d" $(echo "$min_osx_version" | sed 's/\./ /g'))"
@@ -32,9 +32,8 @@ if [  "$int_system_osx_version" -lt "$int_min_osx_version" ]; then
     echo "Installer requires macOS >=${min_osx_version}, but system has ${system_osx_version}."
     exit 1
 fi
-#endif
-#if linux and min_glibc_version
-min_glibc_version="__MIN_GLIBC_VERSION__"
+{%- elif linux and min_glibc_version %}
+min_glibc_version="{{ min_glibc_version }}"
 case "$(ldd --version 2>&1)" in
     *musl*)
         # musl ldd will report musl version; call libc.so directly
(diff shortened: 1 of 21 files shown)


marcoesters:
> This is ready, but I'm inclined to leave it outside of this release in case it breaks something. We have plenty of changes anyway. Thoughts @marcoesters?

Agreed

jaimergp:
Ok, this can go in now that 3.10 is out. @marcoesters

constructor/nsis/main.nsi.tmpl
```diff
@@ -550,13 +545,13 @@ Function .onInit
...
-    @TEMP_EXTRA_FILES@
+    {{ TEMP_EXTRA_FILES }}
     !insertmacro ParseCommandLineArgs
 
     # Select the correct registry to look at, depending
     # on whether it's a 32-bit or 64-bit installer
-    SetRegView @BITS@
-#if win64
+    SetRegView {{ BITS }}
+{%- if win64 %}
```
marcoesters, line 554: Do we want to deprecate win-32 altogether?

constructor/nsis/main.nsi.tmpl
```diff
@@ -692,44 +687,22 @@ Function .onInit
...
-    StrCpy $Ana_PreInstall_State ${BST_UNCHECKED}
-#endif
-#if post_install_exists is True
-    StrCpy $Ana_PostInstall_State ${BST_CHECKED}
-#endif
-#if post_install_exists is False
-    StrCpy $Ana_PostInstall_State ${BST_UNCHECKED}
-#endif
+{%- endif %}
+{%- endif %}
```
marcoesters, line 698: Can these conditions be indented?

constructor/winexe.py
```diff
@@ -242,70 +241,52 @@ def make_nsi(
...
+    variables = {
+        'installer_name': name,
+        'installer_version': info['version'],
+        'company': info.get('company', 'Unknown, Inc.'),
+        'installer_platform': info['_platform'],
+        'arch': '%d-bit' % arch,
+        'py_ver': ".".join(py_version.split(".")[:2]),
+        'pyversion_justdigits': ''.join(py_version.split('.')),
+        'pyversion': py_version,
+        'pyversion_major': py_version.split('.')[0],
```
marcoesters, line 253: This looks like something that can be done by jinja. That would make it more obvious in the template and developers don't have to go look for the definition of these similar python variables.

constructor/winexe.py
```diff
@@ -330,32 +311,71 @@ def make_nsi(
...
+
+    # UPPERCASE variables are unescaped (and unquoted)
+    variables['NAME'] = name
+    variables['NSIS_DIR'] = NSIS_DIR
+    variables['BITS'] = str(arch)
+    variables['PKG_COMMANDS'] = '\n    '.join(pkg_commands(download_dir, dists))
+    variables['SIGNTOOL_COMMAND'] = signing_tool.get_signing_command() if signing_tool else ""
+    variables['SETUP_ENVS'] = '\n    '.join(setup_envs_commands(info, dir_path))
+    variables['WRITE_CONDARC'] = '\n    '.join(add_condarc(info))
+    variables['SIZE'] = str(approx_pkgs_size_kb)
```
marcoesters, line 350: Do we need that conversion or will jinja handle it for us?

constructor/winexe.py
```diff
@@ -330,32 +311,71 @@ def make_nsi(
...
-    ppd["has_license"] = bool(info.get("license_file"))
-    ppd["post_install_pages"] = bool(info.get("post_install_pages"))
-
-    data = preprocess(data, ppd)
-    data = fill_template(data, replace, exceptions=nsis_predefines)
+        variables[key] = win_str_esc(value)
+
+    # From now on, the items added to variables will NOT be escaped
+
+    # These are mostly booleans we use with if-checks
```
marcoesters, line 321: Now that this is all one dictionary, this could be merged into the `variables` initializer.

constructor/header.sh
```diff
@@ -509,26 +490,22 @@ rm -f "$PRECONDA"
 CONDA_QUIET="$BATCH" \
 "$CONDA_EXEC" constructor --prefix "$PREFIX" --extract-conda-pkgs || exit 1
 
-#The templating doesn't support nested if statements
-#if has_pre_install
+{%- if has_pre_install %}
```
marcoesters, line 493: Can the nested conditionals in this file be indented?

constructor/shar.py
```diff
@@ -56,70 +55,60 @@ def get_header(conda_exec, tarball, info):
...
-        'VIRTUAL_SPECS': shlex.join(virtual_specs),
-        'NO_RCS_ARG': info.get('_ignore_condarcs_arg', ''),
-    }
+    variables['installer_name'] = name
+    variables['installer_version'] = info['version']
+    variables['installer_platform'] = info['_platform']
+    variables['installer_md5'] = hash_files([conda_exec, tarball])
+    variables['default_prefix'] = info.get('default_prefix', '${HOME:-/opt}/%s' % name.lower())
+    variables['first_payload_size'] = str(getsize(conda_exec))
+    variables['second_payload_size'] = str(getsize(tarball))
```
marcoesters, line 89: Isn't the string conversion done by jinja?

constructor/osxpkg.py
```diff
@@ -332,29 +331,25 @@ def move_script(src, dst, info, ensure_shebang=False, user_script_type=None):
...
-    data = fill_template(data, replace)
-
-    data = data.replace("_SCRIPT_ENV_VARIABLES_=''", '\n'.join(
-        [f"export {key}='{value}'" for key, value in custom_variables.items()]))
+    variables["pkg_name_lower"] = pkg_name_lower
+    variables["installer_name"] = info['name']
+    variables["installer_version"] = info['version']
+    variables["installer_platform"] = info['_platform']
+    variables["channels"] = ','.join(get_final_channels(info))
+    variables["write_condarc"] = '\n'.join(add_condarc(info))
```
marcoesters, line 339: Simple join statements can be done by jinja, which I think that makes it more transparent. You may want to change that consistently throughout your template variables.

marcoesters (commented):
This looks great!

I just have a couple of comments where I think we could use jinja a little more. If we use jinja's functions more, formatting and data are more cleanly separated and give developers a better idea about how the final product looks like without having to jump between py file and template.

constructor/nsis/main.nsi.tmpl
jaimergp: Is NSIS whitespace sensitive? Only one way to know I guess 👀

constructor/winexe.py
jaimergp: Agreed. I didn't want to move too many things at first so we can do a 1:1 translation of variables and so on. We can then optimize the Jinja/Python split better. WDYT if we do this in a new PR?

constructor/winexe.py
jaimergp: Probably not needed, true.
```suggestion
    variables['SIZE'] = approx_pkgs_size_kb
```

constructor/shar.py
jaimergp:
```suggestion
    variables['first_payload_size'] = getsize(conda_exec)
    variables['second_payload_size'] = getsize(tarball)
```

constructor/nsis/main.nsi.tmpl
jaimergp: Let's do that in another PR for better traceability

constructor/header.sh
jaimergp: Check last commits.

constructor/osxpkg.py
jaimergp: I'd like to do a 2nd PR where we leverage Jinja features better. This PR would get too overwhelming otherwise, I feel.

constructor/winexe.py
jaimergp: We need to do it here separately because the block above have been stringified and escaped, but we want true booleans for these ones.

constructor/osxpkg.py
```diff
@@ -332,29 +331,25 @@ def move_script(src, dst, info, ensure_shebang=False, user_script_type=None):
...
-        'VIRTUAL_SPECS': shlex.join(virtual_specs),
-        'NO_RCS_ARG': info.get('_ignore_condarcs_arg', ''),
-    }
-    data = preprocess(data, ppd)
-    custom_variables = info.get('script_env_variables', {})
-    data = fill_template(data, replace)
-
-    data = data.replace("_SCRIPT_ENV_VARIABLES_=''", '\n'.join(
-        [f"export {key}='{value}'" for key, value in custom_variables.items()]))
+    variables["pkg_name_lower"] = pkg_name_lower
```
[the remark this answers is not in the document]
jaimergp: It's a bit more involved... we need access to the `info` dict in Jinja if we want to do that. See line 324.

constructor/winexe.py
marcoesters: That sounds fair to me.

jaimergp:
There you go: https://github.com/conda/constructor/issues/901, thanks!

PR #898, opened 2024-11-14
Generate explicit lockfiles per environment

jaimergp:
### Description

I want this to be able to release lockfiles for a given installer, so local clients can create the same environment as shipped in the installer without downloading the whole thing again. Useful for application deployments that have a pretty minimal `base` environment, and then an application+version specific environment. A local upgrade would then just consist of creating a new environment with the given lockfile, and the user would obtain the _same_ thing as distributed in the new installer.

### Checklist - did you ...


jaimergp:
Real world example: https://github.com/napari/packaging/actions/runs/11847608732/job/33017821612?pr=201#step:20:39

marcoesters (approved):

PR #914, opened 2024-12-16
fix: Adapt shell init for mamba 2

jjerphan:
### Description

Fix https://github.com/conda-forge/miniforge/issues/697.

### Checklist - did you ...

---

I am interested to have some guidance on the best way to test it.

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -671,11 +671,22 @@ if [ "$BATCH" = "0" ]; then
             *) "$PREFIX/bin/python" -m conda init ;;
         esac
         if [ -f "$PREFIX/bin/mamba" ]; then
-            case $SHELL in
-                # We call the module directly to avoid issues with spaces in shebang
-                *zsh) "$PREFIX/bin/python" -m mamba.mamba init zsh ;;
-                *) "$PREFIX/bin/python" -m mamba.mamba init ;;
-            esac
+            # If the version of mamba is <2.0.0, we preferably use the `mamba` python module
+            # to perform the initialization.
+            #
+            # Otherwise (i.e. as of 2.0.0), we use the `mamba shell init` command
+            if [ "$("$PREFIX/bin/mamba" --version | cut -d' ' -f2 | cut -d'.' -f1)" -lt 2 ]; then
+                case $SHELL in
+                    # We call the module directly to avoid issues with spaces in shebang
+                    *zsh) "$PREFIX/bin/python" -m mamba.mamba init zsh ;;
+                    *) "$PREFIX/bin/python" -m mamba.mamba init ;;
+                esac
+            else
+                case $SHELL in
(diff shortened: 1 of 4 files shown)


travishathaway:
@conda-bot check

news/914-shell-init-on-mamba-2
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Add support of mamba 2
```
jaimergp, line 7:
```suggestion
* Add support of mamba v2 shell initialization. (#914)
```

constructor/header.sh
```diff
@@ -671,11 +671,22 @@ if [ "$BATCH" = "0" ]; then
             *) "$PREFIX/bin/python" -m conda init ;;
         esac
         if [ -f "$PREFIX/bin/mamba" ]; then
```
jaimergp, line 673: Seeing this code now, I think it might result into double initialization, right? Once with conda, once with mamba. Are those compatible or should they be exclusive?

jjerphan: I think conda's initialization is overridden by mamba's; so it works, but it might not be totally sane. I guess we can keep the same behavior when 1.x is used unchanged, and see potential adaptations for 2.x.

What do you think? Is there a way to test that this works with both versions?

jaimergp: We'll need to add cases for it in the `examples/` directory. Maybe you can repurpose the `Miniforge` example (which might need to get synced up with current Miniforge scripts anyway), and then parametrize with some environment variables as we are doing in some other tests. Check for `os.environ` in the `construct.yaml` templates and also `monkeypatch.setenv` calls in `tests/test_examples.py`.

jaimergp:
I think you'll probably want some `mamba` specific tests in https://github.com/jjerphan/constructor/blob/fix/shell-init-on-mamba-2/examples/miniforge/test_install.sh.

jaimergp (approved):

jaimergp:
Thanks!

jjerphan:
Thank you for the guidance and the responsiveness.

jjerphan:
@hmaarrfk: should we redo a PR on Miniforge to use 2.0.5 adding tests there for the installation on all platforms once a new version of `constructor` is released?

hmaarrfk:
yes that would be great!

jjerphan:
@jaimergp: how can we help with a new release of `constructor`?

jaimergp:
We have a few items in the news/ dir, so we can try to cut something next week! WDYT @marcoesters? Are you planning any other PRs soon that would require waiting?

marcoesters:
I think we should submit a patch release for sure because of #922 at least. We can cut a release on Monday.

jjerphan:
FYI, I am using now an installation of miniforge from the generated shell script without any problem.

I think it would help to add test on `conda-forge/miniforge` for all supported platforms.

PR #915, opened 2024-12-18
Implement `INSTALLER_UNATTENDED` for pkg installers

marcoesters:
### Description

Implement `INSTALLER_UNATTENDED` for `pkg` installers via the `COMMAND_LINE_INSTALL` environment variable. See the bottom of the [man page](https://www.unix.com/man-page/OSX/8/installer/) for the documentation of this variable.

### Checklist - did you ...

### DIFF
--- a/constructor/osx/run_user_script.sh
+++ b/constructor/osx/run_user_script.sh
@@ -30,7 +30,7 @@ export INSTALLER_NAME="{{ installer_name }}"
 export INSTALLER_VER="{{ installer_version }}"
 export INSTALLER_PLAT="{{ installer_platform }}"
 export INSTALLER_TYPE="PKG"
-export INSTALLER_UNATTENDED="?"
+export INSTALLER_UNATTENDED="${COMMAND_LINE_INSTALL:-0}"
 export PRE_OR_POST="{{ pre_or_post }}"
 {{ script_env_variables }}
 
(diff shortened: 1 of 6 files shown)


constructor/osx/run_user_script.sh
```diff
@@ -30,7 +30,7 @@ export INSTALLER_NAME="{{ installer_name }}"
 export INSTALLER_VER="{{ installer_version }}"
 export INSTALLER_PLAT="{{ installer_platform }}"
 export INSTALLER_TYPE="PKG"
-export INSTALLER_UNATTENDED="?"
+export INSTALLER_UNATTENDED="${COMMAND_LINE_INSTALL:-0}"
```
jaimergp, line 33: Is `COMMAND_LINE_INSTALL` guaranteed to be set to `1`? The `man` page doesn't specify 🤔

marcoesters: It doesn't specify, but that's what my tests have shown (and the tests here pass). I will see if I can find explicit documentation on it before mark this ready for review.

jaimergp: It might be safer (but a bit more code) to check if it's non-empty and set UNATTENDED to `1` in that case, `0` otherwise.

marcoesters: I think I found an easy way to do this: if we just follow the documentation, the value is either unset or set to something. So, we set it to 0 if interactive and if not, we set it to 1 instead of relying on an undocumented value.

jaimergp (approved):

PR #916, opened 2024-12-20
also catch OSError when detecting conda_exe for cross-platform builds

bollwyvl:
### Description

This PR handles some observations when `--platform` and `--conda-exe` are specified.

- before the `utils.py` patch:
```
OSError: [Errno 8] Exec format error: '~/build/standalone/linux-aarch64/standalone_conda/conda.exe'
```
- without the `main.py` tpach:
```
  File "~/build/lib/python3.12/site-packages/constructor/shar.py", line 74, in get_header
    variables['enable_shortcuts'] = str(info['_enable_shortcuts']).lower()
                                        ~~~~^^^^^^^^^^^^^^^^^^^^^
  File "~/build/lib/python3.12/site-packages/ruamel/yaml/comments.py", line 853, in __getitem__
    return ordereddict.__getitem__(self, key)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: '_enable_shortcuts'
```

Presently, building with `constructor 3.11` on `python 3.12.1` on `linux-64`:

### Checklist - did you ...

  - > are there tests of cross-platform building?
  - > is there documentation about cross-platform building?

### DIFF
--- a/constructor/main.py
+++ b/constructor/main.py
@@ -187,26 +187,27 @@ def main_build(dir_path, output_dir='.', platform=cc_platform,
             else:
                 env_config[config_key] = value
 
-    if exe_type is None or exe_version is None:
-        logger.warning(
-            "Could not identify conda-standalone / micromamba version. "
-            "Will assume it is compatible with shortcuts."
+    # Installers will provide shortcut options and features only if the user
+    # didn't opt-out by setting every `menu_packages` item to an empty list
+    info['_enable_shortcuts'] = bool(
+        info.get("menu_packages", True)
+        or any(
+            env.get("menu_packages", True)
+            for env in info.get("extra_envs", {}).values()
         )
-    elif sys.platform != "win32" and (
-        exe_type != StandaloneExe.CONDA or (exe_version and exe_version < Version("23.11.0"))
-    ):
-        logger.warning("conda-standalone 23.11.0 or above is required for shortcuts on Unix.")
-        info['_enable_shortcuts'] = "incompatible"
-    else:
-        # Installers will provide shortcut options and features only if the user
-        # didn't opt-out by setting every `menu_packages` item to an empty list
-        info['_enable_shortcuts'] = bool(
-            info.get("menu_packages", True)
-            or any(
-                env.get("menu_packages", True)
-                for env in info.get("extra_envs", {}).values()
+    )
+
+    if info['_enable_shortcuts']:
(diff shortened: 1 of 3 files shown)


jaimergp:
> are there tests of cross-platform building?

I don't think so, but it would be trivial to write some. Take any of the examples and add `--platform`. In `osx-arm64` we can even run the installer with Rosetta to have a proper integration test.

jaimergp:
We'll need the news file but this looks very good to me, thanks.

bollwyvl:
Hmm, ok, then we also need the other conda-standalone up on there...

bollwyvl:
On account of me not going to sign anything that puts me in any kind of legal relationship with Anaconda, Inc, let's consider this a "complex bug report" rather than "a material contribution", whereby I release this code to the public domain. I do hope someone take this contribution and generates a derived work without any obligation to credit me or this work.

constructor/main.py
```diff
@@ -187,26 +187,27 @@ def main_build(dir_path, output_dir='.', platform=cc_platform,
             else:
                 env_config[config_key] = value
 
-    if exe_type is None or exe_version is None:
-        logger.warning(
-            "Could not identify conda-standalone / micromamba version. "
-            "Will assume it is compatible with shortcuts."
+    # Installers will provide shortcut options and features only if the user
+    # didn't opt-out by setting every `menu_packages` item to an empty list
```
marcoesters, line 191:
```suggestion
    # Installers will disable shortcut options and features only if the user
    # opted-out by setting every `menu_packages` item to an empty list
```

I know this wasn't you, but I think this describes what's going on here better since you actively have to opt out instead of opting in.

marcoesters:
> On account of me not going to sign anything that puts me in any kind of legal relationship with Anaconda, Inc, let's consider this a "complex bug report" rather than "a material contribution", whereby I release this code to the public domain. I do hope someone take this contribution and generates a derived work without any obligation to credit me or this work.

I was about to say that the CLA is with Conda, not Anaconda, but the CLA surprisingly mentions Anaconda a lot.

@jaimergp, I think this needs to be looked at. Would [conda/infastructure](https://github.com/conda/infrastructure/) be the right place?

marcoesters:
Looks like this is a work in progress already: https://github.com/conda/cla/issues/1

jaimergp:
Closed by https://github.com/conda/constructor/pull/952

PR #938, opened 2025-02-04
Enforce LF line endings for checksum files

marcoesters:
### Description

On Windows, the checksum files are output with `CRLF` line endings, which breaks the `shasum -c` check because it will treat `\r` as part of the file name. Since the output files are written to work with `shasum`, POSIX line endings needs to be enforced.

I also decided to refactor the tests and parametrize them to cover every use case.

### Checklist - did you ...


jaimergp (approved):
This can be surprising for some users on Windows, so let's make sure there's a note in the documentation about it.

LGTM otherwise.

jaimergp (approved):

PR #939, opened 2025-02-14
Fix ownership of files created by `conda init` for `PKG` installers.

marcoesters:
### Description

When PKG installers are installed outside the `$HOME` directory, the installer installs as `root` (even though `$USER` is still the user executing the installer). This causes files and directories created by `conda init` to be owned by `root`. The current script appears to only pick up the first modified file so that some files are still owned by root. This PR fixes the issue by using a more bash v3 compliant way to read the multi-line string.

I'm not sure how this can be tested on the CI since the only time this problem appears is when the PKG installers are run interactively and the destination path is not inside `$HOME`. I don't know if this can be replicated on the CI.

However, I have created a Miniconda installer with this script and it had the correct permissions on macOS 12 and 15. It even fixed a previous installation that had the incorrect permissions.

### Checklist - did you ...

- [ ] ~Add / update necessary tests?~

### DIFF
--- a/constructor/osx/update_path.sh
+++ b/constructor/osx/update_path.sh
@@ -18,7 +18,10 @@ INIT_FILES=$("$PREFIX/bin/python" -m conda init --all | tee)
 # have the correct owner.
 if [[ "${USER}" != "root" ]]; then
     echo "Fixing permissions..."
-    read -r -a MODIFIED_FILES <<< "$(\
+    MODIFIED_FILES=()
+    while read -r line; do
+        MODIFIED_FILES+=("$line")
+    done <<< "$(\
       echo "${INIT_FILES}" |\
       awk '/modified/{print $2}' |\
       # Only grab files inside $HOME or $PREFIX.
--- a/constructor/osx/update_path.sh
+++ b/constructor/osx/update_path.sh
@@ -23,13 +23,18 @@ if [[ "${USER}" != "root" ]]; then
         MODIFIED_FILES+=("$line")
     done <<< "$(\
       echo "${INIT_FILES}" |\
-      awk '/modified/{print $2}' |\
+      grep -E "^modified" |\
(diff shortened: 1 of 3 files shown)


constructor/osx/update_path.sh
```diff
@@ -18,7 +18,10 @@ INIT_FILES=$("$PREFIX/bin/python" -m conda init --all | tee)
 # have the correct owner.
 if [[ "${USER}" != "root" ]]; then
     echo "Fixing permissions..."
-    read -r -a MODIFIED_FILES <<< "$(\
+    MODIFIED_FILES=()
+    while read -r line; do
+        MODIFIED_FILES+=("$line")
```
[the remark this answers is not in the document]
marcoesters: I don't think it's an issue because you cannot install into a path with spaces when `conda` is part of the installer (at least, `constructor` won't allow it). And without `conda`, the initialization doesn't run. The other possible location is `$HOME`, which I don't think allows spaces either.

However, I came up with a way to make it more robust against space characters anyway.

constructor/osx/update_path.sh
```diff
@@ -18,15 +18,23 @@ INIT_FILES=$("$PREFIX/bin/python" -m conda init --all | tee)
...
+    done <<< "$(\
       echo "${INIT_FILES}" |\
-      awk '/modified/{print $2}' |\
+      grep -E "^modified" |\
+      sed -e 's/^modified */' |\
```
jaimergp, line 27: Hm, don't we need a second clause for the replacement in `s/`? i.e. 
```suggestion
      sed -e 's/^modified *//' |\
```

marcoesters: Yes, it looks like there was a copy-paste error from my virtual machine that I used for testing. I will double check that things are identical before re-requesting a review.

jaimergp (approved):

PR #942, opened 2025-02-25
correct extra files ref issue introduced in 0c80f9b

NixM0nk3y:
### Description

Building windows installers using the latest version of constructor (3.11.1) has broken `extra_files`.  Reverting to 3.11.0 solves the issues.  This looks to have been introduced in 0c80f9b with the recent jinga changes and the casing of the extra_files var was changed.

### Checklist - did you ...

- [ x] Add a file to the `news` directory ([using the template](https://github.com/conda/constructor/blob/main/news/TEMPLATE)) for the next release's release notes?
     
### DIFF
--- a/news/942-fix-win-extra_files
+++ b/news/942-fix-win-extra_files
@@ -4,7 +4,7 @@
 
 ### Bug fixes
 
-* Fix extra_file functionality on exe installers. (#942)
+* Fix `extra_files` functionality on EXE installers. (#942)
 
 ### Deprecations
 
--- a/news/942-fix-win-extra_files
+++ b/news/942-fix-win-extra_files
@@ -1,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fix extra_file functionality on exe installers. (#942)
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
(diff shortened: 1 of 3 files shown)


news/942-fix-win-extra_files
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* Fix extra_file functionality on exe installers. (#942)
```
jaimergp, line 7:
```suggestion
* Fix `extra_files` functionality on EXE installers. (#942)
```

jaimergp:
Hm, I wonder why the tests didn't catch that if that's indeed the bug 🤔 Let's see.

NixM0nk3y:
A quick scan over the testing , could be wrong but I dont see `NSIS_SCRIPTS_RAISE_ERRORS` being set anywhere ?

marcoesters:
> A quick scan over the testing , could be wrong but I dont see `NSIS_SCRIPTS_RAISE_ERRORS` being set anywhere ?

I agree with that assessment. Can you add this to the environment in the test workflow? I will run the workflow afterwards. Thanks for catching that!

marcoesters:
This looks good to me overall - once the CLA is signed and pushed through, I will approve and merge.

PR #950, opened 2025-03-04
Set `CONDA_QUIET=1` for exe installers

### LINKED ISSUE #944: .exe from conda/constructor crashes w/3.11.1 and 3.10.0 while 3.9.3 works fine
### Checklist

### What happened?

A build recipe, which has worked in the past (though I have made some tweaks to it since then), fails in the newest constructor versions, but runs in 3.9.3. In all constructor versions, the creation of the self-installer works fine, but the self-installer fails, usually silently (the installer window just fades off the screen), but I have also seen an error message "::error:: Failed to link extracted packages to...". I am uncertain if that latter error occurs only with 3.10.0 or is just intermittent. I have two .bat scripts that are called during the .exe run ([post-link.bat](https://github.com/AdvancedPhotonSource/GSAS-II-buildtools/blob/main/install/GSAS2new/post-link.bat) from my .conda package and the [constructor post_install .bat](https://github.com/AdvancedPhotonSource/GSAS-II-buildtools/blob/main/install/GSAS2new/g2postinstall.bat)). I tried removing them in case that was the problem, but the .exe continued to fail.  

The build I am doing is done in a GitHub workflow, with [.yml here](https://github.com/AdvancedPhotonSource/GSAS-II-buildtools/blob/main/.github/workflows/gsas2new.yml) and all files for build [here](https://github.com/AdvancedPhotonSource/GSAS-II-buildtools/tree/main/install/GSAS2new). Note that the construct.yaml is created by script, but I include a copy here. 

In any case, I suspect this is impossible to track down without more diagnostics. Is there a way to get more information on what is actually happening inside the self-installer? 

### Additional Context

Clip from 3.11.1 installer failing "silently"
https://github.com/user-attachments/assets/fba8822e-6e9a-4851-af4d-a329f97c54ed

example construct.yaml: [construct.yaml.zip](https://github.com/user-attachments/files/19030616/construct.yaml.zip)

marcoesters:
### Description

Since `conda 24.11.0`, transactions have a spinner instead of the "working" keyword to indicate progress. This spinner is not handled well in the NSIS log, writing characters for each movement of the spinner. For slow installations or large installer files, this results in very long lines.

It is suspected that this causes an overflow and crashing installers. This is evidenced by the fact that a crashing installer can be made to work when setting `CONDA_QUIET=1` (or downgrading `conda-standalone` to `v24.9.2.`, which doesn't have the spinner).

To avoid these overflow crashes, set `CONDA_QUIET=1` for EXE installers. Closes #944.

Screenshot 2025-03-04 at 11 29 31 AM
Screenshot 2025-03-04 at 11 34 08 AM

### Checklist - did you ...

### DIFF
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -1296,6 +1296,9 @@ Section "Install"
     System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_EXTRA_SAFETY_CHECKS", "no").r0'
     System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_ROOT_PREFIX", "$INSTDIR")".r0'
     System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_PKGS_DIRS", "$INSTDIR\pkgs")".r0'
+    # Spinners in conda write a new character with each movement of the spinner.
+    # For long installation times, this may cause a buffer overflow, crashing the installer.
+    System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_QUIET", "1")".r0'
     # Extra info for pre and post install scripts
     # NOTE: If more vars are added, make sure to update the examples/scripts tests too
     #       There's a similar block for the pre_uninstall script, further down this file.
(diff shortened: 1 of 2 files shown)


larsoner:
I think this might close https://github.com/conda/conda-standalone/issues/121 since setting `CONDA_QUIET=1` fixed things in testing over in https://github.com/mne-tools/mne-installers/pull/313.

jaimergp (approved):

PR #962, opened 2025-03-20
Unset `CONDARC` and `MAMBARC` during installation

marcoesters:
### Description

Pre-existing configuration files should not influence the installation process, which is why standalone binaries are currently called with the `--no-rc` option. If `CONDARC` is set, this option is overwritten for `conda-standalone`. For `micromamba`, using `--no-rc` while `CONDARC` or `MAMBARC` are set results in an error.

To prevent this from happening, unset `CONDARC` and `MAMBARC` during the installation process.

I ran the tests with and without the code changes for `exe` on Windows and `sh` installers on Linux and they fail, so them passing is a sign that the changes are working as expected.

### Checklist - did you ...

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -78,6 +78,9 @@ export INSTALLER_NAME='{{ installer_name }}'
 export INSTALLER_VER='{{ installer_version }}'
 export INSTALLER_PLAT='{{ installer_platform }}'
 export INSTALLER_TYPE="SH"
+# Installers should ignore pre-existing configuration files.
+unset CONDARC
+unset MAMBARC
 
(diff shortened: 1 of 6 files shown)


isuruf (approved):

jaimergp:
LGTM! 

@jjerphan @klaim, anything weird here?

jjerphan:
I cannot think of a case to which disabling `MAMBRC` and `CONDARC` would be harmful.

jaimergp:
I'm sure we'll hear about it if that's the case 🤞 Let's merge!

jaimergp (approved):

PR #943, opened 2025-02-28
Define JSON Schema for `construct.yml`

jaimergp:
### Description

This PR converts our venerable `construct.py` "schema" to a proper JSON Schema. The Schema is built with Pydantic at "commit time" as we do with docs regeneration. At runtime, only the lightweight `jsonschema` package is used for validation.

With this change we have:

- IDE autocompletion (examples have been edited to point to local copy; should just work if you open it in VS Code). I'll also submit it to schemastore to associate `construct.yml` to the schema in the repo
- Accurate type validation with detailed error reports
- Self documented schema

However we have lost the ability to report type information in the docs (the autogenerated annotations are too wordy and noisy). I'm not too worried about this because the information was sometimes inaccurate anyway. Instead users are encouraged to use the schema directly, either in their IDE, or exploring it interactively in apps like https://json-schema.app/view/%23?url=https%3A%2F%2Fraw.githubusercontent.com%2Fjaimergp%2Fconstructor%2Frefs%2Fheads%2Fpydantic-schema%2Fconstructor%2Fdata%2Fconstruct.schema.json.

### Checklist - did you ...

- [X] Add a file to the `news` directory ([using the template](https://github.com/conda/constructor/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [X] Add / update outdated documentation?

Additional context:

@bollwyvl mentioned this in a [conda-forge Zulip thread](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/Autocompletion.20and.20.22menuinst.22/near/502591185) and I couldn't resist to attempt a crude conversion. The first commit already gives you inline descriptions and some type validation.

Plenty of work ahead for the very diverse types and actual default values, plus integration with `constructor` itself.

For fun and laughs, this is the crude script that I used to convert our venerable `KEYS` "schema" to a Pydantic model:

```python
from constructor.construct import KEYS
from textwrap import indent

print("""
from pydantic import BaseModel, ConfigDict

class ConstructorModel(BaseModel):
    model_config: ConfigDict = ConfigDict(
        extra='forbid',
        use_attribute_docstrings=True,
    )
"""
)

for key, required, type, desc in KEYS:
    type_name = getattr(type, "__name__", getattr(type.__class__, "__name__"))
    print(f"    {key}: {type_name}", "= ..." if required else f"= {type_name}()")
    print('    """')
    print(indent(desc.strip(), "    "))
    print('    """')
```

PR #965, opened 2025-03-24
Add support for `conda init --condabin`, which adds `condabin/` to PATH

### LINKED ISSUE #960: Support adding `condabin/` to PATH
### Checklist

### What is the idea?

constructors supports adding `bin` (on Windows, `Scripts`) to PATH, but it's discouraged because it pollutes the executable space.

We should provide an option where the installer _only_ offers to add `condabin` to PATH, so `conda` (and `mamba`) can be in PATH without shell initialization.

### Why is this needed?

This Miniforge issue provides a good overview of the rationale: https://github.com/conda-forge/miniforge/issues/453

### What should happen?

If "Add condabin to PATH" is enabled in the installer and selected by the user, then `$INSTDIR/condabin` would be in PATH after the installation.

### Additional Context

This could also be tackled in `conda` as part of `conda init`, but I think this is cleaner? Shell initialization does not necessarily need PATH entries, and viceversa. Open for feedback though.

jaimergp:
### Description

Closes https://github.com/conda/constructor/issues/960.

Based on https://github.com/conda/constructor/pull/943 to avoid conflicts down the line because this adds new options to `construct.yaml`. Excuse the diff while 943 gets reviewed.

Windows:

image

macOS:

image

### Checklist - did you ...


jaimergp:
Actual diff at [`56e35a6` (#965)](https://github.com/conda/constructor/pull/965/commits/56e35a6390a9c34c01e6456afc6004ce9ffbb9e1) for now

jaimergp:
@marcoesters, I think this is ready! It'd be nice to check if it behaves as intended locally too. Maybe I can extend the tests to actually check for `PATH` values but we are already testing that in the `conda` test suite 🤷 

Also, I'd like to hear what you think about the proposed UI/text changes. I'd be happy to change whatever there, not feeling particularly strongly, but I do want to make it less confusing for end users.

CONSTRUCT.md
```diff
@@ -502,6 +502,21 @@ is able to change the default during interactive installation. NOTE: For Windows
...
+### `add_condabin_to_path`
+
+Add an option to the installer so the user can choose whether to add the `condabin/`
+directory to PATH. Only applicable if `conda` is part of the installation.
+On Linux and macOS, `conda >=25.5.0` is required. See also `add_condabin_to_path_default`.
+
+### `add_condabin_to_path_default`
+
+Default value for the option added by `add_condabin_to_path`. The default
+is true for GUI installers (EXE, PKG) and false for shell installers. The user
```
soapy1: what is the motivation behind making this default value different for different types of installers?

constructor/nsis/_nsis.py
```diff
@@ -296,6 +296,16 @@ def add_to_path(pyversion, arch):
     broadcast_environment_settings_change()
 
 
+def add_condabin_to_path():
+    from _system_path import (
+        add_to_system_path,
+        broadcast_environment_settings_change,
+    )
+
+    add_to_system_path(os.path.normpath(os.path.join(ROOT_PREFIX, "condabin")), allusers)
```
soapy1, line 317: I see in `constructor/nsis/main.nsi.tmpl` that addings condabin to path is disabled for alluser installs to address CVE-2022-26526. Should this piece of code also protect against this? Like:

```
if allusers:
  # log error, throw error?
else:
  add_to_system_path(os.path.normpath(os.path.join(ROOT_PREFIX, "condabin")), allusers)

**Review by soapy1 — COMMENTED** [review]
Looks great! Just a couple of questions

**Comment by jaimergp** [comment]
> On Windows, I'm not sure it's obvious what the benefit of one option over the other is, so having both available is not a great idea.

It's on the installer creator to offer both. By default, the `condabin` option is disabled. The installer creator has to opt-in. I would assume that at that point they will know what to do about the classic initialization option (disable it). I tried to discuss this in the docs.

> choose what they want /AddToPath to add to the path 

Oh true, I should add a command line option for this too.

File: constructor/nsis/_nsis.py (hunk 2 above)
Comment (jaimergp) [reply constructor/nsis/_nsis.py:317]: Thanks for spotting this!

File: CONSTRUCT.md (hunk 1 above)
Comment (jaimergp) [reply CONSTRUCT.md]: Consistency with previous UX. I assume the notion is that GUI users don't want to mess too much with the terminal, while shell installer users would be more ok with some tinkering. It's a bad proxy though.

File: constructor/header.sh [hunk 3]
```diff
@@ -123,18 +124,34 @@ Installs ${INSTALLER_NAME} ${INSTALLER_VER}
 -u           update an existing installation
 {%- if has_conda %}
 -t           run package tests after installation (may install conda-build)
+{%-   if add_condabin_to_path %}
+{%-     if add_condabin_to_path_default %}
+-z           do not add PREFIX/condabin to PATH; ignored in interactive mode
+{%-     else %}
+-c           add PREFIX/condabin to PATH; ignored in interactive mode
+{%-     endif %}
+{%-   endif %}
```
marcoesters: I am concerned about adding these additional, complementary flags here. #982 already tried to implement an extra flag for another option. If we add #1004 to the mix, I am concerned about maintainability.

The order of the flags is also getting a little chaotic. I wonder if we should keep the existing behavior of adding to path and initializing should be off in non-interactive mode unless opted in.

constructor/main.py
```diff
@@ -262,6 +262,14 @@ def main_build(
     # '_dists': list[Dist]
     # '_urls': list[Tuple[url, md5]]
 
+    if info.get("add_condabin_to_path"):
+        if not info.get("_has_conda"):
+            sys.exit("Error: 'add_condabin_to_path' requires 'conda' in the base env.")
```
marcoesters:
```suggestion
            sys.exit("Error: 'add_condabin_to_path' requires 'conda' in the base environment.")
```

constructor/nsis/main.nsi.tmpl
```diff
@@ -281,6 +283,9 @@ FunctionEnd
             $\n\
                 /InstallationType=AllUsers [default: JustMe]$\n\
                 /AddToPath=[0|1] [default: 0]$\n\
+{%- if add_condabin_to_path %}
+                /AddCondabinToPath=[0|1] [default: 0]$\n\
```
marcoesters: Should the default be set here to `add_condabin_to_path_default`?

constructor/nsis/main.nsi.tmpl
```diff
@@ -714,6 +740,16 @@ Function .onInit
     {%- endif %}
 {%- endif %}
 
+{%- if add_condabin_to_path %}
+    {%- if add_condabin_to_path_default %}
+    ${If} $InstMode == ${JUST_ME}
+        StrCpy $Ana_AddCondabinToPath_State ${BST_CHECKED}
+    ${EndIf}
```
marcoesters: Should `${Else}` case be added here to be safe?

constructor/shar.py
```diff
@@ -74,10 +74,12 @@ def get_header(conda_exec, tarball, info):
             variables["direct_execute_%s" % key] = has_shebang(info[key])
     variables["initialize_conda"] = info.get("initialize_conda", True)
     variables["initialize_by_default"] = info.get("initialize_by_default", False)
+    variables["add_condabin_to_path"] = info.get("add_condabin_to_path", True)
```
marcoesters: This contradicts the documented default value for `add_condabin_to_path`

constructor/winexe.py
```diff
@@ -165,6 +165,11 @@ def make_nsi(
         "enable_shortcuts": "yes" if info["_enable_shortcuts"] is True else "no",
         "show_register_python": "yes" if info.get("register_python", True) else "no",
         "show_add_to_path": "yes" if info.get("initialize_conda", True) else "no",
+        "show_add_condabin_to_path": (
+            "yes"
+            if info.get("add_condabin_to_path", True) and info.get("_has_conda", False)
```
marcoesters: The default value here is different from the documented default and contradicts the `None` a few lines below.

docs/source/howto.md
```diff
@@ -201,3 +201,58 @@ finding these files may rely on environment variables, especially `$HOME`.
 For more detailed implementation notes, see the documentation of the standalone application:
 
 * [conda-standalone](https://github.com/conda/conda-standalone)
+
+
+## Control how `conda` runs on your machine
+
+The traditional installation mechanism for `conda` is for the `constructor`-generated installer to
+run the initialization logic once the files have been copied in the target directory `$INSTDIR`:
```
marcoesters:
```suggestion
run the initialization logic once the files have been copied into the target directory `$INSTDIR`:
```

docs/source/howto.md
```diff
@@ -201,3 +201,58 @@ finding these files may rely on environment variables, especially `$HOME`.
...
+On most shells, `conda init` will modify your shell configuration (`.bashrc` and similar files) to
+inject some shell functions so you can run `conda` from anywhere. That is, the `conda` executable
+you usually run is not the Python entry point, but a shell function that ends up calling that entry
+point. This wrapper is needed by `conda (de)activate` to operate correctly, since it needs to
+modify the currently running shell session. You can find mode details in the [activation deep dive
+guide](https://docs.conda.io/projects/conda/en/stable/dev-guide/deep-dives/activation.html).
+
+The convenience of `conda activate` makes for the extra complexity required at installation time.
+However, some folks might prefer not to modify their shell scripts at all. For those users, an
+alternative initialization method is provided on `conda 25.5.0` and later.
```
marcoesters:
```suggestion
alternative initialization method is provided with `conda 25.5.0` and later.
```

docs/source/howto.md
```diff
@@ -201,3 +201,58 @@ finding these files may rely on environment variables, especially `$HOME`.
...
+```bash
+$INSTDIR/bin/conda init --all
+```
+````
+`````
+
+On most shells, `conda init` will modify your shell configuration (`.bashrc` and similar files) to
+inject some shell functions so you can run `conda` from anywhere. That is, the `conda` executable
+you usually run is not the Python entry point, but a shell function that ends up calling that entry
+point. This wrapper is needed by `conda (de)activate` to operate correctly, since it needs to
```
marcoesters: The sentence starting with "That is, ..." was a little confusing for me to read. Is it needed at all?

docs/source/howto.md
```diff
@@ -201,3 +201,58 @@ finding these files may rely on environment variables, especially `$HOME`.
...
+This new option only adds `$INSTDIR/condabin` to PATH, which is a minimally invasive change to your
+shell configuration and has no runtime overhead. This directory is special because it is guaranteed
+to only contain the `conda` executables and nothing else.
+
+As an installer author, you can control which of these options are made available to the end user:
+
+- `initialize_conda`: the classic initialization logic. On by default.
+- `add_condabin_to_path`: the new, lightweight PATH-only logic. Off by default.
+
+We do NOT recommend enabling both, so pick one! The default value for each option is controlled by
```
marcoesters:
```suggestion
We do NOT recommend enabling both! The default value for each option is controlled by
```

docs/source/howto.md
```diff
@@ -201,3 +201,58 @@ finding these files may rely on environment variables, especially `$HOME`.
...
+`initialize_by_default` and `add_condabin_to_path_default`, respectively.
+
+:::{note}
+The `--condabin` initialization won't be sufficient to run `conda activate`, and `conda` will error
+out saying you need to fully initialize your installation. This might get fixed in the future, but
+for now you can rely on an experimental plugin to use a different activation strategy that doesn't
```
marcoesters:
```suggestion
for now, you can rely on an experimental plugin to use a different activation strategy that doesn't
```

docs/source/howto.md
```diff
@@ -201,3 +201,58 @@ finding these files may rely on environment variables, especially `$HOME`.
...
+require shell modifications: [`conda-spawn`](https://github.com/conda-incubator/conda-spawn). Add
+it to your `specs` definition and will be available in your installations as `conda spawn`.
```
marcoesters:
```suggestion
it to your `specs` definition and it will be available in your installations as `conda spawn`.
```

CONSTRUCT.md
marcoesters: I think consistency with previous UX is to not make it the default because in past UX, that option did not exist. If I updated `constructor` without making changes to the `construct.yaml` file, my installer would behave differently.

jaimergp: An installer built with an updated constructor wouldn't see any changes in behaviour because `add_condabin_to_path` defaults to `false`. This is the default value _once_ `add_condabin_to_path` is `true`.

constructor/header.sh
jaimergp: We still need the flags to opt-in so which ones do you recommend? We can also make it a single flag option like `-c OPTION   Whether to initialize conda in some way: shell, condabin, none`. This would make `initialize_conda` and `add_condabin_to_path` mutually exclusive though. Do you think we should turn `initialize_conda` into an enum type instead of a bool directly?

constructor/shar.py
jaimergp:
```suggestion
    variables["add_condabin_to_path"] = info.get("add_condabin_to_path", False)
```

constructor/winexe.py
jaimergp:
```suggestion
            if info.get("add_condabin_to_path", False) and info.get("_has_conda", False)
```

docs/source/howto.md
jaimergp: I think it's an important distinction to make, but happy to hear alternative wordings to make it clearer.

constructor/nsis/main.nsi.tmpl
jaimergp: I'm following the same pattern as in the `initialize_conda` block, so  I don't think we need it (or we need it in both).

constructor/header.sh
marcoesters: Yes, we need flags to opt in. My concern is more to have two different flags for the opt-in/opt-out behavior. It would be ideal if we used long options, which are more explicit than `-c`. What do you think?

docs/source/howto.md
marcoesters:
```suggestion
inject some shell functions so you can run `conda` from anywhere. That is, the `conda` "executable"
you usually run is not the `conda` entry point, but a shell function that ends up calling the actual entry
point. This wrapper is needed by `conda (de)activate` to operate correctly, since it needs to
```

I think the "Python entry point", while technically correct, is coming a little out of nowhere. Maybe this?

constructor/header.sh
jaimergp: `getopt` doesn't support long options, only single character options. We'd need to change it for something else, but `getopts` is not that cross-platform (subtle differences between Linux/macOS), that's why we're stuck with the bash builtin. I think our only option here is to pick a letter and make it take an argument for the type of initialization required.

docs/source/howto.md
jaimergp: I think the confusion might come from "entry point". Let me change that to "Python script" or similar.

jaimergp: Reworded this section.

constructor/header.sh
marcoesters: Yes, for long options, we would have to use `while/case/shift`, but that's a separate conversation.

constructor/header.sh
```diff
@@ -675,9 +694,66 @@ if [ "${PYTHONPATH:-}" != "" ]; then
     printf "    directories of packages that are compatible with the Python interpreter\\n"
     printf "    in %s: %s\\n" "${INSTALLER_NAME}" "$PREFIX"
 fi
+{% if has_conda %}
+{%- if initialize_conda == 'condabin'%}
```
marcoesters:
```suggestion
{%- if initialize_conda == 'condabin' %}
```

tests/test_header.py
```diff
@@ -85,7 +86,7 @@ def test_osxpkg_scripts_shellcheck(arch, check_path_spaces, script):
 @pytest.mark.parametrize("keep_pkgs", [True])
 @pytest.mark.parametrize("has_conda", [False, True])
 @pytest.mark.parametrize("has_license", [True])
-@pytest.mark.parametrize("initialize_conda", [True])
+@pytest.mark.parametrize("initialize_conda", ["classic", "condabin", False])
```
marcoesters:
```suggestion
@pytest.mark.parametrize("initialize_conda", [True, "classic", "condabin", False])
```

I think we should still test this for backwards compatibility.

tests/test_examples.py
```diff
@@ -960,6 +960,43 @@ def test_virtual_specs_override(tmp_path, request, monkeypatch):
...
+                text=True,
+            )
+        else:  # Windows
+            try:
+                out = subprocess.check_output(
+                    'cmd /V:ON /C "echo !PATH!"',
+                    shell=True,
+                    text=True,
+                )
+                assert str(install_dir / "condabin") in out.split(os.pathsep)
```
```suggestion
                assert str(install_dir / "condabin") in out.strip().split(os.pathsep)
```

tests/test_examples.py
```diff
@@ -960,6 +960,43 @@ def test_virtual_specs_override(tmp_path, request, monkeypatch):
         )
 
 
+@pytest.mark.skipif(not ON_CI, reason="Run on CI only")
+def test_condabin(tmp_path, request, monkeypatch):
+    input_path = _example_path("condabin")
+    for installer, install_dir in create_installer(input_path, tmp_path):
+        if installer.suffix == ".exe":
+            options = ["/AddToPath=1"]
```
marcoesters: I think the test is failing on Windows because you're admin on GitHub runner.

jaimergp: I'm afraid that's true. I'm looking at options to run as a different user but [they are really hacky](https://stackoverflow.com/questions/70904273/can-i-specify-a-user-account-for-a-github-action-step-on-a-windows-runner).

jaimergp: I tried my best as you can see in all the commits underneath, but there seems to be a permissions issue to run `pytest` from the non-admin account. That said, I checked on the VM and:

1. The test doesnt' pass there either
2. BUT the PATH environment var modification can be seen in the Control Panel dialog.
3. The assertion does pass in a new terminal window

I tried several ways to start a fresh process (`subprocess` flags, `os.startfile`...), and no luck.

I did check locally that the PATH is indeed modified, so it works, but it can't be tested on CI :/

marcoesters: So, it looks like you essentially have to restart your shell and environment variables do not propagate into the child processes `pytest` starts. Does checking the registry work on your local system?

If so, I have the following suggestion that actually checks two things at once:

* Check if user is an admin via `ctypes.windll.shell32.IsUserAnAdmin()`
* Force the installation to use `/InstallationType=JustMe`.
* If the user is an admin, mark the test as `xfail`. This will also check whether the CVE that disables adding to path with an admin installation has been re-introduced.
* Parametrize the test to check for both `classic` and `condabin` - I don't think there is a test for that yet.

leofang:
Hello @jaimergp, thanks a ton for working on this, I've subscribed to this PR for some time and glad to see a momentum building up! 

I have a use case for this, but I am not certain the PR as-is would help me once it's merged (mainly because I did not bother reading/understanding the codebase 😛). My use case is
- I get access to a remote Windows machine
- An admin-level account will execute my ansible playbooks to install things I need, including miniforge
- I log in as a user (with admin access too) after playbooks are executed. I expect at this point miniforge's `conda` is in my `PATH` so that I don't have to do anything to run `conda create ...`.

Would this work? Thanks!

jaimergp:
> An admin-level account will execute my ansible playbooks to install things I need, including miniforge

Hey @leofang! Windows admin accounts are not allowed to modify PATH with these installers due to [CVE-2022-26526](https://nvd.nist.gov/vuln/detail/CVE-2022-26526). You can, however, run `/path/to/installation/python.exe -m conda init --condabin` in your ansible playbooks after the installation and it should work.

marcoesters (approved):

PR #972, opened 2025-04-03
Pin `pydantic` to `v2.11.*`

### LINKED ISSUE #971: Schema output depends on `pydantic` version
The schema validation tests currently fail because the schema output depends on the version of `pydantic`. See here for an example run: https://github.com/conda/constructor/actions/runs/14206683147/job/39805869197

In this particular instance, ~`pydantic 3.11`~`pydantic 2.11` added `additionalProperties: true` to objects explicitly, which was not the case for ~`pydantic 3.10`~`pydantic 2.10`. While this is an easy fix, these schemas are published. So, a decision must be made about whether

* the schema is updated and kept up-to-date with the newest version of `pydantic`, or
* the version of `pydantic` is pinned to have a consistent schema output.

marcoesters:
### Description

Pin `pydantic` to `v2.11.*` so that the schema output does not change when a new version of `pydantic` is released.

Closes #971 

### Checklist - did you ...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-schema = ["pydantic ==2.11.*"]
+schema = ["pydantic >=2.11,<2.12"]
 
 [project.scripts]
 constructor = "constructor.main:main"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-schema = ["pydantic >=2"]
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -20,7 +20,7 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-schema = ["pydantic >=2"]
+schema = ["pydantic ==2.11.*"]
```
jaimergp, line 23:
```suggestion
schema = ["pydantic >=2.11,<2.12"]
```

jaimergp (approved):

PR #979, opened 2025-04-21
Fix `hash` keyword for `build_outputs` schema

marcoesters:
### Description

The hash schema contained an incorrect key (`hash_` instead of `hash`), which is fixed in this PR. I also noticed that there are no tests to check that all outputs are generated (which would have caught this), so I added the test here.

### Checklist - did you ...

### DIFF
--- a/constructor/_schema.py
+++ b/constructor/_schema.py
@@ -152,7 +152,7 @@ class HashBuildOutput(BaseModel):
     """
 
     model_config: ConfigDict = _base_config_dict
-    hash_: _HashBuildOutputOptions
+    hash: _HashBuildOutputOptions
 
 
 class InfoJsonBuildOutput(BaseModel):
--- a/constructor/_schema.py
+++ b/constructor/_schema.py
@@ -152,7 +152,7 @@ class HashBuildOutput(BaseModel):
     """
 
     model_config: ConfigDict = _base_config_dict
-    hash: _HashBuildOutputOptions
+    hash_: _HashBuildOutputOptions = Field(..., alias="hash")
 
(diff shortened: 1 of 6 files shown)


constructor/_schema.py
```diff
@@ -152,7 +152,7 @@ class HashBuildOutput(BaseModel):
     """
 
     model_config: ConfigDict = _base_config_dict
-    hash_: _HashBuildOutputOptions
+    hash: _HashBuildOutputOptions
```
jaimergp, line 155: It's better to use the `alias` keyword in `Field` so we don't override the `hash()` builtin in this scope (not that it matters here, but it's a better practice).

jaimergp, line 155:
```suggestion
    hash_: _HashBuildOutputOptions = Field(..., alias="hash")
```

constructor/build_outputs.py
```diff
@@ -41,11 +41,14 @@ def process_build_outputs(info):
...
         outpath = handler(info, **config)
-        logger.info("build_outputs: '%s' created '%s'.", name, outpath)
+        if outpath:
+            logger.info("build_outputs: '%s' created '%s'.", name, outpath)
 
 
 def dump_hash(info, algorithm=None):
-    algorithm = algorithm or []
+    if not algorithm:
+        logger.warning("`hash` requires an algorithm. No has files will be output.")
```
jaimergp, line 50:
```suggestion
        logger.warning("`hash` requires an algorithm. No hash files will be output.")
```

jaimergp (approved):

PR #982, opened 2025-05-01
Provide an option to ovewrite the existing directory by default on Unix

hmaarrfk:
Provides an option for installers to be created with the `-f` option by default.

I would love to provide my users with the ability to do this.

We use constructor to "create a relocatable installer" and we don't expect our users to use for conda, but for the "remaining attached payload".

### Checklist - did you ...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 [//]: # (current developments)
 
 ## 2025-03-20   3.11.3:
-
 ### Bug fixes
 
 * Unset `CONDARC` and `MAMBARC` to prevent pre-existing configuration files from influencing the installation. (#962)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 [//]: # (current developments)
 
 ## 2025-03-20   3.11.3:
+
 ### Bug fixes
 
 * Unset `CONDARC` and `MAMBARC` to prevent pre-existing configuration files from influencing the installation. (#962)
--- a/constructor/shar.py
+++ b/constructor/shar.py
@@ -110,6 +110,11 @@ def get_header(conda_exec, tarball, info):
     variables["min_glibc_version"] = min_glibc_version
 
     variables["script_env_variables"] = info.get("script_env_variables", {})
+    header_template = info.get("header_template", None)
+    if header_template is None:
+        template = read_header_template()
+    else:
+        header_template = str(header_template)
 
     return render_template(read_header_template(), **variables)
 
--- a/constructor/shar.py
+++ b/constructor/shar.py
@@ -110,11 +110,6 @@ def get_header(conda_exec, tarball, info):
     variables["min_glibc_version"] = min_glibc_version
 
     variables["script_env_variables"] = info.get("script_env_variables", {})
-    header_template = info.get("header_template", None)
-    if header_template is None:
-        template = read_header_template()
-    else:
-        header_template = str(header_template)
 
     return render_template(read_header_template(), **variables)
 
--- a/constructor/shar.py
+++ b/constructor/shar.py
@@ -63,6 +63,7 @@ def get_header(conda_exec, tarball, info):
     variables = ns_platform(info["_platform"])
     variables["keep_pkgs"] = bool(info.get("keep_pkgs", False))
     variables["batch_mode"] = bool(info.get("batch_mode", False))
+    variables["force_by_default"] = bool(info.get("force_by_default", False))
     variables["has_license"] = has_license
     if variables["batch_mode"] and has_license:
         raise Exception(
--- a/docs/source/construct-yaml.md
+++ b/docs/source/construct-yaml.md
@@ -239,6 +239,14 @@ Only affects `.sh` installers. If `False`, the installer launches
 an interactive wizard guiding the user through the available options. If
 `True`, the installer runs automatically as if `-b` was passed.
 
+### `force_by_default`
+
+Only affects `.sh` installers. If `True`, the installer will remove the
+existing installation without prompting the user. If set, the user can
+disable this option with the `-e` option. If `False`, the installer will
+stop if the installation path already exists. The user can override this
+option by passing `-f` to the installer.
+
 ### `signing_identity_name`
 
 By default, the MacOS pkg installer isn't signed. If an identity name is specified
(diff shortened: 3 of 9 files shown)


docs/source/construct-yaml.md
```diff
@@ -239,6 +239,14 @@ Only affects `.sh` installers. If `False`, the installer launches
 an interactive wizard guiding the user through the available options. If
 `True`, the installer runs automatically as if `-b` was passed.
 
+### `force_by_default`
+
+Only affects `.sh` installers. If `True`, the installer will remove the
+existing installation without prompting the user. If set, the user can
```
marcoesters, line 245: > the installer will remove the existing installation without prompting the user.

The installer does not perform any removal operations. It will just clobber over the existing installations. So, the final installation directory contains a mix of new and old installations. See also https://github.com/conda/constructor/issues/679.

marcoesters (changes requested):
Apart from that some status checks need to pass still and my other comments, I'm not sure I understand the use case of this.

The way I understand the PR description, you have installers that just have a payload without using `conda` (or `mamba`) to install packages, is that correct? For those installers, you just want to extract the payload into a directory by default? This would overwrite existing files with the same name and leaving old files behind, leaving a mix of old an new installation.

In my opinion, the `-f` flag is dangerous because the `conda` installation clobbers over the existing environments without any checks. So, I'm not sure if providing this option unchecked by default is a safe thing to do.

My concern is that we are expanding on a feature here that I think is more broken than functional. So, I would like to understand more about the use case and why users can't run the `-f` option instead.

docs/source/construct-yaml.md
hmaarrfk: ah this is a bug. yeah, i thought that `-f` would remove, then overwrite.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 [//]: # (current developments)
 
 ## 2025-03-20   3.11.3:
+
```
hmaarrfk, line 4:
```suggestion
```

constructor/shar.py
```diff
@@ -109,6 +110,11 @@ def get_header(conda_exec, tarball, info):
     variables["min_glibc_version"] = min_glibc_version
 
     variables["script_env_variables"] = info.get("script_env_variables", {})
+    header_template = info.get("header_template", None)
+    if header_template is None:
+        template = read_header_template()
+    else:
+        header_template = str(header_template)
```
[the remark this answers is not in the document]
hmaarrfk: sorry, longer answer below.

docs/source/construct-yaml.md
hmaarrfk: a bug in my understanding.

hmaarrfk:
> n, you have installers that just have a payload without using conda (or mamba) to install packages, is that correct?

We will typically bundle in conda/mamba into the installer but they aren't for the user to use to create virtual environment. They are to "self update". But self updates can be messy with large dependency changes.

> For those installers, you just want to extract the payload into a directory by default? 

I was hoping to first `rm -rf` then to install there yes.

> In my opinion, the -f flag is dangerous because the conda installation clobbers over the existing environments without any checks.

I think I misunderstood. I thought they would `rm -rf` first.

I understand the dangers. These are machines that "we manage" and I think there is little difference between 

```
./use_our_installer.sh -f
```
and
```
`./use_our_installer.sh
```

if we tell people to "blindly follow instructions". In both cases they would "force" the install.

----------------------------------------------------------

As to why there was some `template_header` code, as I was going through this, there were some other modifications on my wish list. I was going to suggest after seeing what you thought of this, to simply have an option to specify a new `header.sh` template that we would control outside of constructor.

I have the skills to fork it, but I generally think that a bit of customization on linux side would be nice. Perhaps even to add that `rf -rf ${HOME}/user_prefix` to help with the install process.


--------------------------------------------------------------

In an othre world, I would have a minimal installer, 50-200MB, that would then "detect the platform, hardware specifics", and install the real optimized payload, perhaps 2-3GB large.

marcoesters:
Thank you for the detailed answer! I think there are several things here that all deserve separate issues, but I will address them here.

> I was hoping to first rm -rf then to install there yes.

So, you understood `-f` to be a "force re-install" option? That's a very reasonable assumption and much cleaner than what we have now, in my opinion. It would be a pretty big UX change, but I'm not sure how often that flag is used or how many people have the same understanding you had. That's a separate PR though.

> These are machines that "we manage" and I think there is little difference between
> 
> ` ./use_our_installer.sh -f`
>
> and
>
> `./use_our_installer.sh`
>
> if we tell people to "blindly follow instructions". In both cases they would "force" the install.

That's exactly what I'm wondering, if this is that much of a UX improvement that we want to maintain the extra logic with the `-e` flag? This particular feature doesn't look too bad, but I'm wondering if this is definitely a one-off or if we would need to support similar behavior with other (or new) flags, too, and/or support the same ability to customize with PKG and EXE installers.

I'd be curious about @jaimergp's opinion here, too.

The `-e` option will have to be implemented in this PR, including integration tests.

> As to why there was some template_header code, as I was going through this, there were some other modifications on my wish list. I was going to suggest after seeing what you thought of this, to simply have an option to specify a new header.sh template that we would control outside of constructor.

We have that for Windows already, so I don't see a reason to reject this for SH installers. That would be a separate PR though. I'd really like to see that wish list, so if you could create issues for each item, that would be amazing!

> In an othre world, I would have a minimal installer, 50-200MB, that would then "detect the platform, hardware specifics", and install the real optimized payload, perhaps 2-3GB large.

We could get partially there by allowing online installations, but that's a separate issue. Could even be environment-specific. In principle, this would only require removing the `--offline` flag when installing packages.

hmaarrfk:
ok i'll try to gather my thoughts.

for the "force" flag, i'm happy to implement it with my "understanding", but do you have an idea for the "flag", would it "-g"????? or "-o"???? let me know.

marcoesters:
> ok i'll try to gather my thoughts.
> 
> for the "force" flag, i'm happy to implement it with my "understanding", but do you have an idea for the "flag", would it "-g"????? or "-o"???? let me know.

Let's discuss this in an issue. I do believe that what you describe is actually what `-f` **should** do (even though it doesn't right now), so we don't need to find a new flag. It just needs to be implemented in a way that's backwards compatible.

PR #991, opened 2025-05-22
EXE installers: inherit permissions from `$INSTDIR` for all-users installations

### LINKED ISSUE #828: Struggling with permissions issues for the AllUsers settings on Windows
### Checklist

### What happened?

We build a constructor Windows installer annually and install for `AllUsers` in silent mode. This year many/most of the installed scripts in `<install-directory>/Scripts/` have permissions that prevent execution. For example, opening the Anaconda prompt and typing `Jupyter lab` gives `Error executing Jupyter command 'lab': [WinError 5] Access is denied`.   Similar errors are produced when trying to launch lab from the Anaconda Navigator. Running `spyder` from the prompt or Navigator works. We tried building the new installer with the constructor version (3.4.5) we used last year but still get the permission errors. We have many details of our trial and error in this issue: https://gitlab.ewi.tudelft.nl/bhmgerritsen/anaconda-dee-config/-/issues/64 but I'll summarize below.

Constructor setup:

- Channel: conda-forge (with anaconda-navigator, console_shortcut, menuinst, powershell_shortcut using `defaults::` notation and with conda channel priority set to "strict". The idea was to install everything form conda-forge except for these few anaconda specific packages.
- We install with silent mode like `tudelft-anaconda-v2024.01-Windows-x86_64.exe /InstallationType=AllUsers /RegisterPython=1 /S /D=C:\Programs\tudelft-anaconda` (but also have tried graphical installs)
- `construct.yaml` is here: https://gitlab.ewi.tudelft.nl/bhmgerritsen/anaconda-dee-config/-/blob/master/construct.yaml
- Our latest installer is here: https://gitlab.ewi.tudelft.nl/bhmgerritsen/anaconda-dee-config/-/releases

If we download the latest official Anaconda windows installer and run it in silent mode the permissions are correct. But our installer leaves many files with restricted permissions.

Changes from last year:

- The approach was have channel priority with `defaults` then `conda-forge` and install the `anaconda` package and then add various extra packages on top of anaconda from conda-forge.
- We set the `channel_priority=strict` this year and didn't last year.
- We hard pin the versions of all packages this year.
- Our constructor files from last year's release are here: https://gitlab.ewi.tudelft.nl/bhmgerritsen/anaconda-dee-config/-/tree/62d723fbbfabe495a0bb38c408b51a40dd1a8adc

We have run out of ideas as to what causes the permissions errors. Do you have any suggestions for what we may be doing wrong?

### Conda Info

```shell
NA
```

### Conda Config

```shell
NA
```

### Conda list

```shell
NA
```

### Additional Context

_No response_

marcoesters:
### Description

For all-user installations, the permissions of the installation directory are changed manually to remove write access for authenticated, domain, and built-in users, and to add read and execute rights for domain and built-in users. This assumes that permissions applied to the installation directories are inherited by all files and subdirectories.

If a file has the `SDDL_PROTECTED` flag set (see [Security Descriptors](https://learn.microsoft.com/en-us/windows/win32/secauthz/security-descriptor-string-format)), these permissions are no inherited and thus not changed. The result is that users who are not part of the Administrators group(s) cannot use these files.

One example where this flag is set is the exe entry points created by `_conda.exe`: `conda-standalone` creates a hard link from the entry point in the extracted temporary directory to the installation directory. Windows automatically sets the `SDDL_PROTECTED` flag in that case and the permissions are not changed at the end of the installation.

The solution is to enable inheritance for all files and directories inside `$INSTDIR` while still protecting `$INSTDIR` itself. This is done by a call to `icacls` because it is much faster than calling `AccessControl::EnableFileInheritance` on every file (local tests with only `jupyter` show a difference of 30 seconds).

The added tests fail for the Miniforge example without this addition, as expected.

Closes #828.

### Checklist - did you ...


tests/test_examples.py
```diff
@@ -929,6 +960,144 @@ def test_virtual_specs_override(tmp_path, request, monkeypatch):
         )
 
 
+@pytest.mark.skipif(not ON_CI, reason="CI only")
+@pytest.mark.skipif(not sys.platform.startswith("win"), reason="Windows only")
+def test_allusers_exe(tmp_path, request):
```
jaimergp, line 965: This is a super useful test that nonetheless might feel foreign when compared to the other tests in this module. Can you add a docstring explaining what we are testing here and why? It's obvious now in the context of this PR, but it might not be in the future when this all of a sudden fails in the future for whatever reason.

jaimergp (dismissed):

tests/test_examples.py
marcoesters: That's a good call. I added a detailed doc string to the function that also talks about the limitations of this test. It only tests that permissions are correctly applied. A more complete test would go through all permissions attached to each file and ensures that write permissions are unset for users who are not part of the local administrator group. However, this isn't part of the installer code either.

tests/test_examples.py
```diff
@@ -929,6 +960,158 @@ def test_virtual_specs_override(tmp_path, request, monkeypatch):
...
+    """Ensure that AllUsers installations have the correct permissions for built-in users,
+    domain users, and authenticated users.
+
+    For AllUsers installations, these users should only have read and execute permissions,
+    but no write permissions. Addtionally, the installation directory must not inherit from
```
jaimergp:
```suggestion
    but no write permissions. Additionally, the installation directory must not inherit from
```

jaimergp (approved):

PR #993, opened 2025-06-02
Support conda-standalone `onedir` variant

### LINKED ISSUE #996: Constructor tests failed
The Build workflow failed on 2025-06-09 21:35 UTC

Full run: https://github.com/conda/constructor/actions/runs/15544426937

(This post will be updated if another test fails, as long as this issue remains open.)

jaimergp:
### Description

The conda-standalone `onedir` variant ships an executable that loads libraries from an adjacent `_internal` directory. These files need to be shipped in the installer bundle too. Since we can't use conda.exe to extract the extra files (it needs them to run) and we don't want to rely on the system's `tar` (I found no reason not to in the git history, but let's stick to this convention because there must have been a good reason), we need to unpack them with the same `dd` technique _per-file_. I'm using Jinja to generate those `extract_range` commands. Not pretty but it does the job. Note that this means the `_internal` directory is shipped as is, uncompressed.

### Checklist - did you ...

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -428,6 +428,10 @@ extract_range () {
     blk_siz=16384
     dd1_beg=$1
     dd3_end=$2
+    range_size=$(( dd3_end - dd1_beg ))
+    if [ $blk_siz -gt $range_size ]; then
+        blk_siz=$range_size
+    fi
     dd1_end=$(( ( dd1_beg / blk_siz + 1 ) * blk_siz ))
     dd1_cnt=$(( dd1_end - dd1_beg ))
     dd2_end=$(( dd3_end / blk_siz ))
@@ -475,7 +479,11 @@ extract_range "${boundary0}" "${boundary1}" > "$CONDA_EXEC"
 chmod +x "$CONDA_EXEC"
 {%- for filename, (start, end) in conda_exe_payloads|items %}
 mkdir -p "$(dirname "$PREFIX/{{ filename }}")"
-extract_range (( boundary1 + {{ start }} )) (( boundary1 + {{ end }} ))  > "$PREFIX/{{ filename }}"
+{%- if start == end %}
+touch "$PREFIX/{{ filename }}"
+{%- else %}
+extract_range $(( boundary1 + {{ start }} )) $(( boundary1 + {{ end }} ))  > "$PREFIX/{{ filename }}"
+{%- endif %}
 {%- endfor %}
 
 export TMP_BACKUP="${TMP:-}"
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -442,15 +442,20 @@ extract_range () {
 
 # the line marking the end of the shell header and the beginning of the payload
 last_line=$(grep -anm 1 '^@@END_HEADER@@' "$THIS_PATH" | sed 's/:.*//')
+# first payload: conda.exe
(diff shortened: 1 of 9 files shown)


jaimergp:
@marcoesters, ready for review!

constructor/header.sh
```diff
@@ -494,8 +494,14 @@ fi
 
 # the second binary payload: the tarball of packages
 printf "Unpacking payload ...\n"
+{%- if bootstrap_with_tar %}
+{#- we only use tar for conda-standalone onedir builds, because otherwise conda cannot extract itself #}
+extract_range "${boundary1}" "${boundary2}" | \
```
marcoesters: Since we know which internal files we have via `copy_conda_exe`, could we extract them the same way we extract `_conda`? For example, via a Jinja-templated for-loop? That would avoid bootstrapping with `tar`.

jaimergp: That's going to be a LOT of bookkeeping and boundary annotation. Isn't `tar` expected to be present in every *nix machine anyway? I dove in the git history but this code is there from the beginning (>12 years) so I wonder if those design constraints to not rely on tar still apply today.

jaimergp: I went for it, let me know what you think. The PR description have been updated accordingly.

jaimergp:
`conda-standalone-nightly` jobs expected to fail until https://github.com/conda/conda-standalone/pull/146 lands.

jaimergp:
@marcoesters, do you recall if we need anything special for uninstallation? We are going to leave a `_internal` directory behind if we are naive about this... Do we need to copy "ourselves" to a temporary directory just like the single-binary does?

constructor/header.sh
marcoesters: > I dove in the git history but this code is there from the beginning (>12 years) so I wonder if those design constraints to not rely on tar still apply today.

That is a good question. It is plausible that this feature was reused because of Windows. In either case, there is no reason for the if condition. If the design decisions still hold, we shouldn't use `tar`. If they don't, we can just use `tar` in both cases.

I do see the advantage of using the `conda-standalone` process because it can apply data filters.

marcoesters:
> @marcoesters, do you recall if we need anything special for uninstallation? We are going to leave a `_internal` directory behind if we are naive about this... Do we need to copy "ourselves" to a temporary directory just like the single-binary does?

`conda-standalone` does have a test for self-deletion: https://github.com/conda/conda-standalone/blob/8d33502ec0c2ad42f793d18ed453189d0d2e9932/tests/test_uninstall.py#L456-L483

This will leave `.conda_trash` files behind on Windows. This is mitigated by a call to `RMDIR` in the uninstaller: https://github.com/conda/constructor/blob/62534b40861ecbc1cc5fce979fa8e7b70edda236/constructor/nsis/main.nsi.tmpl#L1612-L1616

The uninstaller is also tested here where `_run_uninstaller_exe` checks the number of files left behind: https://github.com/conda/constructor/blob/62534b40861ecbc1cc5fce979fa8e7b70edda236/tests/test_examples.py#L982-L1057

jaimergp:
I see, that means that we should be ok, then?

Otherwise, I don't see where `_run_uninstaller_exe` is counting the amount of files left, though. Should I add checks for `_internal/*` there?

marcoesters (approved):

PR #1014, opened 2025-07-17
Prepare 3.12.0

jaimergp:
✂️✂️✂️ Snip, snip... new release! ✂️✂️✂️


jaimergp:
cc @Jrice1317

AUTHORS.md
```diff
@@ -67,5 +68,8 @@ Authors are sorted alphabetically.
 * XuehaiPan
 * astaric
 * bkreider
+* conda-bot
+* dependabot[bot]
```
Jrice1317, line 72:
```suggestion
```
We should remove the bots, correct?

AUTHORS.md
```diff
@@ -67,5 +68,8 @@ Authors are sorted alphabetically.
...
 * guimondmm
+* pre-commit-ci[bot]
```
Jrice1317, line 74:
```suggestion
```

Jrice1317 (approved):
Outside of the bots, LGTM! Thank you for including me! I'll start the package update as soon as it's released.

AUTHORS.md
jaimergp: They'll keep popping up so tbh I don't really mind them here, I only remove them from the changelog, and not always.

The cool thing would be to have an `exclude` option in `rever.xsh` but I didn't look into it.

marcoesters (approved):

PR #1021, opened 2025-07-25
Add examples to test source files

marcoesters:
### Description

Canary builds are failing because the examples are missing in the source files for tests.

### Checklist - did you ...

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -35,7 +35,7 @@ requirements:
 
 test:
   source_files:
-    - examples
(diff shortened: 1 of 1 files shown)


recipe/meta.yaml
```diff
@@ -35,6 +35,7 @@ requirements:
 
 test:
   source_files:
+    - examples
```
jaimergp:
```suggestion
    - examples/miniforge
```

I think we only need this one.

PR #1058, opened 2025-09-09
Protected envs

Jrice1317:
### Checklist - did you ...

### DIFF
--- a/constructor/main.py
+++ b/constructor/main.py
@@ -122,6 +122,12 @@ def main_build(
     ):
         sys.exit("Error: uninstalling with conda.exe requires conda-standalone 24.11.0 or newer.")
 
+    if (info.get("CONDA_PROTECT_FROZEN_ENVS")
+        and exe_type == StandaloneExe.CONDA
+        and exe_version
+        and exe_version <= Version("25.5.1")):
+        sys.exit("Error: installing with protected base environment requires conda-standalone newer than 25.5.1.")
+
     logger.debug("conda packages download: %s", info["_download_dir"])
 
     for key in ("welcome_image_text", "header_image_text"):
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -1296,11 +1296,12 @@ Section "Install"
 {%- endfor %}
     System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_SAFETY_CHECKS", "disabled").r0'
     System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_EXTRA_SAFETY_CHECKS", "no").r0'
-    System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_ROOT_PREFIX", "$INSTDIR")".r0'
-    System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_PKGS_DIRS", "$INSTDIR\pkgs")".r0'
+    System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_ROOT_PREFIX", "$INSTDIR").r0'
+    System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_PKGS_DIRS", "$INSTDIR\pkgs").r0'
+    System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_PROTECT_FROZEN_ENVS", "0").r0'
     # Spinners in conda write a new character with each movement of the spinner.
     # For long installation times, this may cause a buffer overflow, crashing the installer.
-    System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_QUIET", "1")".r0'
+    System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_QUIET", "1").r0'
     # Extra info for pre and post install scripts
     # NOTE: If more vars are added, make sure to update the examples/scripts tests too
     #       There's a similar block for the pre_uninstall script, further down this file.
--- a/examples/protected_base/conda-meta/frozen
+++ b/examples/protected_base/conda-meta/frozen
@@ -1,0 +1,1 @@
+# This environment is protected.
--- a/examples/protected_base/construct.yaml
+++ b/examples/protected_base/construct.yaml
@@ -1,0 +1,22 @@
+# yaml-language-server: $schema=../../constructor/data/construct.schema.json
+"$schema": "../../constructor/data/construct.schema.json"
+
+name: ProtectedBaseEnv
+version: X
+installer_type: all
+
+channels:
+  - defaults
+
+specs:
+  - python
+  - conda
+
+extra_envs:
+  default:
+    specs:
+      - python
+      - pip
+
+extra_files:
+  - frozen.json: conda-meta/frozen
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -1370,3 +1370,18 @@ def test_regressions(tmp_path, request):
             check_subprocess=True,
             uninstall=True,
         )
+
+def test_frozen_environment(tmp_path, request):
+    input_path = _example_path("protected_base")
+    for installer, install_dir in create_installer(input_path, tmp_path):
+        frozen_file = install_dir / "conda-meta" / "frozen"
+        assert frozen_file.exists()
+        _run_installer(
+            input_path,
+            installer,
+            install_dir,
+            request=request,
+            check_subprocess=True,
+            uninstall=True
+        )
+        assert frozen_file.exists()
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -591,6 +591,7 @@ shortcuts=""
 
 {%- set channels = final_channels|join(",") %}
 # shellcheck disable=SC2086
+CONDA_PROTECT_FROZEN_ENVS="0" \
 CONDA_ROOT_PREFIX="$PREFIX" \
 CONDA_REGISTER_ENVS="{{ register_envs }}" \
 CONDA_SAFETY_CHECKS=disabled \
@@ -631,6 +632,7 @@ for env_pkgs in "${PREFIX}"/pkgs/envs/*/; do
     env_shortcuts=""
     {%- endif %}
     # shellcheck disable=SC2086
+    CONDA_PROTECT_FROZEN_ENVS="0" \
     CONDA_ROOT_PREFIX="$PREFIX" \
     CONDA_REGISTER_ENVS="{{ register_envs }}" \
     CONDA_SAFETY_CHECKS=disabled \
--- a/constructor/osx/run_installation.sh
+++ b/constructor/osx/run_installation.sh
@@ -47,6 +47,7 @@ fi
 notify "Installing packages. This might take a few minutes."
 # shellcheck disable=SC2086
 if ! \
+CONDA_PROTECT_FROZEN_ENVS="0" \
 CONDA_REGISTER_ENVS="{{ register_envs }}" \
 CONDA_ROOT_PREFIX="$PREFIX" \
 CONDA_SAFETY_CHECKS=disabled \
@@ -93,6 +94,7 @@ for env_pkgs in "${PREFIX}"/pkgs/envs/*/; do
     fi
 
     # shellcheck disable=SC2086
+    CONDA_PROTECT_FROZEN_ENVS="0" \
     CONDA_ROOT_PREFIX="$PREFIX" \
     CONDA_REGISTER_ENVS="{{ register_envs }}" \
     CONDA_SAFETY_CHECKS=disabled \


examples/protected_base/conda-meta/frozen
```diff
@@ -0,0 +1 @@
+# This environment is protected.
```
jaimergp:
```suggestion
{"message": "This environment is protected."}

```

This file must be JSON or totally empty.

examples/protected_base/construct.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - conda
+
+extra_envs:
+  default:
+    specs:
+      - python
+      - pip
+
+extra_files:
+  - frozen.json: conda-meta/frozen
```
jaimergp:
```suggestion
  - conda-meta/frozen: conda-meta/frozen
```

I think you mean this.

Jrice1317: Yes, thank you!

Jrice1317: Well, maybe...I'm not entirely sure how this mapping works for the extra_files.

constructor/nsis/main.nsi.tmpl
```diff
@@ -1296,11 +1296,12 @@ Section "Install"
...
     System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_SAFETY_CHECKS", "disabled").r0'
     System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_EXTRA_SAFETY_CHECKS", "no").r0'
-    System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_ROOT_PREFIX", "$INSTDIR")".r0'
-    System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_PKGS_DIRS", "$INSTDIR\pkgs")".r0'
+    System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_ROOT_PREFIX", "$INSTDIR").r0'
+    System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_PKGS_DIRS", "$INSTDIR\pkgs").r0'
+    System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_PROTECT_FROZEN_ENVS", "0").r0'
     # Spinners in conda write a new character with each movement of the spinner.
     # For long installation times, this may cause a buffer overflow, crashing the installer.
-    System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_QUIET", "1")".r0'
```
jaimergp, line 1303: Intriguing that this typo didn't cause a problem so far...

constructor/main.py
```diff
@@ -177,6 +180,27 @@ def main_build(
                     new_extras.append({orig: dest})
         info[extra_type] = new_extras
 
+    def has_frozen_file(extra_files: list[str | dict[str, str]]) -> bool:
+        def check_path(path_str: str) -> bool:
+            path = Path(path_str)
+            return path.name == "frozen" and path.parent.name == "conda-meta"
```
jaimergp:
```suggestion
            return path.parts[-2:] == ("conda-meta", "frozen")
```

Tidier?

jaimergp: This however will give a false positive when someone passes `not/an/env/conda-meta/frozen`. So maybe we should check whether it's exactly `conda-meta/frozen` or `envs/<some-env-name>/conda-meta/frozen`:

```suggestion
            return path.parts == ("conda-meta", "frozen") or (len(path.parts) == 4 and path.parts[0] == "envs" and path.parts[-2:] == ("conda-meta", "frozen"))
```

constructor/main.py
```diff
@@ -177,6 +180,27 @@ def main_build(
                     new_extras.append({orig: dest})
         info[extra_type] = new_extras
 
+    def has_frozen_file(extra_files: list[str | dict[str, str]]) -> bool:
+        def check_path(path_str: str) -> bool:
```
jaimergp:
```suggestion
        def is_conda_meta_frozen(path_str: str) -> bool:
```

constructor/main.py
```diff
@@ -177,6 +180,27 @@ def main_build(
...
+            if isinstance(file, str) and check_path(file):
+                return True
+            elif isinstance(file, dict) and any(check_path(val) for val in file.values()):
+                return True
+        return False
+
+    if (
+        has_frozen_file(info.get("extra_files", []))
+        and exe_type == StandaloneExe.CONDA
+        and check_version(exe_version, min_version="25.5.0", max_version="25.7.0")
```
jaimergp, line 187:
```suggestion
        and not check_version(exe_version, min_version="25.7.0")
```

I think?

constructor/main.py
```diff
@@ -177,6 +180,27 @@ def main_build(
...
+    ):
+        sys.exit(
+            "Error: installing with protected base environment requires conda-standalone newer than 25.5.x"
```
jaimergp:
```suggestion
            "Error: handling conda-meta/frozen marker files requires conda-standalone newer than 25.7.x"
```

constructor/utils.py
```diff
@@ -25,6 +25,13 @@
 
 from ruamel.yaml import YAML
 
+try:
+    from conda.models.version import VersionOrder
+
+    has_conda_interface = True
+except ImportError:
+    has_conda_interface = False
```
jaimergp: Why would this fail? It's a dependency.

```suggestion
from conda.models.version import VersionOrder
```

constructor/utils.py
```diff
@@ -343,6 +350,26 @@ def identify_conda_exe(conda_exe: str | Path | None = None) -> tuple[StandaloneE
...
+def check_version(
+    exe_version: str | VersionOrder | None = None,
+    min_version: str | None = None,
+    max_version: str | None = None,
+) -> bool:
+    """Check if a version is within a version range.
+
+    The minimum version is assumed to be inclusive, the maximum version is not inclusive.
+    """
+    if not exe_version or not has_conda_interface:
```
jaimergp:
```suggestion
    if not exe_version:
```

constructor/utils.py
marcoesters: I wrote that helper function and I saw this part: https://github.com/conda/constructor/blob/d992bd7fbda7958c2d1d1d279a13727d555d7c87/constructor/conda_interface.py#L39-L46

Now that I look at the `git blame`, this was written eight years ago and is probably obsolete, so I agree that the `try` block can be removed.

tests/test_examples.py
```diff
@@ -1380,3 +1378,26 @@ def test_regressions(tmp_path, request):
...
             uninstall=True,
         )
+
+
+@pytest.mark.xfail(
+    condition=(
+        CONDA_EXE == StandaloneExe.CONDA
+        and check_version(CONDA_EXE_VERSION, min_version="25.5.0", max_version="25.7.0")
+    ),
+    reason="conda-standalone 25.5.x fails with protected base environments and older versions are ignored",
```
```suggestion
    reason="conda-standalone 25.5.x fails with protected environments",
```

I don't think that last aside is necessary since the tests will run and pass for older versions.

examples/protected_base/construct.yaml
marcoesters, line 23:
```suggestion
extra_files:
  - frozen.json: conda-meta/frozen
  - frozen.json: envs/defaults/conda-meta/frozen
```

Since the extra environments are handled in separate code blocks than the `base` environment for SH and PKG, I think the `default` environment needs to be frozen, too, to get more test coverage. `test_frozen_environment` needs to be adjusted to test for that `frozen` file, too, in this case.

constructor/main.py
```diff
@@ -177,6 +180,27 @@ def main_build(
                     new_extras.append({orig: dest})
         info[extra_type] = new_extras
 
+    def has_frozen_file(extra_files: list[str | dict[str, str]]) -> bool:
+        def check_path(path_str: str) -> bool:
+            path = Path(path_str)
+            return path.name == "frozen" and path.parent.name == "conda-meta"
+
+        for file in extra_files:
+            if isinstance(file, str) and check_path(file):
```
Jrice1317:
```suggestion
            if isinstance(file, str) and is_conda_meta_frozen(file):
```

constructor/main.py
```diff
@@ -177,6 +180,27 @@ def main_build(
...
+                return True
+            elif isinstance(file, dict) and any(check_path(val) for val in file.values()):
```
Jrice1317:
```suggestion
            elif isinstance(file, dict) and any(is_conda_meta_frozen(val) for val in file.values()):
```

constructor/main.py
Jrice1317: I originally went this route but was told to only flag 25.5.x and not flag any older versions, though this makes the most sense to me unless there is something else I'm missing. CC: @marcoesters for clarity

constructor/nsis/main.nsi.tmpl
Jrice1317: Agreed. I almost left it because I thought it was intentional, but I figured if it was, someone would let me know here. Lol

marcoesters: I'm afraid I'm just as puzzled as everyone else here. 😂

constructor/main.py
marcoesters: The code by @Jrice1317 is correct here. `conda-standalone 25.5.x` does not properly handle the `CONDA_PROTECT_FROZEN_ENVS` environment variable because `conda 25.5.x` doesn't. For `25.7.x`, we used a patch to correct that behavior: https://github.com/conda/conda-standalone/commit/913c8f408a1341fc9de2f99e9fcfa76abb01611a

Versions before `25.5.0` do not have have CEP-22 implemented. So, `conda-meta/frozen` would just be ignored and the installation succeeds.

The reason why we have to prohibit `25.5.x` is that the installation would crash for EXE and PKG installers.

constructor/main.py
```diff
@@ -177,6 +180,31 @@ def main_build(
                     new_extras.append({orig: dest})
         info[extra_type] = new_extras
 
+    def has_frozen_file(extra_files: list[str | dict[str, str]]) -> bool:
+        def is_conda_meta_frozen(path_str: str) -> bool:
+            path = Path(path_str)
+            return path.parts == ("conda-meta", "frozen") or (
+                len(path.parts) == 4
+                and path.parts[0] == "envs"
```
marcoesters, line 173:
```suggestion
                len(path.parts) >= 4
                and path.parts[-3] == "envs"
```

I don't think the original works the way it's written for strings. In the loop above, all strings are converted into absolute paths, so `envs/default/conda-meta/frozen` will not be caught (only inside a mapping).

We can either change it to the suggestion here (but then miss `not/an/env/envs/<some name>/conda-meta/frozen`); or the code in its current form must be moved before this for-loop: https://github.com/conda/constructor/blob/d992bd7fbda7958c2d1d1d279a13727d555d7c87/constructor/main.py#L165-L178

I suggest the latter. @jaimergp , thoughts?

tests/test_examples.py
```diff
@@ -1380,3 +1378,39 @@ def test_regressions(tmp_path, request):
...
+        actual_frozen_paths = set()
+        for env in install_dir.glob("**/conda-meta/history"):
+            frozen_file = env.parent / "frozen"
+            assert frozen_file.exists()
+            actual_frozen_paths.add(frozen_file)
+
+        assert expected_frozen_paths == actual_frozen_paths, (
+            f"Expected: {sorted(str(p) for p in expected_frozen_paths)}\n"
+            f"Found: {sorted(str(p) for p in actual_frozen_paths)}"
+        )
```
```suggestion
        assert expected_frozen_paths == actual_frozen_paths
```

I think the way you set up the `assert` statement, you will already get a very nice diff that tells you just as much as the custom error message.

constructor/nsis/main.nsi.tmpl
jaimergp: Shall we do this in a separate PR in case we need to revert for some weird reason? Although it's only a couple quotes so maybe it doesn't matter.

constructor/main.py
jaimergp: Sure, but if someone is setting a frozen file via extra files _and_ using conda-standalone<25.5 we won't warn them that this is not going to work? I'd rather get an error so folks upgrade their conda-standalone or remove the extra files entry.

constructor/main.py
jaimergp: This will be clearer with once we merge https://github.com/conda/constructor/pull/1063. For now let's go with your suggestion (move before for-loop), @marcoesters.

constructor/main.py
marcoesters: Whether protected environments work do not depend on the `conda-standalone` version, but the `conda` version in the `base` environment. For example, if you use `conda-standalone < 25.5` and package `conda 25.7.0`, protecting environments will work. So, I don't think an error would be appropriate here.

constructor/nsis/main.nsi.tmpl
marcoesters: My guess is that this is not a syntax error because it's within single quotes. I think if changing this was a problem, we'd have seen it.

constructor/main.py
jaimergp: Ok, sorry I mixed the two. We can close this thread then.

tests/test_examples.py
```diff
@@ -1380,3 +1378,39 @@ def test_regressions(tmp_path, request):
...
+        expected_frozen_paths = {
+            install_dir / "conda-meta" / "frozen",
+            install_dir / "envs" / "default" / "conda-meta" / "frozen",
+        }
+
+        actual_frozen_paths = set()
+        for env in install_dir.glob("**/conda-meta/history"):
+            frozen_file = env.parent / "frozen"
+            assert frozen_file.exists()
+            actual_frozen_paths.add(frozen_file)
```
```suggestion
            if frozen_file.exists():
                actual_frozen_paths.add(frozen_file)
```

tests/test_examples.py
```diff
@@ -1380,3 +1378,36 @@ def test_regressions(tmp_path, request):
...
+        expected_frozen_paths = {
+            install_dir / "conda-meta" / "frozen",
+            install_dir / "envs" / "default" / "conda-meta" / "frozen",
+        }
+
+        actual_frozen_paths = set()
+        for env in install_dir.glob("**/conda-meta/history"):
+            frozen_file = env.parent / "frozen"
+            if frozen_file.exists():
+                actual_frozen_paths.add(frozen_file)
```
marcoesters, line 1401:
```suggestion
        actual_frozen_paths = {frozen_file for frozen_file in expected_frozen_paths if frozen_file.exists()}
```

marcoesters:
pre-commit.ci autofix

tests/test_examples.py
```diff
@@ -317,6 +317,8 @@ def _run_installer(
             check=check_subprocess,
             options=options,
         )
+        if request and ON_CI:
+            request.addfinalizer(lambda: shutil.rmtree(str(install_dir), ignore_errors=True))
```
jaimergp, line 322: Should we do this for all installers? Regardless, let's add a comment that we are doing this to free space and prevent "running out of storage" errors.

jaimergp (dismissed):

tests/test_examples.py
marcoesters: Since we uninstall by default on Windows, I'm not worried about EXE installers. PKG installers already have that finalizer. I did add a comment though.

PR #1059, opened 2025-09-11
Move `pkgs/env.txt` lockfile to `conda-meta/initial-state.explicit.txt` to preserve it

### LINKED ISSUE #1052: Transform envs.txt into persistent explicit lockfile
### Checklist

#### Why is this needed?
The lockfile needs to be shipped inside the installer and remain accessible after installation. Currently, envs.txt contains the @EXPLICIT directive with exact package specifications but gets deleted during the installation process.
#### What should happen?
Modify preconda.py to preserve the existing envs.txt file by transforming it into a persistent explicit lockfile:

- Rename/copy envs.txt to conda-meta/initial-state.explicit.txt during the installer process (before it gets deleted)
- Ensure the file remains accessible after installation completes

#### Implementation details:

Modify the installer creation process to always preserve envs.txt as a lockfile, and add integration tests for installer creation with lockfiles preserved.

#### Additional Context:
Since envs.txt already contains the @EXPLICIT directive used for exact package installation, transforming it into a persistent lockfile is more straightforward than generating and copying a separate artifact. This can be done by default for all installers.

jaimergp:
### Description

Closes #1052 

### Checklist - did you ...

### DIFF
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -1375,15 +1375,12 @@ Section "Install"
     ${Print} "Setting up the {{ env.name }} environment..."
     SetDetailsPrint listonly
 
-    # List of packages to install
-    SetOutPath "{{ env.env_txt_dir }}"
-    File "{{ env.env_txt_abspath }}"
 
     # A conda-meta\history file is required for a valid conda prefix
     SetOutPath "{{ env.conda_meta }}"
     File "{{ env.history_abspath }}"
-    # We also place a copy of the env.txt file here, as a lockfile
-    File /oname=initial-state.lockfile.txt "{{ env.env_txt_abspath }}"
+    # List of packages to install, as a lockfile
+    File "{{ env.lockfile_txt_abspath }}"
 
     # Set channels
     System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_CHANNELS", "{{ channels }}").r0'
@@ -1393,20 +1390,16 @@ Section "Install"
     # Run conda install
     ${If} $Ana_CreateShortcuts_State = ${BST_CHECKED}
         ${Print} "Installing packages for {{ env.name }}, creating shortcuts if necessary..."
-        push '"$INSTDIR\_conda.exe" install --offline -yp "{{ env.prefix }}" --file "{{ env.env_txt }}" {{ env.shortcuts }} {{ env.no_rcs_arg }}'
+        push '"$INSTDIR\_conda.exe" install --offline -yp "{{ env.prefix }}" --file "{{ env.lockfile_txt }}" {{ env.shortcuts }} {{ env.no_rcs_arg }}'
     ${Else}
         ${Print} "Installing packages for {{ env.name }}..."
-        push '"$INSTDIR\_conda.exe" install --offline -yp "{{ env.prefix }}" --file "{{ env.env_txt }}" --no-shortcuts {{ env.no_rcs_arg }}'
+        push '"$INSTDIR\_conda.exe" install --offline -yp "{{ env.prefix }}" --file "{{ env.lockfile_txt }}" --no-shortcuts {{ env.no_rcs_arg }}'
     ${EndIf}
     push 'Failed to link extracted packages to {{ env.prefix }}!'
     push 'WithLog'
     SetDetailsPrint listonly
     call AbortRetryNSExecWait
     SetDetailsPrint both
-
-    # Cleanup {{ env.name }} env.txt
-    SetOutPath "$INSTDIR"
-    Delete "{{ env.env_txt }}"
 
     # Restore shipped conda-meta\history for remapped
     # channels and retain only the first transaction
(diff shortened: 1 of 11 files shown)


constructor/nsis/main.nsi.tmpl
```diff
@@ -1382,6 +1382,8 @@ Section "Install"
     # A conda-meta\history file is required for a valid conda prefix
     SetOutPath "{{ env.conda_meta }}"
     File "{{ env.history_abspath }}"
+    # We also place a copy of the env.txt file here, as a lockfile
+    File /oname=initial-state.lockfile.txt "{{ env.env_txt_abspath }}"
```
marcoesters: Instead of renaming the `env.txt` file during the installation, why don't we just package `env.txt` as `conda-meta/initial-state.lockfile.txt` and use that file for the `--file` argument?

tests/test_examples.py
```diff
@@ -465,7 +465,15 @@ def test_example_extra_pages_win(tmp_path, request, extra_pages, monkeypatch):
 def test_example_extra_envs(tmp_path, request):
     input_path = _example_path("extra_envs")
     for installer, install_dir in create_installer(input_path, tmp_path):
-        _run_installer(input_path, installer, install_dir, request=request)
+        _run_installer(input_path, installer, install_dir, request=request, uninstall=False)
+        assert (
+            "@EXPLICIT" in (install_dir / "conda-meta" / "initial-state.lockfile.txt").read_text()
+        )
+        for envtxt in install_dir.glob("envs/*/conda-meta/initial-state.lockfile.txt"):
+            assert "@EXPLICIT" in envtxt.read_text()
```
```suggestion
        for env in install_dir.glob("envs/*/conda-meta/"):
            envtxt / "initial-state.lockfile.txt"
            assert envtxt.exists()
            assert "@EXPLICIT" in envtxt.read_text()
```

Otherwise, your test will incorrectly pass if the lock file doesn't exist and we want all environments to have a lock file.

constructor/nsis/main.nsi.tmpl
jaimergp: That works too. I wanted to start with the smallest diff possible.

tests/test_examples.py
```diff
@@ -465,7 +465,17 @@ def test_example_extra_pages_win(tmp_path, request, extra_pages, monkeypatch):
 def test_example_extra_envs(tmp_path, request):
     input_path = _example_path("extra_envs")
     for installer, install_dir in create_installer(input_path, tmp_path):
-        _run_installer(input_path, installer, install_dir, request=request)
+        _run_installer(input_path, installer, install_dir, request=request, uninstall=False)
+        assert (
+            "@EXPLICIT" in (install_dir / "conda-meta" / "initial-state.lockfile.txt").read_text()
+        )
+        for env in install_dir.glob("envs/*/conda-meta/"):
+            envtxt / "initial-state.lockfile.txt"
```
```suggestion
            envtxt = env / "initial-state.lockfile.txt"
```

My bad! I introduced a syntax error

constructor/nsis/main.nsi.tmpl
jaimergp: Changed. Let's see if I got it right :P

constructor/osxpkg.py
```diff
@@ -562,8 +562,18 @@ def create(info, verbose=False):
...
+    os.makedirs(conda_meta)
     preconda.write_files(info, pkgs_dir)
+    # We need to move the initial-state.explicit.txt files from pkgs/ to their conda-meta targets
+    shutil.move(join(pkgs_dir, "initial-state.explicit.txt"), conda_meta)
+    if isdir(join(pkgs_dir, "envs")):
+        for envname in os.listdir(join(pkgs_dir, "envs")):
+            lockfile = join(pkgs_dir, "envs", envname, "initial-state.explicit.txt")
+            env_conda_meta = join(prefix, "envs", envname, "conda-meta")
+            os.makedirs(env_conda_meta)
+            shutil.move(lockfile, env_conda_meta)
```
jaimergp: Not super proud of this block, but the alternative is refactoring the logic in `preconda` so it's multi-directory (and not just assume that everything goes in `dst_dir == "pkgs/"`).

What do you think @marcoesters? Worth the refactor or not? There'll still be differences between Unix and Windows.

marcoesters: I'm not sure if it's much of a refactor. Since `write_files` writes very specific files with very specific file locations, you could pass in `prefix` and adjust the paths inside the function accordingly. This would make the SH code a little simpler, too.

jaimergp: Let's see: [fe1e040](https://github.com/conda/constructor/pull/1059/commits/fe1e040a7152f8cb0214a2c48d366740645749fc)

marcoesters: I think this looks pretty clean and extendable. The directory creation can probably be further abstracted by using `Path().parent.mkdir(parents=True, exist_ok=True)`, but that's probably better suited for #1063.

marcoesters (approved):

Issue #1020, opened 2025-07-22
Conda uninstaller create file when directory contains space in it path


CelestineSauvage:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

I have a problem when using the uninstaller with PowerShell. My installation folder is located in `C:\Program Files`.
When I run the uninstallation with
`Start-Process -FilePath "C:\Program Files\Miniconda3\Uninstall-Miniconda3.exe" -ArgumentList "/S" -NoNewWindow -Wait`
a file named `Program` is created at the root of my `C` drive. This file is non-empty and contains (in french) : 

```
'""C:\Program' n'est pas reconnu en tant que commande interne
ou externe, un programme ex‚cutable ou un fichier de commandes.
```

I have tried the options listed in the documentation here: https://conda.github.io/constructor/cli-options/#windows-installers, using the /D option where they mention space in path (if I understand correctly?), but I still get the same result.

marcoesters:
Thank you for your report! I was able to reproduce it and get the same file with the same error. I do not think the error is because of how you call the uninstaller. The error points to something not being quoted correctly, possibly inside a batch script.

I performed just the [file removal step](https://github.com/conda/constructor/blob/e39990d0c37235dd5d94a14f2ce4d61b9d665f88/constructor/nsis/main.nsi.tmpl#L1632) and found the following error:

```
WARNING conda.core.link:run_script(1581): pre-unlink script failed for package defaults/win-64::conda-anaconda-tos-0.2.1-py313haa95532_0
consider notifying the package maintainer
```

My first thought was that the pre-unlink has an error because of spaces in the path, but when I install Miniconda into `%USERPROFILE%\miniconda with spaces` and remove `conda-anaconda-tos`, I do not get this warning.

So, I don't think the issue is with `constructor` itself, but I don't quite know what the root cause is. I will keep debugging.

Issue #549, opened 2022-08-15
RFC: Undoing constructor / conda-standalone cross "bundling"


jaimergp:
`constructor` relies on a frozen version of `conda` named `conda-standalone`. `conda-standalone` is built off two feedstocks:

* https://github.com/conda-forge/conda-standalone-feedstock
* https://github.com/AnacondaRecipes/conda-standalone-feedstock

We will be syncing both to `conda/conda-standalone` soon.

`conda-standalone` is not _just_ a frozen `conda`. It contains some modifications:

Patches:
* `noarch` entry-points compatibility for Windows: `Scripts/conda.exe` won't be available in the PyInstaller executable, so we add it ourselves from the conda sources (`conda/shell/cli-$ARCH.exe`)
* Do not conda-activate the wrapped subprocess for Windows; just modify the PATH as needed.

Modified CLI:
* The frozen conda.exe contains a hook to detect whether the CLI is using a `constructor` subcommand (e.g. `conda constructor...`) and redirects those calls to a bundled `_nsis.py` file.

[
`_nsis.py` comes from `constructor`](https://github.com/conda/constructor/blob/main/constructor/nsis/_nsis.py)! It's a helper script to aid Windows installers with some steps, like menu items creation.

Windows' NSIS installers end up using `_nsis.py` in two different ways:

* Directly, via the bundled Python. This is the reason why [`python` is ALWAYS added to the list of packages](https://github.com/conda/constructor/blob/7e76a572d5b87b3aa319218fd73c3cc48506b6e3/constructor/fcp.py#L257) in the installer. 
* Through conda-standalone's `constructor` entry point.

The thing is that `conda-standalone` already features a FULL Python installation inside the frozen executable, so there's no need to force feed Python into the installer again.

My plan is to:

1. Move the `_nsis.py` (and dependents) to `conda/conda-standalone`
2. Replace the direct calls to `_nsis.py` with equivalent functionality through `conda-standalone`
3. Do not require Python in the `base` environment anymore, but leave `conda.exe` in the installation, because it will be needed at uninstall time.
4. (At some point) Package the `constructor` helpers in `conda-standalone` as a conda plugin, so building `conda-standalone` is less hacky

Thoughts, @conda/constructor?

cc @wolfv - since this might affect the micromamba integrations.

jezdez:
Nice, this sounds like a great plan and should make things much more obvious.

jaimergp:
Still ongoing!

PR #1069, opened 2025-09-18
Convert script and directory `_nsis.py` subcommands into NSIS code

marcoesters:
### Description

Convert simple script and directory subcommands into NSIS code to further decouple the python dependency from EXE installers (#549). Scripts are already executed directly via `cmd.exe` for pre-install scripts and `mkdirs` is a simple directory creation command.

Note that this means that `post-install` and `pre-uninstall` script failures will cause the installation/uninstallation process to fail now. I think this is good behavior as it has masked issues in the past (see https://github.com/conda/constructor/pull/942 and https://github.com/conda/constructor/issues/1020).

### Checklist - did you ...

### DIFF
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -1414,11 +1414,10 @@ Section "Install"
     AddSize {{ SIZE }}
 
 {%- if has_conda %}
-    ${Print} "Initializing conda directories..."
-    push '"$INSTDIR\pythonw.exe" -E -s "$INSTDIR\Lib\_nsis.py" mkdirs'
-    push 'Failed to initialize conda directories'
-    push 'WithLog'
-    call AbortRetryNSExecWait
+    StrCpy $R0 "$INSTDIR\envs"
+    ${IfNot} ${FileExists} "$R0"
+        CreateDirectory "$R0"
+    ${EndIf}
 {%- endif %}
 
     ${If} $Ana_PostInstall_State = ${BST_CHECKED}
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -1617,7 +1617,6 @@ Section "Uninstall"
         call un.AbortRetryNSExecWait
     ${EndIf}
     !insertmacro AbortRetryNSExecWaitLibNsisCmd "rmpath"
-    !insertmacro AbortRetryNSExecWaitLibNsisCmd "rmreg"
 
     # Parse arguments
     StrCpy $R0 ""
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -1600,7 +1600,12 @@ Section "Uninstall"
     ${EndIf}
 
 {%- if uninstall_with_conda_exe %}
-    !insertmacro AbortRetryNSExecWaitLibNsisCmd "pre_uninstall"
+    ${If} ${FilExists} "$INSTDIR\pkgs\pre_uninstall.bat"
+        ${Print} "Running pre_uninstall scripts..."
+        push '${CMD_EXE} /D /C "$INSTDIR\pkgs\pre_uninstall.bat"'
+        push "Failed to run pre_uninstall"
+        push 'WithLog'
+    ${EndIf}
     !insertmacro AbortRetryNSExecWaitLibNsisCmd "rmpath"
     !insertmacro AbortRetryNSExecWaitLibNsisCmd "rmreg"
 
@@ -1650,7 +1655,12 @@ Section "Uninstall"
     call un.AbortRetryNSExecWait
     SetDetailsPrint both
 {%- endfor %}
-    !insertmacro AbortRetryNSExecWaitLibNsisCmd "pre_uninstall"
+    ${If} ${FilExists} "$INSTDIR\pkgs\pre_uninstall.bat"
+        ${Print} "Running pre_uninstall scripts..."
+        push '${CMD_EXE} /D /C "$INSTDIR\pkgs\pre_uninstall.bat"'
+        push "Failed to run pre_uninstall"
+        push 'WithLog'
+    ${EndIf}
     !insertmacro AbortRetryNSExecWaitLibNsisCmd "rmpath"
     !insertmacro AbortRetryNSExecWaitLibNsisCmd "rmreg"
 
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -132,6 +132,19 @@ var /global InstMode # 0 = Just Me, 1 = All Users.
 !define JUST_ME 0
 !define ALL_USERS 1
 
+# Find cmd.exe
+var /global CMD_EXE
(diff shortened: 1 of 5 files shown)


marcoesters:
> @marcoesters Is this something that is awaiting review? Let me know and I can review it.

You are free to leave comments, but in the end, I need approval from a `constructor` maintainer.

constructor/nsis/main.nsi.tmpl
```diff
@@ -1352,17 +1370,7 @@ Section "Install"
...
-        # This 'FileExists' also returns True for directories
-        ${If} ${FileExists} "$5"
-            push '"$5\System32\cmd.exe" /D /C "$INSTDIR\pkgs\pre_install.bat"'
-        ${ElseIf} ${FileExists} "$6"
-            push '"$6\System32\cmd.exe" /D /C "$INSTDIR\pkgs\pre_install.bat"'
-        ${Else}
-            # Cross our fingers CMD is in PATH
-            push 'cmd.exe /D /C "$INSTDIR\pkgs\pre_install.bat"'
-        ${EndIf}
+        push '$CMD_EXE /D /C "$INSTDIR\pkgs\pre_install.bat"'
```
lrandersson:
```suggestion
        push '"$CMD_EXE" /D /C "$INSTDIR\pkgs\pre_install.bat"'
```
I don't think we ever will find `cmd.exe` on a path with spaces, but, to be extra safe. Feel free to just ignore this suggestion.

constructor/nsis/main.nsi.tmpl
```diff
@@ -1414,19 +1422,20 @@ Section "Install"
...
-    ${If} $Ana_PostInstall_State = ${BST_CHECKED}
-        ${Print} "Running post install..."
-        push '"$INSTDIR\pythonw.exe" -E -s "$INSTDIR\Lib\_nsis.py" post_install'
-        push 'Failed to run post install script'
-        push 'WithLog'
-        call AbortRetryNSExecWait
+    ${If} ${FileExists} "$INSTDIR\pkgs\post_install.bat"
+        ${If} $Ana_PostInstall_State = ${BST_CHECKED}
+            ${Print} "Running post install..."
+            push '$CMD_EXE /D /C "$INSTDIR\pkgs\post_install.bat"'
```
lrandersson:
```suggestion
            push '"$CMD_EXE" /D /C "$INSTDIR\pkgs\post_install.bat"'
```
Same here, optional.

constructor/nsis/main.nsi.tmpl
```diff
@@ -1598,9 +1609,14 @@ Section "Uninstall"
     ${EndIf}
 
 {%- if uninstall_with_conda_exe %}
-    !insertmacro AbortRetryNSExecWaitLibNsisCmd "pre_uninstall"
+    ${If} ${FileExists} "$INSTDIR\pkgs\pre_uninstall.bat"
+        ${Print} "Running pre_uninstall scripts..."
+        push '$CMD_EXE /D /C "$INSTDIR\pkgs\pre_uninstall.bat"'
```
lrandersson:
```suggestion
        push '"$CMD_EXE" /D /C "$INSTDIR\pkgs\pre_uninstall.bat"'
```

constructor/nsis/main.nsi.tmpl
```diff
@@ -1648,7 +1664,13 @@ Section "Uninstall"
     call un.AbortRetryNSExecWait
     SetDetailsPrint both
 {%- endfor %}
-    !insertmacro AbortRetryNSExecWaitLibNsisCmd "pre_uninstall"
+    ${If} ${FileExists} "$INSTDIR\pkgs\pre_uninstall.bat"
+        ${Print} "Running pre_uninstall scripts..."
+        push '$CMD_EXE /D /C "$INSTDIR\pkgs\pre_uninstall.bat"'
```
lrandersson:
```suggestion
        push '"$CMD_EXE" /D /C "$INSTDIR\pkgs\pre_uninstall.bat"'
```

constructor/nsis/main.nsi.tmpl
```diff
@@ -1598,9 +1609,14 @@ Section "Uninstall"
...
+        push "Failed to run pre_uninstall"
+        push 'WithLog'
+        call un.AbortRetryNSExecWait
+    ${EndIf}
     !insertmacro AbortRetryNSExecWaitLibNsisCmd "rmpath"
-    !insertmacro AbortRetryNSExecWaitLibNsisCmd "rmreg"
```
lrandersson, line 1603: Why is this no longer needed?

marcoesters: It was actually never needed. The uninstallation via `conda-standalone` already performs the same function during the installation process. I included that line back then by mistake.

lrandersson: I see, thanks!

jaimergp (approved):
Nice cleanups! Thanks!

marcoesters:
I found a way to remove the `rmreg` command with uninstallers that don't use `conda-standalone`. I think this should be part of this PR.

constructor/nsis/main.nsi.tmpl
```diff
@@ -1673,11 +1673,16 @@ Section "Uninstall"
     ${EndIf}
     !insertmacro AbortRetryNSExecWaitLibNsisCmd "rmpath"
 {%- if has_conda %}
+    ${If} ${FileExists} "$INSTDIR\.nonadmin"
```
lrandersson, line 1676: Optional:
In the Function `un.onInit` we also have this logic:
```
    ${If} ${FileExists} "$INSTDIR\.nonadmin"
        SetShellVarContext Current
    ${Else}
        SetShellVarContext All
    ${EndIf}
```
If you prefer, you can create a new variable (maybe `un.IsUserInstall`?) and assign that variable within that function, to reduce the number of `if .nonadmin exists`.

I have no strong opinion though, I also like the simplicity with simply using `$R0` as in your commit.

lrandersson (approved):

constructor/nsis/main.nsi.tmpl
```diff
@@ -1648,9 +1664,29 @@ Section "Uninstall"
...
+        StrCpy $R0 "user"
+    ${Else}
+        StrCpy $R0 "system"
+    ${EndIf}
+    # When running conda.bat directly, there is a non-fatal error
+    # that DOSKEY (called by conda_hook.bat) is not a valid command.
+    # While the operation still succeeds, this error is confusing.
+    # Calling via cmd.exe fixes that.
+    push '"$CMD_EXE" /D /C "$INSTDIR\condabin\conda.bat" init cmd.exe --reverse --$R0'
```
jaimergp, line 1685:
```suggestion
    push '"$CMD_EXE" /D /C "$INSTDIR\condabin\conda.bat" init --reverse --$R0'
```

Should we revert for both cmd and powershell (which is achieved by passing no positional arguments)?

marcoesters: The powershell function does not perform path validation, so it will also revert profiles that point to a different installation. The cmd.exe function, on the other hand, only reverts the autorun entry that belongs to the installation.

PR #1082, opened 2025-10-08
Add unset of environment variables

lrandersson:
### Description
Added unset of additional environment variables for consistency.

### Checklist - did you ...

- [x] Add / update necessary tests? - _sort of_, by verifying the state of the variables in `post_install.sh`

### DIFF
--- a/news/1082-unset-variables
+++ b/news/1082-unset-variables
@@ -1,6 +1,6 @@
 ### Enhancements
 
-* The shell-based installers now sanitize additional environment variables for consistency with security best practices. (#1082)
+* Unset additional environment variables in shell-based installers to avoid accidental loading of external libraries. (#1082)
 
 ### Bug fixes
 


constructor/header.sh
```diff
@@ -10,10 +10,10 @@
 set -eu
 
 {%- if osx %}
-unset DYLD_LIBRARY_PATH DYLD_FALLBACK_LIBRARY_PATH
+unset DYLD_LIBRARY_PATH DYLD_FALLBACK_LIBRARY_PATH DYLD_INSERT_LIBRARIES DYLD_FRAMEWORK_PATH
```
lrandersson, line 13: I also just discovered that in `constructor/osx/run_installation.sh`, line 23 we have:
`unset DYLD_LIBRARY_PATH`, how does this relate to the header? Can the entry in `run_installation.sh` be removed?

marcoesters: Good catch. They need to be added there, too - these scripts are for the PKG installer whereas the header is for the SH installer.

examples/scripts/post_install.sh
```diff
@@ -23,13 +23,25 @@ test "${CUSTOM_VARIABLE_2}" = '$ECOND-CUSTOM_'\''STRING'\'' WITH SPACES AND @*!
...
+        fi
+    done
+    return $failed
+}
+
 if [[ $(uname -s) == Linux ]]; then
-    if [[ ${INSTALLER_PLAT} != linux-* ]]; then
-        exit 1
-    fi
+    [[ ${INSTALLER_PLAT} != linux-* ]] && exit 1
```
lrandersson: With respect to the changes done here https://github.com/conda/constructor/pull/1085/files#diff-138af3e33914ad28d6920a90ef40e891dc0400b8d7e6c3ff46f376026af87666R7, I am going to change all of these one-liners back to make it easier to read - and for consistency.

constructor/header.sh
```diff
@@ -10,10 +10,10 @@
...
 {%- else %}
 export OLD_LD_LIBRARY_PATH="${LD_LIBRARY_PATH:-}"
```
marcoesters: Do you see a reason to keep `OLD_LIBRARY_PATH`? It seems to be unused.

news/1082-unset-variables
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* The shell-based installers now sanitize additional environment variables for consistency with security best practices. (#1082)
```
marcoesters:
```suggestion
* Unset additional environment variables in shell-based installers to avoid accidental loading of external libraries. (#1082)
```

Since I don't think this is really exploitable, I would avoid raising the specter or a security vulnerability here. However, a user could accidentally shoot themselves in the foot here.

lrandersson: I agree, good idea. Updated in 8ff68ab

constructor/header.sh
lrandersson: No, you're right, I don't see a reason. There was an example that did `echo`, but it seems unnecessary, see dc3a58fe. There is also this release note entry from 2017:
```
  * proved access to LD_LIBRARY_PATH in Linux install scripts by storing it
    as OLD_LD_LIBRARY_PATH
```
But reading this makes me think this is an old artifact that is no longer needed. Let me know if you disagree.

Issue #967, opened 2025-03-26
Use Briefcase as an installer backend


mhsmith:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

This issue is a summary of some recent discussions between the BeeWare and Conda teams. We're posting it in public to get feedback from the larger community.

Briefcase (part of the BeeWare project) is an open-source packaging tool which has some overlap with Constructor's functionality. It's a mature and well-tested tool that supports building Mac, Linux and Windows installers in multiple formats, including signing and notarization.

Briefcase currently combines two functions: creating a directory tree containing a Python app, and packaging it into an installer. If it had an option to use the packaging step independently, then it could create installers with arbitrary content. In this mode, Briefcase’s own Python app functionality would not be involved.

Constructor could then use Briefcase to build installers, with the following division of responsibility:

* Briefcase provides non-conda-specific installer mechanisms.
* Constructor uses those mechanisms to implement conda-specific functionality.

From the Constructor user’s point of view, very little would change: they would continue to use Constructor’s existing CLI and configuration file format. The fact that Constructor used Briefcase to build installers would only be an implementation detail.

Constructor could start using Briefcase incrementally, one installer format at a time, without disturbing the others.

### Why is this needed?

As described in #646, for several years the Constructor developers have wanted to stop using NSIS because of its developer-unfriendliness. However, the effort required to produce an entirely new Windows backend has so far been too much to face.

Briefcase has an actively-maintained Windows installer implementation based on the modern WiX toolkit. Reusing this should be significantly easier than creating a WiX backend from scratch.

Briefcase could also be used to build installers in other formats, e.g. .pkg installers for macOS, which would relieve the Constructor project of even more maintenance burden.

mhsmith:
### Remaining features required for production use

Feel free to edit this comment as necessary.

#### In Briefcase (add links to issues once they exist):

* Welcome screen 
  * With custom text and branding image.
* Conclusion screen
  * With custom text, and some means of launching web pages or one of the installed applications.
* Developer configuration of default target directory.
* User configuration of actual target directory.
  * This is currently always Program Files when installing for "all users", and AppData\Local\Programs when installing for "just me".
  * Would require thought about how to handle multiple concurrent installs.
* Logging improvements
  * The install script currently runs in a separate console window, which allows better terminal animations, but is potentially confusing to the user.
* Uninstaller options

#### In Constructor:

[...]

Issue #646, opened 2023-02-08
Alternatives to NSIS


jaimergp:
I think it's safe to say NSIS is not the most development friendly engine ever, specially coming over from the Python world. While incredibly performant, powerful, customizable and extensible, it comes with a few gotchas:

- Tricky to debug and test. The hoops we go through in the `run_examples.py` showcase this nicely.
- Difficult-to-obtain logs. No output in stdout!
- Scripting syntax is sometimes too low-level.

Ideally, I would love to see if we can find a replacement with the following features:

- Free and redistributable (and open-source?)
- Scriptable in a sane way that allows us to run custom code
- Customizable and extensible UI (branding, text, extra pages)
- Good development/debugging experience (CLI installs should be as informative as GUI ones)

[This website](http://www.installsite.org/pages/en/msi/authoring.htm) ([plus this one](http://www.installsite.org/pages/en/tt_nonmsi.htm)) has TONS of info about the whole landscape. I'll summarize the main findings here, focusing on free products:

- [MSI installers](http://www.installsite.org/pages/en/w2k_aboutmsi.htm)
  - The "canonical" solution (like PKGs in macOS).
  - Robust, reliable, but not extensible.
  - It even has a standard library package, [`msilib`](https://docs.python.org/3/library/msilib.html) (deprecated in 3.11 though). Low-level.
- [WiX](https://wixtoolset.org/)
  - An abstraction on top of the MSI technology
  - Used by other tools like `briefcase` ([see code](https://github.com/beeware/briefcase/blob/8d503064b40313e2a0c4559e59cdc7c24b72adac/src/briefcase/platforms/windows/__init__.py#L107))
  - [Customizable](https://wixtoolset.org/docs/v3/wixui/wixui_customizations/)
  - V4 recently released, might be a good time to give it a try
  - Based on XML inputs, apparently extensible with C# (not sure about this part)
  - [Real-life example with a lot of documentation](https://helgeklein.com/blog/real-world-example-wix-msi-application-installer/)
  - [Examples that run custom code](https://cs.github.com/?scopeName=All+repos&scope=&q=WixQuietExec64+path%3A*.wxs)
- [Inno Setup](https://jrsoftware.org/isinfo.php)
  - Extensible with Pascal (might be a bit saner than NSIS' but still...)
  - Does not depend on MSI technology
  - Another veteran with good track record
  - [Good CLI options](https://jrsoftware.org/ishelp/index.php?topic=setupcmdline)
[...]

PR #1084, opened 2025-10-09
Initial MSI implementation, based on Briefcase

mhsmith:
### Description

Initial work towards:
* #646
* #967

Requires Briefcase to be installed in the environment, including the following updates:
* https://github.com/beeware/briefcase/pull/2496
* https://github.com/beeware/briefcase/pull/2534

As discussed, this PR is sufficient to build a simple MSI installer (with no install-time options) from the [miniconda3 recipe](https://github.com/anaconda/installers/tree/main/products/miniconda3). After the Briefcase updates are merged, I'll post a comment here with instructions on how to do that.

@freakboy3742: FYI

### Checklist - did you ...

### DIFF
--- a/dev/environment.yml
+++ b/dev/environment.yml
@@ -10,5 +10,3 @@ dependencies:
   - jinja2
   - jsonschema >=4
   - pydantic 2.11.*
-  - briefcase >=0.3.26  # [win]
-  - tomli-w >=1.2.0  # [win]
--- a/dev/extra-requirements-windows.txt
+++ b/dev/extra-requirements-windows.txt
@@ -1,1 +1,3 @@
+conda-forge::briefcase>=0.3.26
 conda-forge::nsis>=3.08=*_log_*
+conda-forge::tomli-w>=1.2.0


constructor/briefcase.py
```diff
@@ -0,0 +1,150 @@
...
+        [briefcase, "package"] + (["-v"] if verbose else []),
+        cwd=tmp_dir,
+        check=True,
+    )
+
+    dist_dir = tmp_dir / "dist"
+    msi_paths = list(dist_dir.glob("*.msi"))
+    if len(msi_paths) != 1:
+        raise RuntimeError(f"Found {len(msi_paths)} MSI files in {dist_dir}")
+    shutil.copy(msi_paths[0], info["_outpath"])
```
jaimergp: Should this be a `move` command instead? In heavy installers and reduced disk storage availability, it might be the difference between success and error.

constructor/data/construct.schema.json
```diff
@@ -224,6 +224,7 @@
       "enum": [
         "all",
         "exe",
+        "msi",
```
jaimergp, line 227: This change should be applied in `_schema.py` and then propagated.

tests/test_briefcase.py
jaimergp, line 1: We should also have an integration test in `test_examples.py` once the dependencies are ready.

constructor/briefcase.py
mhsmith: Done.

constructor/data/construct.schema.json
mhsmith: Done.

mhsmith:
Actually, I'll move this to #967.

constructor/briefcase.py
```diff
@@ -0,0 +1,153 @@
...
+BRIEFCASE_DIR = Path(__file__).parent / "briefcase"
+EXTERNAL_PACKAGE_PATH = "external"
+
+logger = logging.getLogger(__name__)
+
+
+def get_name_version(info):
+    name = info["name"]
+    if not name:
+        raise ValueError("Name is empty")
```
```suggestion
    if not (name := info.get("name")):
        raise ValueError("Name is empty")
```

If we are concerned about empty values, we should use `get`, too.

tests/test_briefcase.py
marcoesters: I think most integration tests have `installer_type: all`, which will break if features are missing. I suggest replacing them with `installer_type: {{ 'exe' if sys.platform == 'win32' else 'all'` for those that fail because of missing features. That will also give us a good idea about when we have feature parity.

constructor/briefcase.py
```diff
@@ -0,0 +1,153 @@
...
+
+    # Briefcase requires version numbers to be in the canonical Python format, and some
```
marcoesters, line 34: Python is not fully compatible with SemVer, so that could be a pretty significant limitation.

It will at least require a few version changes in our integration test examples: https://github.com/conda/constructor/blob/170417a1800c0dd12a5d29f4882091b9d818dbaf/examples/scripts/construct.yaml#L5

constructor/briefcase.py
```diff
@@ -0,0 +1,153 @@
...
+        )
+    match = matches[-1]
+    version = match.group()
+
+    # Treat anything else in the version string as part of the name.
+    start, end = match.span()
+    strip_chars = " .-_"
+    before = info["version"][:start].strip(strip_chars)
+    after = info["version"][end:].strip(strip_chars)
+    name = " ".join(s for s in [name, before, after] if s)
```
marcoesters, line 62: So, for something like `Miniconda3-py310_25.9.1-3-Windows-x86_64.msi`, we have the following?

```
name = Miniconda3 py310
version = 25.9.1.3
```

constructor/briefcase.py
```diff
@@ -0,0 +1,153 @@
...
+
+    return name, version
+
+
+# Some installer types use the reverse domain ID to detect when the product is already
+# installed, so it should be both unique between different products, and stable between
+# different versions of a product.
+def get_bundle_app_name(info, name):
+    # If reverse_domain_identifier is provided, use it as-is, but verify that the last
+    # component is a valid Python package name, as Briefcase requires.
```
marcoesters: Since `constructor` doesn't require Python-based packages, will we run into issues with non-Python namespaces?

tests/test_briefcase.py
lrandersson: @mhsmith just for completeness, I've added (and working on finalizing it) the integration tests in `test_examples.py` in this branch: https://github.com/lrandersson/constructor/tree/briefcase
I branched from the branch that's in this PR.

constructor/briefcase.py
```diff
@@ -0,0 +1,153 @@
...
+
+    copy_conda_exe(external_dir, "_conda.exe", info["_conda_exe"])
+
+    briefcase = Path(sysconfig.get_path("scripts")) / "briefcase.exe"
+    logger.info("Building installer")
+    run(
+        [briefcase, "package"] + (["-v"] if verbose else []),
+        cwd=tmp_dir,
+        check=True,
+    )
```
marcoesters, line 159: I have two questions here:

1. Is it possible to use `briefcase` as an API rather than a CLI?
2. This looks like Windows - do we need any guards here?

Either way, we should make sure that `briefcase.exe` exists before calling it.

constructor/briefcase/run_installation.bat
```diff
@@ -0,0 +1,9 @@
+_conda constructor --prefix . --extract-conda-pkgs
```
marcoesters: Let's use absolute paths instead of relative paths.

pyproject.toml
```diff
@@ -16,7 +16,8 @@ dependencies = [
     "ruamel.yaml >=0.11.14,<0.19",
     "pillow >=3.1 ; platform_system=='Windows' or platform_system=='Darwin'",
     "jinja2",
-    "jsonschema >=4"
+    "jsonschema >=4",
+    "tomli-w >=1.2.0",
```
marcoesters: Since this is only needed for MSI installers, this should probably only be added for Windows - same in the recipe.

This is also missing `briefcase` as a dependency.

tests/test_briefcase.py
lrandersson: https://github.com/mhsmith/constructor/pull/1

constructor/briefcase.py
lrandersson: I fixed it in https://github.com/mhsmith/constructor/pull/1

constructor/briefcase.py
lrandersson: Fixed _temporarily_ in https://github.com/mhsmith/constructor/pull/1

constructor/briefcase.py
lrandersson: I added 2) in https://github.com/mhsmith/constructor/pull/1, the check for `briefcase.exe` is also there

pyproject.toml
lrandersson: In https://github.com/mhsmith/constructor/pull/1

constructor/briefcase.py
mhsmith: As far as MSI is concerned, the version number is only used for 2 purposes:

* Display in the installed apps list. This is just for the user's information, so I think it's acceptable if some of the construct.yaml `version` gets moved into the `name` for display purposes, as long as all the information is still present.

* Blocking the installer of an old version if a new version is already installed. Since constructor itself doesn't define any version ordering rules, it's impossible to do this perfectly for all version schemes. The current code covers a reasonably large number of cases, but it could be extended if necessary.

Notice the current code only uses the last valid Python package version number it finds. This is to accommodate the Miniconda construct.yaml file, which sets `version` to something like `py313_25.1.2-3`. It's better to transform this into `25.1.2.3` rather than `313.25.1.2.3`.

mhsmith: For versions with no numbers in them at all, we could either reject them as the PR currently does, or display a warning and fall back to a default like 0.0.1 or 1.0.0. That's probably a good idea, since it would allow all existing construct.yaml files to at least build, and the integration tests wouldn't need to change so much.

constructor/briefcase.py
mhsmith: Yes, those would be the name and version from Briefcase's point of view, and that's how they'd be displayed in the Windows apps list. 

The current code can generate these values from a construct.yaml file where `py310` is part of the `version`. There's an example of that in `test_name_version` in test_briefcase.py.

constructor/briefcase.py
mhsmith: As with the version, I guess we could make a best effort to transform it into something valid. In the case of MSI, this value is only used internally by Briefcase and is never visible to the user. So it doesn't matter if it's slightly different to the construct.yaml value, as long as it's consistent.

constructor/briefcase.py
mhsmith: 1) No, the CLI is the only stable interface.
2) The check in [mhsmith#1](https://github.com/mhsmith/constructor/pull/1) looks good to me. The fact that we're running on WIndows is already guarded by `os_allowed` in main.py.

constructor/briefcase.py
mhsmith: I've copied that fix to this PR.

constructor/briefcase.py
mhsmith: I've made it use a default of 0.0.1, so that if a valid version is added in the future, it'll be treated as an upgrade.

constructor/briefcase.py
mhsmith: Done.

constructor/briefcase.py
mhsmith: I've copied the `briefcase.exe` check to this PR.

constructor/briefcase/run_installation.bat
mhsmith: Done.

pyproject.toml
mhsmith: I've updated pyproject.toml, meta.yaml and environment.yml. A new enough version of Briefcase is [now on conda-forge](https://anaconda.org/channels/conda-forge/packages/briefcase/overview), so it no longer needs to be installed manually.

mhsmith: Briefcase 0.3.26 also has a couple of new MSI features:
* Interactive selection of the install directory.
* Uninstaller options – though as it says [here](https://briefcase.beeware.org/en/stable/reference/platforms/windows/#using-options), these currently only appear when you choose "Modify" from the app list rather than "Uninstall", so we'll need to revisit this in the future.

dev/environment.yml
```diff
@@ -10,3 +10,5 @@ dependencies:
   - jinja2
   - jsonschema >=4
   - pydantic 2.11.*
+  - briefcase >=0.3.26  # [win]
+  - tomli-w >=1.2.0  # [win]
```
marcoesters: Environment YAML files do not recognize these selectors as far as I know. the `pillow` line doesn't make sense. I'd rather remove those Windows specific dependencies here.

Those dependencies should go into `extra-requirements-windows.txt`

mhsmith: OK, I've moved the 2 new requirements to `extra-requirements-windows.txt`. The `pillow` line was already there, so I'll leave it alone to avoid conflict with any fix on the main branch.

Issue #1074, opened 2025-09-26
Include license files in PKG and SH installers


marcoesters:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

EXE installers include the license file in the installation directory. PKG and SH installers, on the other hand, do not include the license as a file.

### Why is this needed?

This is useful to refer back to the license of the installer after the installation is completed.

### What should happen?

Similar to Windows, the license file should be added to the installation directory.

PR #1085, opened 2025-10-09
Add license file to .pkg installers

lrandersson:
### Description
To resolve https://github.com/conda/constructor/issues/1074

### Checklist - did you ...

### DIFF
--- a/constructor/preconda.py
+++ b/constructor/preconda.py
@@ -329,7 +329,7 @@ def copy_extra_files(
         list[os.PathLike]: List of normalized paths of copied locations.
     """
 
-    def check_file_exists(file_path):
+    def validate_file_path(file_path: str) -> Path:
         fpath = Path(file_path)
         if not fpath.exists():
             raise FileNotFoundError(f"File {file_path} does not exist.")
@@ -340,13 +340,12 @@ def copy_extra_files(
     copied = []
     for path in extra_files:
         if isinstance(path, str):
-            orig_path = check_file_exists(path)
+            orig_path = validate_file_path(path)
             copied.append(shutil.copy(orig_path, workdir))
         elif isinstance(path, dict):
             assert len(path) == 1
             origin, destination = next(iter(path.items()))
-
-            orig_path = check_file_exists(origin)
+            orig_path = validate_file_path(origin)
             dest_path = Path(workdir) / destination
             dest_path.parent.mkdir(parents=True, exist_ok=True)
             copied.append(shutil.copy(orig_path, dest_path))
--- a/constructor/shar.py
+++ b/constructor/shar.py
@@ -187,13 +187,10 @@ def create(info, verbose=False):
 
     tarball = join(tmp_dir, "pkgs", "tmp.tar")
     t = tarfile.open(tarball, "w")
-
     t.add(preconda_tarball, basename(preconda_tarball))
     t.add(postconda_tarball, basename(postconda_tarball))
-
     if "license_file" in info:
         t.add(info["license_file"], "LICENSE.txt")
-
     for dist in all_dists:
         fn = filename_dist(dist)
         t.add(join(info["_download_dir"], fn), "pkgs/" + fn)
--- a/examples/extra_files/test_install.sh
+++ b/examples/extra_files/test_install.sh
@@ -3,16 +3,30 @@ set -euxo pipefail
 echo "Added by test-install script" > "$PREFIX/test_install_sentinel.txt"
 
 missing=false
-test -f "$PREFIX/more_data/README.md" || { echo "Missing: $PREFIX/more_data/README.md"; missing=true; }
-test -f "$PREFIX/something2.txt" || { echo "Missing: $PREFIX/something2.txt"; missing=true; }
+
+if [ ! -f "$PREFIX/more_data/README.md" ]; then
+    echo "Missing: $PREFIX/more_data/README.md"
+    missing=true
+fi
+
+if [ ! -f "$PREFIX/something2.txt" ]; then
+    echo "Missing: $PREFIX/something2.txt"
+    missing=true
+fi
 
 # Ideally we should test the .pkg and .sh installers separately since
 # the current behavior for .sh-installers is to include but also rename the license file to LICENSE.txt,
-# but for macOS the name of the provided license file remains unchanged.
+# but for .pkg the name of the provided license file remains unchanged.
 if [ "$INSTALLER_TYPE" = "SH" ]; then
-    test -f "$PREFIX/LICENSE.txt" || { echo "Missing: $PREFIX/LICENSE.txt"; missing=true; }
-else # .pkg
-    test -f "$PREFIX/TEST_LICENSE.txt" || { echo "Missing: $PREFIX/TEST_LICENSE.txt"; missing=true; }
+    if [ ! -f "$PREFIX/LICENSE.txt" ]; then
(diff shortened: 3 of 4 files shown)


constructor/shar.py
```diff
@@ -187,10 +187,13 @@ def create(info, verbose=False):
 
     tarball = join(tmp_dir, "pkgs", "tmp.tar")
     t = tarfile.open(tarball, "w")
+
     t.add(preconda_tarball, basename(preconda_tarball))
     t.add(postconda_tarball, basename(postconda_tarball))
+
```
lrandersson: Looks like `.sh` installers already include the license file but always renames it to `LICENSE.txt`? Or am I missing something?
```
 if "license_file" in info:
        t.add(info["license_file"], "LICENSE.txt")
```

To follow-up: Since the `.sh` installers rename the license file, should we also do this for the `.pkg` installers? I like the idea of consistency but I also think that perhaps we should not change the name of the license file for the user? No strong opinion.

marcoesters: I must have missed that in my installations, but you are right. PKG is sufficient then.

lrandersson: No problem, I just want to ensure I got the correct understanding!

examples/extra_files/test_install.sh
```diff
@@ -4,3 +4,7 @@ echo "Added by test-install script" > "$PREFIX/test_install_sentinel.txt"
 
 test -f "$PREFIX/more_data/README.md"
 test -f "$PREFIX/something2.txt"
+# Ideally we should test the .pkg and .sh installers separately since
+# the current behavior for .sh-installers is to include but also rename the license file to LICENSE.txt,
+# but for macOS the name of the provided license file remains unchanged.
+test -f "$PREFIX/LICENSE.txt"
```
lrandersson: This feels a bit clunky, and ideally I would like to do this differently and test all the different scenarios. To my understanding:
1. For Windows we hardcode `placeholder_license.txt` as its license file (via `winexe.py`)
2. Shell-based installers always install the license file as `$PREFIX/LICENSE.txt`
3. macOS installs (currently) the license file into `$PREFIX` (no renaming). We can change this easily though in this PR to at least make it consistent with 2) above.

Is there any simple way while `test_install.sh` is running to check if we are a `.pkg` installer or `.sh` installer? This could improve the testing at least between the different platforms and installer types.

marcoesters: > Is there any simple way while test_install.sh is running to check if we are a .pkg installer or .sh installer? 

Yes, using the `INSTALLER_TYPE` environment variable.

I agree that this is clunky though. Using a post-install script to test for files that are persistent is not ideal. Ideally, you would make these checks in `python` to be more transparent. However, on Windows, the installations are uninstalled by default, so that would have to change: https://github.com/conda/constructor/blob/bc0de3a133884604dbb5e1b28988dc9fedf6dc56/tests/test_examples.py#L298

The integration test could probably use a refactor.

lrandersson: I did a minor improvement to the test for now.
Question:
* Should we make the renaming of `LICENSE.txt` consistent between `.sh` and `.pkg`? Let me know and I will update it. Regardless if we do or not, I think we should also document it to make it clear. I will do that once we have reached a decision.

marcoesters: If there is no convention, then I think keeping the file name is the better move.

constructor/preconda.py
```diff
@@ -328,18 +328,25 @@ def copy_extra_files(
     Returns:
         list[os.PathLike]: List of normalized paths of copied locations.
     """
+
+    def check_file_exists(file_path):
```
marcoesters:
```suggestion
    def check_file_exists(file_path: str) -> Path:
```

Let's try and add typing where we can. I'm not sure about that function name either - checking isn't really its primary purpose (or that would return a `bool`).

examples/extra_files/test_install.sh
```diff
@@ -2,5 +2,19 @@
 set -euxo pipefail
 echo "Added by test-install script" > "$PREFIX/test_install_sentinel.txt"
 
-test -f "$PREFIX/more_data/README.md"
-test -f "$PREFIX/something2.txt"
+missing=false
+test -f "$PREFIX/more_data/README.md" || { echo "Missing: $PREFIX/more_data/README.md"; missing=true; }
+test -f "$PREFIX/something2.txt" || { echo "Missing: $PREFIX/something2.txt"; missing=true; }
```
marcoesters: Using proper `if` statements would make the code easier to read.

examples/extra_files/test_install.sh
```diff
@@ -2,5 +2,19 @@
...
+
+# Ideally we should test the .pkg and .sh installers separately since
+# the current behavior for .sh-installers is to include but also rename the license file to LICENSE.txt,
+# but for macOS the name of the provided license file remains unchanged.
```
marcoesters:
```suggestion
# but for .pkg the name of the provided license file remains unchanged.
```

news/1085-include-installer-license-file
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* The license file is now also included in the installation directory for macOS package (.pkg) installers.
```
marcoesters:
```suggestion
* Include the license file in PKG installers. (#1074 via #1085)
```

constructor/preconda.py
```diff
@@ -328,18 +328,25 @@ def copy_extra_files(
...
-            copied.append(shutil.copy(path, workdir))
+            orig_path = check_file_exists(path)
+            copied.append(shutil.copy(orig_path, workdir))
         elif isinstance(path, dict):
             assert len(path) == 1
             origin, destination = next(iter(path.items()))
-            orig_path = Path(origin)
-            if not orig_path.exists():
-                raise FileNotFoundError(f"File {origin} does not exist.")
+
```
marcoesters:
```suggestion
```

Let's keep diffs small by not adding additional blank lines where they're not necessary.

constructor/shar.py
```diff
@@ -187,10 +187,13 @@ def create(info, verbose=False):
 
     tarball = join(tmp_dir, "pkgs", "tmp.tar")
     t = tarfile.open(tarball, "w")
+
```
marcoesters:
```suggestion
```

constructor/shar.py
marcoesters:
```suggestion
```

constructor/shar.py
```diff
@@ -187,10 +187,13 @@ def create(info, verbose=False):
...
     t.add(preconda_tarball, basename(preconda_tarball))
     t.add(postconda_tarball, basename(postconda_tarball))
+
     if "license_file" in info:
         t.add(info["license_file"], "LICENSE.txt")
+
```
marcoesters:
```suggestion
```

constructor/preconda.py
lrandersson: Thanks for catching that, will make sure I do that in the future.
I've applied all the changes from the first review, see 436cae7.
For some reason it doesn't work to "Add suggestion to batch", even when I go to "File mode" as the UI suggests.

examples/extra_files/test_install.sh
lrandersson: Fully agree, I changed all of them.

constructor/preconda.py
marcoesters: You may have to switch between the old and new UI to accept suggestions.

Issue #1070, opened 2025-09-22
Several confusing and misleading statements in macOS pkg installer


mhsmith:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

The Miniconda3 py313_25.7.0-2 installer includes the following on its Read Me page:

> By default, this installer modifies your bash profile to activate the base environment of Miniconda3 when your shell starts up.

It actually modifies a lot more than that:
```
modified      /Users/msmith/.bash_profile
modified      /Users/msmith/.config/fish/config.fish
modified      /Users/msmith/.tcshrc
modified      /Users/msmith/.xonshrc
modified      /Users/msmith/.zshrc
```

> If you decline this option, the executables installed by this installer will not be available on PATH.  You will need to use the full executable path to run commands, or otherwise initialize the base environment of Miniconda3 on your own.

If I was unfamiliar with Conda, I would probably want to decline this rather than risk it messing up my shell startup files. However, it's very difficult to work out how to do this: you need to go through the next three pages of the installer and then click the Customize button at the bottom. It would be helpful for the Readme to explain this.

Also, "initialize" suggests I might actually have to *create* the base environment on my own, which isn't the case. "Activate" would be a better word.

> The "Install for me only" option will install Miniconda3 to the default location, ~/miniconda3.

There is no "Install for me only" option, only "Install for all users of this computer" and "Install on a specific disk".

"Install on a specific disk" does not default to ~/miniconda3, but to /opt/miniconda3, which is hidden by default in Finder.

In fact, it seems to be impossible to install to ~/miniconda3 even with a custom location. If I select /Users/msmith, then it installs to /Users/msmith/opt/miniconda3. If I create /Users/msmith/miniconda3 as a new folder and select it, then it installs to /Users/msmith/miniconda3/miniconda3.

marcoesters:
The issues have been partially addressed, but there is one outstanding:

> There is no "Install for me only" option, only "Install for all users of this computer" and "Install on a specific disk".

This is a decision by the installer provider, not `constructor`. There two options:

1. template the readme to account for the domains and the default installation locations (both are set in `construct.yaml`).
2. allow installer providers to write their own readme header.

PR #1088, opened 2025-10-15
Update schema default value and readme

lrandersson:
### Description
1. Based on https://github.com/conda/constructor/issues/1070 this PR updates the default value in the schema to properly reflect what the documentation says.
2. Moreover, a minor change to the readme based on the comments in the same PR.

### Checklist - did you ...

### DIFF
--- a/news/1088-update-schema-default-value
+++ b/news/1088-update-schema-default-value
@@ -4,7 +4,7 @@
 
 ### Bug fixes
 
-* The default value for `enable_currentUserHome` has been changed from `false` to `true` to properly reflect the documentation. This change only affects `.pkg` installers.
+* PKG: Restore the default value of `enable_currentUserHome` to the old default value (`true`). (#1070 via #1088)
 
 ### Deprecations
 
@@ -12,7 +12,7 @@
 
 ### Docs
 
-* The displayed text during installation of `.pkg` installers has been updated to clarify affected shells with default installer behavior.
+* PKG: Clarify that the profile of all available shells will be modified by default. (#1070 via #1088)
 
 ### Other
 


news/1088-update-schema-default-value
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* The default value for `enable_currentUserHome` has been changed from `false` to `true` to properly reflect the documentation. This change only affects `.pkg` installers.
```
marcoesters:
```suggestion
* PKG: Restore the default value of `enable_currentUserHome` to the old default value (`true`). (#1070 via #1088)
```

Imperative mood is preferred overall. I would also like to clarify that this used to be the default value, it's not just a mishap in the documentation.

news/1088-update-schema-default-value
```diff
@@ -0,0 +1,19 @@
...
+
+### Deprecations
+
+* <news item>
+
+### Docs
+
+* The displayed text during installation of `.pkg` installers has been updated to clarify affected shells with default installer behavior.
```
marcoesters:
```suggestion
* PKG: Clarify that the profile of all available shells will be modified by default. (#1070 via #1088)
```

More specific that way.

marcoesters (commented):
The code is good, but the change log could be more specific.

news/1088-update-schema-default-value
lrandersson: Great ideas, applied the suggestion.

news/1088-update-schema-default-value
lrandersson: I agree. I applied the suggestion.

marcoesters (approved):

Issue #1004, opened 2025-06-17
Inconsistencies in `initialize_conda` / `initialize_by_default` handling across installers


jaimergp:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

First I noticed that the Python registration always run even when `register_python` is False in NSIS installers.

The default value for the NSIS variable might end up being `BST_CHECKED`, and this is not cross-checked with `register_python` and `register_python_default` configurations.

Initialization:

https://github.com/conda/constructor/blob/3026fd11c2aa524ab1bee3bf655486f01fd4a361/constructor/nsis/OptionsDialog.nsh#L30-L53

Parsing:

https://github.com/conda/constructor/blob/3026fd11c2aa524ab1bee3bf655486f01fd4a361/constructor/nsis/main.nsi.tmpl#L329-L337

Check with `register_python`:

https://github.com/conda/constructor/blob/3026fd11c2aa524ab1bee3bf655486f01fd4a361/constructor/nsis/main.nsi.tmpl#L717-L719

Execution:

https://github.com/conda/constructor/blob/3026fd11c2aa524ab1bee3bf655486f01fd4a361/constructor/nsis/main.nsi.tmpl#L1449-L1465

---

It feels like we should have a better logic for these knobs. It's different from the other shell installers too. In practical terms, we need to answer these questions:

- `initialize_conda` and `register_python` should control whether these options are _shown_ to the user, or _enabled_ in the installer, or _both_? Right now the behavior varies across installers. On SH installers, they control both the UI/UX knobs plus the actual logic. If false, they are not presented nor processed. However, in EXE installers, it's way more complex as shown above, and the logic might get executed under certain conditions.
- If `initialize_conda` and `register_python` only control the UI/UX, we will allow that logic to run without user interaction. Right now this default value is _not_ controlled by `initialize_by_default` or `register_python_default` in some cases.
- Do we want to allow installer creators to force the feature onto users without them being able to switch those options off? Like, if I want my installer to forcibly add `Scripts/` to PATH and I don't want anyone to do that... is that a good idea?

Overall, I propose to simplify this down to:

- `initialize_conda` and `register_python` should follow the SH installer behavior. If false, those actions should not show up in the installer at all. No UI/UX, no logic, no code.
- `initialize_by_default` and `register_python_default` should only control the default value when the above settings are true. They should have no effect otherwise.

| initialize_conda | initialize_by_default | effect |
[...]

Issue #1003, opened 2025-06-17
NSIS: `/AddToPath` and `/RegisterPython` always available


jaimergp:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

I noticed these CLI options are always available and parsed, regardless the value of `initialize_conda` and `register_python`. This means they show up in the help message and can induce parsing errors even when they should be totally ignored.

PR #1089, opened 2025-10-16
Remove custom python commands from EXE installers

### LINKED ISSUE #549: RFC: Undoing constructor / conda-standalone cross "bundling"

marcoesters:
### Description

Remove custom Python commands for `conda-standalone` versions that support the `windows` subcommand (see https://github.com/conda/conda-standalone/pull/197). This removes the need to have `python` as a required dependency for newer `conda-standalone` versions, so I only added it where it may still be needed.

Note that I am using a fairly blunt instrument for the Python-specific parts of the EXE installer. A more fine-tuned way should be implemented with #1003 and #1004.

Closes #549 

### Checklist - did you ...

### DIFF
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -59,6 +59,8 @@ var /global StdOutHandleSet
 !include "x64.nsh"
 
 !include "FileFunc.nsh"
+!include "StrFunc.nsh"
+${Using:StrFunc} StrStr
 !insertmacro GetParameters
 !insertmacro GetOptions
 
@@ -135,18 +137,22 @@ var /global InstMode # 0 = Just Me, 1 = All Users.
 !define ALL_USERS 1
 
 var /global CMD_EXE
+var /global ICACLS_EXE
 
-!macro FindCmdExe
+!macro FindWindowsBinaries
     # Find cmd.exe
     ReadEnvStr $R0 SystemRoot
     ReadEnvStr $R1 windir
     ${If} ${FileExists} "$R0"
         StrCpy $CMD_EXE "$R0\System32\cmd.exe"
+        StrCpy $ICACLS_EXE "$R0\System32\icacls.exe"
     ${ElseIf} ${FileExists} "$R1"
         StrCpy $CMD_EXE "$R1\System32\cmd.exe"
+        StrCpy $ICACLS_EXE "$R1\System32\icacls.exe"
     ${Else}
-        # Cross our fingers CMD is in PATH
+        # Cross our fingers binaries are in PATH
         StrCpy $CMD_EXE "cmd.exe"
+        StrCpy $ICACLS_EXE "icacls.exe"
     ${EndIf}
 !macroend
 
@@ -1292,6 +1298,78 @@ FunctionEnd
 {%- endif %}
 !macroend
 
+!macro setInstdirPermissions
+    # To address CVE-2022-26526.
+    # Revoke the write permission on directory "$INSTDIR" for Users. Users are:
+    #   AU - authenticated users (NT AUTHORITY\Authenticated Users)
+    #   BU - built-in (local) users (BUILTIN\Users)
+    #   DU - domain users  (<domain name>\DOMAIN USERS)
+    # This also applies for single-user installations to avoid giving other users
+    # full access on shared drives.
+    ${If} ${UAC_IsAdmin}
+        StrCpy $0 "(AU) (BU) (DU)"
+    ${Else}
+        # Not every directory grants write access to Users (e.g., %USERPROFILE%),
+        # so test whether user groups have the necessary rights.
+        nsExec::ExecToStack '"$ICACLS_EXE" "$INSTDIR"'
+        Pop $R0
+        Pop $R1
+        ${If} $R0 != "0"
+          StrCpy $R1 \
+              "Unable to determine the defaults permissions of the installation directory. "\
+              "Ensure that you have read access to $INSTDIR and icacls.exe is in your PATH."
+          ${Print} $R1
+          Abort
+        ${EndIf}
+        StrCpy $0 ""
+        StrCpy $R2 "NT AUTHORITY\Authenticated Users|BUILTIN\Users|\Domain Users"
+        StrCpy $R3 "(AU)|(BU)|(DU)"
+        StrCpy $R4 1
+        loop_single_user_default_access:
+            ${WordFind} $R2 "|" "E+$R4" $R5
+            ${WordFind} $R3 "|" "E+$R4" $R6
+            IfErrors endloop_single_user_default_access
+            ${StrStr} $R7 $R1 $R5
+            ${If} $R7 == ""
+                goto increment_loop_single_user_default_access
+            ${EndIf}
+            # If the user group has a deny permission directive, do not change permissions.
+            # Granting (RX) permissions may increase the permissions that are inherited and
+            # it is very unlikely that a user is granted write permissions but denied others.
+            ${StrStr} $R7 $R1 "$R5:(D)"
+            ${If} $R7 != ""
+                goto increment_loop_single_user_default_access
+            ${EndIf}
+            StrCpy $0 "$0 $R6"
+            increment_loop_single_user_default_access:
+            IntOp $R4 $R4 + 1
+            goto loop_single_user_default_access
+        endloop_single_user_default_access:
+    ${EndIf}
+    AccessControl::DisableFileInheritance "$INSTDIR"
+    ${If} $0 != ""
+        StrCpy $1 1
+        loop_access:
+            ${WordFind} $0 " " "E+$1" $2
+            IfErrors endloop_access
+            AccessControl::RevokeOnFile "$INSTDIR" "$2" "GenericWrite"
+            AccessControl::SetOnFile "$INSTDIR" "$2" "GenericRead + GenericExecute"
+            IntOp $1 $1 + 1
+            goto loop_access
+        endloop_access:
+    ${EndIf}
+    ${IfNot} ${UAC_IsAdmin}
+        # Ensure that creator has full access
+        ReadEnvStr $R0 USERDOMAIN
+        ReadEnvStr $R1 USERNAME
+        ${If} $R0 == ""
+          AccessControl::SetOnFile "$INSTDIR" "$R1" "FullAccess"
+        ${Else}
+          AccessControl::SetOnFile "$INSTDIR" "$R0\$R1" "FullAccess"
+        ${EndIf}
+    ${EndIf}
+!macroend
+
 # Installer sections
 Section "Install"
     ${LogSet} on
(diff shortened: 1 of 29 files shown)


constructor/main.py
```diff
@@ -76,6 +81,24 @@ def get_output_filename(info):
...
 
+def _base_needs_python(info: dict[str, Any]) -> bool:
+    if sys.platform == "win32" and info.get("_conda_exe_type") == StandaloneExe.CONDA:
+        conda_exe = info["_conda_exe"]
+        results = subprocess.run(
+            [conda_exe, "constructor", "windows", "--help"],
+            capture_output=True,
+            check=False,
+        )
+        if results.returncode == 2:
```
lrandersson: For clarity, could we add a comment here explaining what returncode `2` means?

constructor/winexe.py
```diff
@@ -225,16 +225,26 @@ def make_nsi(
...
+    for dist in dists:
+        py_name, py_version, _ = filename_dist(dist).rsplit("-", 2)
+        if py_name == "python":
+            variables["has_python"] = True
+            variables["pyver_components"] = py_version.split(".")
+            break
+
+    if variables["has_python"]:
+        variables["register_python"] = info.get("register_python", True)
+        variables["register_python_default"] = info.get("register_python_default", None)
```
lrandersson, line 240: I'm just curious, why is `register_python_default` typed as `bool | None` but by default `False`? Does `None` lead to any different behavior compared to `False`? To my understanding it seems like it could be a strict boolean and get rid of `None`.

marcoesters: The way these flags are handled needs to be changed. See the issues I linked in the PR description. This is out of scope for this PR though.

constructor/main.py
```diff
@@ -76,6 +81,24 @@ def get_output_filename(info):
...
+            return True
+    specs = {MatchSpec(spec) for spec in info.get("specs", ())}
+    # mamba 2 and newer does not need Python. However, without running the solver,
```
jaimergp: We should assume we'd get `mamba 2.x` here. We can add an additional post-solve check to make sure we didn't use `mamba 1.x`.

constructor/nsis/main.nsi.tmpl
```diff
@@ -317,8 +321,13 @@ FunctionEnd
                 Install for all users, but don't add to PATH env var:$\n\
                     > $EXEFILE /InstallationType=AllUsers$\n\
                 $\n\
+{%- if has_python %}
                 Install for just me, add to PATH and register as system Python:$\n\
                     > $EXEFILE /RegisterPython=1 /AddToPath=1$\n\
+{%- else %}
+                Install for just me and add to PATH:$\n\
+                    > $EXEFILE /AddToPath=1$\n\
+{%- endif %}
```
jaimergp: Does this option depend on having Python or not? I'd say the are independent:

```suggestion
{%- endif %}
                Install for just me and add to PATH:$\n\
                    > $EXEFILE /AddToPath=1$\n\
```

constructor/winexe.py
```diff
@@ -225,16 +225,26 @@ def make_nsi(
...
+        default_uninstall_name += " (Python ${PYVERSION} ${ARCH})"
+    else:
+        variables["register_python"] = False
```
jaimergp, line 243: Should we also define `register_python_default` here for completeness?

constructor/nsis/main.nsi.tmpl
```diff
@@ -1235,6 +1248,36 @@ FunctionEnd
...
 
+{% if not base_needs_python %}
+!macro AddRemovePath add_remove un
+{%-     if initialize_conda == 'condabin' %}
+    push '"$INSTDIR\_conda.exe" constructor windows path --${add_remove}=user --prefix "$INSTDIR\condabin"'
+    push 'Failed to add condabin to PATH'
+    push 'WithLog'
+    call ${un}AbortRetryNSExecWait
+{%-     else %}
+    StrCpy $R0 "Library\mingw-w64\bin Library\usr\bin Library\bin Scripts"
```
jaimergp: I wish we didn't need to hardcode this here. Wouldn't it be more adequate to have this logic in conda-standalone too?

constructor/nsis/main.nsi.tmpl
```diff
@@ -1448,20 +1494,32 @@ Section "Install"
...
-        push '"$INSTDIR\pythonw.exe" -E -s "$INSTDIR\Lib\_nsis.py" addpath ${PYVERSION} ${NAME} ${VERSION} ${ARCH}'
+            ${Print} "Adding $INSTDIR\Scripts & Library\bin to PATH..."
+            push '"$INSTDIR\pythonw.exe" -E -s "$INSTDIR\Lib\_nsis.py" addpath ${PYVERSION} ${NAME} ${VERSION} ${ARCH}'
 {%-     endif %}
-        push 'Failed to add {{ NAME }} to PATH'
-        push 'WithLog'
-        call AbortRetryNSExecWait
+            push 'Failed to add {{ NAME }} to PATH'
+            push 'WithLog'
+            call AbortRetryNSExecWait
```
jaimergp: Why do we implement this twice? Can't we just depend on the new NSIS macro?

marcoesters: Are you suggesting that I move the `if base_needs_python` logic into the macro instead?

constructor/nsis/main.nsi.tmpl
marcoesters: It doesn't depend on Python, no, but I figured I'd not add an extra line to the output since it already has to be broken apart.

constructor/nsis/main.nsi.tmpl
marcoesters: It has to be hard-coded somewhere. I made the choice to have an abstract feature in `conda-standalone` and let the installer implementation decide which paths to add. Do you think it's better to have `conda-standalone` do the special-casing?

constructor/nsis/main.nsi.tmpl
jaimergp: From what I can understand, the `AddRemovePath` macro already implements the same behavior as `'"$INSTDIR\pythonw.exe" -E -s "$INSTDIR\Lib\_nsis.py" addcondabinpath | addpath'`, so I'm wondering whether we need to keep both. If the macro is sufficient, then we don't need the Python-based logic, do we?

constructor/nsis/main.nsi.tmpl
jaimergp: My concern is that these directories are defined by convention (although we're trying [to standardize it here](https://github.com/conda/ceps/pull/124) when building packages, so if at some point we decide to put binaries somewhere else in Windows (e.g. [removing the `Library/` bit](https://github.com/conda-forge/conda-forge.github.io/issues/2295)), this hardcoded list might be too far removed from lists defined in other places like `conda` or other CLIs.

A future addition might be an `arm64` implementation of `mingw`, which would need to be added here. 

That said, we can address that in a future refactor. For now let's link to the CEP draft.

constructor/main.py
jaimergp: Agreed

constructor/nsis/main.nsi.tmpl
jaimergp: > it already has to be broken apart.

I don't get this part. Are they mutually exclusive or not?

marcoesters: They are not mutually exclusive, I just didn't want the message to be overly long with examples. I was referring to the comment a few lines above:

```
        # There seems to be a limit to how many chars per ${Print} we can pass.
        # The message will get truncated silently, no errors.
        # That's why we split the help message in two calls.
```

constructor/nsis/main.nsi.tmpl
marcoesters: The macro implements the same behavior using `conda-standalone`. However, that requires the `conda constructor windows` subcommand to be available, which is currently only the case for the canary builds. The python commands must stay available for now to ensure backwards compatibility.

marcoesters: However, your comment makes me think that it's better to move that logic into the macro, too. I will go ahead and do that - it should clear this up a little more.

constructor/nsis/main.nsi.tmpl
marcoesters: I'm a little concerned to leave this to a future refactor. If this gets moved into `conda-standalone`, we have to do yet another version check and adjust the template. I think I would rather implement this in `conda-standalone` then, which is probably much better at architecture detection than NSIS.

constructor/nsis/main.nsi.tmpl
jaimergp: Works for me. Let's leave a comment to deprecate that part once we bump the minimum conda requirement.

constructor/nsis/main.nsi.tmpl
jaimergp: In that case we can split the help message in one more call. I'd rather leave the examples there.

constructor/nsis/main.nsi.tmpl
marcoesters: I refactored it so that everything is in one place. Should be easier to understand now.

constructor/nsis/main.nsi.tmpl
marcoesters: I refactored it so that it uses https://github.com/conda/conda-standalone/pull/215

constructor/winexe.py
marcoesters: Done

constructor/nsis/main.nsi.tmpl
marcoesters: Done - splitting wasn't needed.

constructor/main.py
marcoesters: I found that `mamba 1.x` actually depends on python, so that check isn't necessary. This makes removing the implicit dependency on Python in the `base` environment much simpler.

constructor/main.py
marcoesters: I added a comment explaining what that return code is. This is the code `argparse` returns when the subcommand does not exist.

jaimergp (approved):

PR #1105, opened 2025-11-10
Improve handling of initialize_conda / register_python including its default values

lrandersson:
### Description
This PR is to address:
1. https://github.com/conda/constructor/issues/1003
2. https://github.com/conda/constructor/issues/1004

### Checklist - did you ...

### DIFF
--- a/CONSTRUCT.md
+++ b/CONSTRUCT.md
@@ -526,16 +526,18 @@ See also `initialize_by_default`.
 ### `initialize_by_default`
 
 Default value for the option added by `initialize_conda`. The default is
-true for PKG installers, and false for EXE (GUI and CLI installations) and shell installers. The user
-is able to change the default during interactive installation. NOTE: For Windows,
-`AddToPath` is disabled when `InstallationType=AllUsers`.
+true for PKG installers, and false for EXE and SH shell installers.
+The user is able to change the default during interactive installations.
+Non-interactive installations are not affected by this value: users must explicitly request
+to add to `PATH` via CLI options.
+NOTE: For Windows, `/AddToPath` is disabled when `/InstallationType=AllUsers`.
 
 Only applies if `initialize_conda` is not false.
 
 ### `register_python`
 
 If the installer installs a Python instance, offer the user an option to register the installed Python instance as the
-system's default Python. (Windows only)
+system's default Python. Defaults to `true` for GUI and `false` for CLI installations. (Windows only)
 
 ### `register_python_default`
 
--- a/constructor/data/construct.schema.json
+++ b/constructor/data/construct.schema.json
@@ -750,7 +750,7 @@
         }
       ],
       "default": null,
-      "description": "Default value for the option added by `initialize_conda`. The default is true for PKG installers, and false for EXE (GUI and CLI installations) and shell installers. The user is able to change the default during interactive installation. NOTE: For Windows, `AddToPath` is disabled when `InstallationType=AllUsers`.\nOnly applies if `initialize_conda` is not false.",
+      "description": "Default value for the option added by `initialize_conda`. The default is true for PKG installers, and false for EXE and SH shell installers. The user is able to change the default during interactive installations. Non-interactive installations are not affected by this value: users must explicitly request to add to `PATH` via CLI options. NOTE: For Windows, `/AddToPath` is disabled when `/InstallationType=AllUsers`.\nOnly applies if `initialize_conda` is not false.",
       "title": "Initialize By Default"
     },
     "initialize_conda": {
@@ -1076,7 +1076,7 @@
     },
     "register_python": {
       "default": true,
-      "description": "If the installer installs a Python instance, offer the user an option to register the installed Python instance as the system's default Python. (Windows only)",
+      "description": "If the installer installs a Python instance, offer the user an option to register the installed Python instance as the system's default Python. Defaults to `true` for GUI and `false` for CLI installations. (Windows only)",
       "title": "Register Python",
       "type": "boolean"
     },
--- a/docs/source/construct-yaml.md
+++ b/docs/source/construct-yaml.md
@@ -526,16 +526,18 @@ See also `initialize_by_default`.
 ### `initialize_by_default`
 
 Default value for the option added by `initialize_conda`. The default is
-true for PKG installers, and false for EXE (GUI and CLI installations) and shell installers. The user
-is able to change the default during interactive installation. NOTE: For Windows,
-`AddToPath` is disabled when `InstallationType=AllUsers`.
+true for PKG installers, and false for EXE and SH shell installers.
+The user is able to change the default during interactive installations.
+Non-interactive installations are not affected by this value: users must explicitly request
+to add to `PATH` via CLI options.
+NOTE: For Windows, `/AddToPath` is disabled when `/InstallationType=AllUsers`.
 
 Only applies if `initialize_conda` is not false.
 
 ### `register_python`
 
 If the installer installs a Python instance, offer the user an option to register the installed Python instance as the
-system's default Python. (Windows only)
+system's default Python. Defaults to `true` for GUI and `false` for CLI installations. (Windows only)
 
 ### `register_python_default`
 


constructor/nsis/main.nsi.tmpl
```diff
@@ -1532,25 +1544,28 @@ Section "Install"
...
-        push '"$INSTDIR\pythonw.exe" -E -s "$INSTDIR\Lib\_nsis.py" addpath ${PYVERSION} ${NAME} ${VERSION} ${ARCH}'
-{%-     endif %}
-        push 'Failed to add {{ NAME }} to PATH'
-        push 'WithLog'
-        call AbortRetryNSExecWait
-    ${EndIf}
-{%- endif %}
+    !if ${INIT_CONDA_OPTION} == 1
+        # For consistency with other code, only for "Just Me" installation
+        ${If} $InstMode = ${JUST_ME}
```
lrandersson: Note that `${If} $InstMode = ${JUST_ME}` is new (the option was still correctly disabled but this code should also account for this, correct?), but seems like it was missing based on the comment in `OptionsDialog.nsh`? I quote:
```
# AddToPath is only an option for JustMe installations; it is disabled for AllUsers
```

constructor/nsis/main.nsi.tmpl
```diff
@@ -349,14 +355,18 @@ FunctionEnd
     ${EndIf}
 
     ClearErrors
-    ${GetOptions} $ARGV "/RegisterPython=" $ARGV_RegisterPython
```
lrandersson, line 362: The old code (and not the updated one - yet) does not reflect what the docstring says:
```
/RegisterPython=[0|1] [default: AllUsers: 1, JustMe: 0]
```
i.e. the default value was never set based on scope. Should I update the code to reflect this? Or should we always just default it to 0?

news/1105-improve-handling-of-certain-options
```diff
@@ -0,0 +1,20 @@
+### Enhancements
+
+* EXE: Improve handling of options `initialize_conda`, `register_python` with their corresponding default values. The behavior of these options
+  with respect to `initialize_conda_default` and `register_python_default` is now consistent with `.sh` and `.pkg` installers.
```
marcoesters:
```suggestion
* EXE: Improve handling of options `initialize_conda`, `register_python` with their corresponding default values. The behavior of these options
  with respect to `initialize_conda_default` and `register_python_default` is now consistent with `.sh` and `.pkg` installers. (#1003, #1004 via $1105)
```

constructor/nsis/OptionsDialog.nsh
```diff
@@ -28,21 +26,32 @@ Var Ana_PostInstall_Label
 Var Ana_PreInstall_Label
 
 Function mui_AnaCustomOptions_InitDefaults
-    # Initialize defaults
-    ${If} $Ana_AddToPath_State == ""
-        StrCpy $Ana_AddToPath_State ${BST_UNCHECKED}
+    # AddToPath / conda init default
+    !if ${INIT_CONDA_OPTION} == 1
```
marcoesters, line 83: I see you are overall using compile-time checks over Jinja templating. I can see why you're doing it here, but do you generally see an advantage using compile-time checks over Jinja for if-conditions? What about nested if-conditions?

Part of me is hesitant mixing those methods.

constructor/nsis/main.nsi.tmpl
```diff
@@ -302,9 +308,9 @@ FunctionEnd
             -------$\n\
             $\n\
                 /InstallationType=AllUsers [default: JustMe]$\n\
-                /AddToPath=[0|1] [default: 0]$\n\
+                /AddToPath=[0|1] [default: 0] ('initialize_conda' must not be false)$\n\
```
marcoesters: This is a user-facing options, so we shouldn't be using `construct.yaml` specifics here (since the user can't influence it). You may need to use Jinja here to make these options appear or disappear.

The default values also need to be adjusted to based on `initialize_by_default`. Same for the Python option.

constructor/nsis/main.nsi.tmpl
marcoesters: The help text should be adjusted to what the code has been doing all along.

constructor/nsis/OptionsDialog.nsh
```diff
@@ -167,25 +213,36 @@ FunctionEnd
...
+    ${If} $1 == ${BST_CHECKED}
+        StrCpy $INIT_CONDA 1
+    ${Else}
+        StrCpy $INIT_CONDA 0
+    ${EndIf}
+
     ShowWindow $Ana_AddToPath_Label ${SW_HIDE}
-    ${NSD_GetState} $0 $Ana_AddToPath_State
-    ${If} $Ana_AddToPath_State == ${BST_UNCHECKED}
+    ${If} $INIT_CONDA == 1
```
marcoesters, line 279: The original coloring was actually correct. Adding `condabin` to `PATH` is pretty harmless. It should only be red when the mode is `classic`.

constructor/nsis/main.nsi.tmpl
```diff
@@ -578,6 +593,19 @@ Function .onInit
     Push $R1
     Push $R2
 
+    # Assign the runtime variables
```
marcoesters: This is one example where Jinja would lead to more compact code.

constructor/nsis/OptionsDialog.nsh
```diff
@@ -14,8 +14,6 @@ Var mui_AnaCustomOptions.ClearPkgCache
 Var mui_AnaCustomOptions.CreateShortcuts
 
 # These are the checkbox states, to be used by the installer
-Var Ana_AddToPath_State
-Var Ana_RegisterSystemPython_State
```
marcoesters, line 18: Why did you decide to replace those? From what I can see, your new global variables do the same thing. The defaults just need to be initiated properly.

news/1105-improve-handling-of-certain-options
marcoesters: Did you find any breaking changes this introduces? If so, this should be highlighted. If not, feel free to resolve this comment.

marcoesters (commented):
My initial set of comments.

constructor/nsis/main.nsi.tmpl
marcoesters: It is checked in the python code that's called, but I think I found the same thing. The code in `main` should have a check like that already.

constructor/nsis/main.nsi.tmpl
lrandersson: It's now `/RegisterPython=[0|1] [default: 0]`

constructor/nsis/main.nsi.tmpl
lrandersson: I'd say the gain is minimal, depending on what is of priority, but see my answer below (regarding choosing between nsis/jinja)

constructor/nsis/main.nsi.tmpl
lrandersson: If I understand correctly, the additional `${If} $InstMode = ${JUST_ME}` I added is redundant and can safely be removed?
Update:
I saw when resolving the merge conflicts that you added `${If} ${FileExists} "$INSTDIR\.nonadmin"` so I have removed the `${If} $InstMode = ${JUST_ME}`

constructor/nsis/OptionsDialog.nsh
lrandersson: I decided to more or less replace them because:
1. The new variable names (I think) better reflect their purpose (since they are tied to the options)
2. Have the variables defined on top (in `main.nsi.tmpl`), since they have a scope for UI-based and non UI-based installers. And to have them declared early together with the `!define ...` that is using the jinja templating.

constructor/nsis/OptionsDialog.nsh
lrandersson: In general I think the compile-time checks are better to use for features that work well in NSIS, and if-conditions are basic enough I don't see much gain for a developer to switch to jinja unless necessary.
If you have to at some point also debug the generated template, there is less code that is affected by the jinja templating (since everything else is static).
To me at least, if the jinja-templating is not strictly necessary, it is easier to understand the code if we have code structured as:
```
< do necessary jinja templating first, and define variables that depend on it >

<regular NSIS code>
```
The advantage is that you can see the effect of for example `{% if initialize_conda %}` directly and how it affects the template.

Nesting works well with `!macro` and LogicLib. You can guard macro expansions accurately.

But I am probably also a bit biased since I am used to NSIS without combining it with jinja.

constructor/nsis/OptionsDialog.nsh
lrandersson: Ensured its red also within the `*_OnClick` function

constructor/nsis/main.nsi.tmpl
```diff
@@ -304,11 +310,13 @@ FunctionEnd
...
             $\n\
                 /InstallationType=AllUsers [default: JustMe]$\n\
+{% if initialize_conda %}
                 /AddToPath=[0|1] [default: 0]$\n\
+{% endif %}
                 /KeepPkgCache=[0|1] [default: {{ 1 if keep_pkgs else 0 }}]$\n\
-{%- if has_python %}
-                /RegisterPython=[0|1] [default: AllUsers: 1, JustMe: 0]$\n\
-{%- endif %}
+{% if has_python and register_python %}
```
lrandersson: @marcoesters I resolved the merge conflict, but is `has_python` necessary here if I guard it with simply `{% if register_python %}`?

news/1105-improve-handling-of-certain-options
lrandersson: I didn't find any breaking changes.

constructor/nsis/OptionsDialog.nsh
marcoesters: > If you have to at some point also debug the generated template, there is less code that is affected by the jinja templating (since everything else is static).

That depends on what you look at. The line numbers in the error messages are closer to the template, but the generated `main.nsi` still contains the `!if` lines, doesn't it? Rendering with Jinja would make those disappear. So, you get a "pre-compile time" look, which is less code to look at.

> To me at least, if the jinja-templating is not strictly necessary, it is easier to understand the code if we have code structured as:

Point taken, even though the for-loops would still be Jinja since for-loops are not straight-forward in NSIS.

> if-conditions are basic enough I don't see much gain for a developer to switch to jinja unless necessary.

I would agree with that statement we were switching to Jinja - but that's not what's happening in this PR. The PR is switching away from Jinja and introduces two different ways to create quasi compile-time if-conditions. Are you suggesting we should move all jinja-based if-conditions to NSIS-coded compile time statements in a future PR?

constructor/nsis/OptionsDialog.nsh
marcoesters: The same could be said about the other state variables. In isolation, I don't disagree with you. It's the inconsistency this introduces that gives me pause.

constructor/nsis/main.nsi.tmpl
marcoesters: I think the issue starts earlier: `register_python` doesn't make sense if `has_python` is `False`. There needs to be some logic to ensure that `register_python` isn't set to `True` if Python isn't shipped with the installer.

constructor/nsis/OptionsDialog.nsh
lrandersson: Yeah I can see that. In general I'd be in favor for renaming all of them, I think the `Ana_*` is unnecessary and seems like a legacy thing.

constructor/nsis/OptionsDialog.nsh
lrandersson: I agree, some logic (like for-loops) are pretty bad in NSIS.
However, I'd say what I've added in this PR is not new, for example we have a mix of these kind of if-statements in the code already since before, some examples:
1. `!if '{{ SIGNTOOL_COMMAND }}' != ''`
2. Plenty of `${If} ...`
3. Then we have the `!if` and `!ifndef` and more of the compile-time checks.

> Are you suggesting we should move all jinja-based if-conditions to NSIS-coded compile time statements in a future PR?

I'd say yes, but that depends a bit on below. I think in general there is a bit of overlap between what NSIS and jinja can provide:
1. Why was jinja originally introduced and is it solving a problem that NSIS could not?
2. It might not be that worth if `.msi` is going to replace this in the near future.
But if the changes above are outside of the guidelines of how we want the NSIS code formatted, let me know and I will adjust it.

constructor/nsis/main.nsi.tmpl
lrandersson: I see. When I wrote this I had not though of the possibility of no Python at all, and especially since the recent PR just introduced `has_python`. Leave this thread open and I will fix that when we get through the other open threads.

marcoesters: The requirement that a `base` environment must have Python is artificial and was particularly due to NSIS string length limitations. Specifically, the `PATH` environment variable is too long so that Python had to be used to perform some installer functions. Recent PRs aimed at eliminating that dependency so that Python is not required anymore.

constructor/nsis/OptionsDialog.nsh
marcoesters: > Why was jinja originally introduced and is it solving a problem that NSIS could not?

Because it was better than the custom templating solution that was there before.

> It might not be that worth if .msi is going to replace this in the near future.

I think EXE will stick around for a while and needs to be supported.

We don't have that many `!if*def` or `!if` instances, certainly much less than Jinja. If you think it's worth replacing Jinja for if-conditions, go ahead and create an issue.

constructor/nsis/OptionsDialog.nsh
marcoesters: If you wish to tackle that task, go ahead and create an issue so that this can be tracked. I'm not partial to any solution, I'd just like to see consistency.

constructor/nsis/main.nsi.tmpl
```diff
@@ -358,16 +368,18 @@ FunctionEnd
...
-        ${ElseIf} $ARGV_RegisterPython = "0"
-            StrCpy $Ana_RegisterSystemPython_State ${BST_UNCHECKED}
+    !if ${REGISTER_PYTHON_OPTION} == 1
+        ${GetOptions} $ARGV "/RegisterPython=" $ARGV_RegisterPython
+        ${IfNot} ${Errors}
+            ${If} $ARGV_RegisterPython == "1"
+                StrCpy $REG_PY 1
+            ${ElseIf} $ARGV_RegisterPython == "0"
+                StrCpy $REG_PY 0
+            ${EndIf}
```
marcoesters, line 381: Looking at #1004 and your help text, this should default to `0` in the CLI, just like the SH installer. As it stands, it defaults to the default value. So, we need an else clause here for when that option has not been found. Some with the conda initialization.

constructor/nsis/OptionsDialog.nsh
```diff
@@ -28,21 +26,32 @@ Var Ana_PostInstall_Label
...
 Function mui_AnaCustomOptions_InitDefaults
-    # Initialize defaults
-    ${If} $Ana_AddToPath_State == ""
-        StrCpy $Ana_AddToPath_State ${BST_UNCHECKED}
+    # AddToPath / conda init default
+    !if ${INIT_CONDA_OPTION} == 1
+        # Ensure we initialize from compile-time default value
+        StrCpy $INIT_CONDA ${INIT_CONDA_DEFAULT_VALUE}
+    !else
+        StrCpy $INIT_CONDA 0
```
marcoesters, line 87: We need to update the documentation in the schema that the default values have no effect if `initialize_conda` and `register_python` are `false`. Maybe also add a sentence or two into the release notes for this since this is a behavior change?

constructor/nsis/OptionsDialog.nsh
lrandersson: https://github.com/conda/constructor/issues/1110

constructor/nsis/OptionsDialog.nsh
lrandersson: https://github.com/conda/constructor/issues/1109

constructor/nsis/OptionsDialog.nsh
lrandersson: So for `initialize_by_default` it says already:
_Only applies if `initialize_conda` is not false._

And for `register_python_default`:
_Only applies if `register_python` is true._

Are you thinking of anything else in addition to this?

constructor/nsis/main.nsi.tmpl
lrandersson: This has been fixed

constructor/nsis/OptionsDialog.nsh
marcoesters: Ah, I didn't see those isolated paragraphs, my bad. The release notes should be updated though since this is new for Windows. The documentation should be updated either way to emphasize that these values only apply to interactive installations and are `false` for CLI installations.

constructor/nsis/main.nsi.tmpl
lrandersson: Ahh, thanks now I understand that better. I updated the code to account for this here as well:
https://github.com/conda/constructor/pull/1105/commits/bc6a315803e4ceb0aa3d9763689347e1fd5d7713

constructor/nsis/main.nsi.tmpl
lrandersson: If I'm not missing anything it does default to zero, however I have made it more clear (and safe?) in this commit https://github.com/conda/constructor/pull/1105/commits
I also noticed another doc-page that claimed that `1` is default so I changed that in the same commit.

constructor/nsis/OptionsDialog.nsh
lrandersson: Updated again in https://github.com/conda/constructor/pull/1105/commits/db68a7c3cdeaeaf53a0d832b100eb3bd666ca424
It got pretty long, please leave some feedback. I'm also trying to stay consistent with the existing style of the release notes.

constructor/nsis/main.nsi.tmpl
marcoesters: I think us coming to different conclusions shows that it's necessary to be safe here 😅 Can you initialize `$INIT_CONDA` the same way?

news/1105-improve-handling-of-certain-options
```diff
@@ -0,0 +1,22 @@
+### Enhancements
+
+* EXE: Improve handling of options `initialize_conda`, `register_python` with their corresponding default values. The behavior of these options
+  with respect to `initialize_by_default` and `register_python_default` is now consistent with `.sh` and `.pkg` installers.
+  The options `initialize_by_default` and `register_python_default` only applies for interactive installations.
+  Note that when installing via the command-line, the corresponding options `AddToPath` and `RegisterPython` are by default set to `0`. (#1003, #1004 via #1105)
```
marcoesters:
```suggestion
  Windows CLI installations now don't `conda` to `PATH` or register Python by default. (#1003, #1004 via #1105)
```

constructor/nsis/OptionsDialog.nsh
marcoesters: I left some feedback to be more user-oriented. We also need to update the description in the schema so that this is documented in `CONSTRUCT.md`. I think after those small comments are addressed, we're in business.

lrandersson: Okay so I have made a couple of changes, and there is already some text in the documentation I think already explains the situation. I'll try to explain it so we are aligned. First here are the changes from last 2 commits I've just done: https://github.com/conda/constructor/pull/1105/files/db68a7c3cdeaeaf53a0d832b100eb3bd666ca424..bf24d45b2aadbf0010e26118f5b89a380a107d34
1. We had `initialize_by_default: bool | None = None`, but comparing it to `register_python_default` I changed it to `initialize_by_default: bool | None = False`. And to be honest, I think we can drop the entire `None`-situation, `True/False` should suffice but I assume this was implemented as `None` because the option might be irrelevant depending on its parent (`initialize_conda`/`register_python`).
2. It says that `initialize_by_default` is `True` for GUI-installers, I quote: _The default is true for GUI installers (EXE, PKG)_. But this is not correct because in `winexe.py` we can see `variables["initialize_by_default"] = info.get("initialize_by_default", None)`, which means it's by default `None`. In the commit I linked to I also changed `None` to `False` for consistency with the other installers, the impact this has in the NSIS code is the same.
3. For `register_python` I added the additional text "_If the installer bundles a Python instance,_", to account for the situation that `has_python` could be false.

Finally, this leaves us with the following:
```

### `initialize_conda`

Add an option to the installer so the user can choose whether to run `conda init`
after the installation (Unix), or to add certain subdirectories of the installation
to PATH (Windows). Requires `conda` to be part of the `base` environment. Valid options:

- `classic` or `True`: runs `conda init` on Unix, which injects a shell function in the
  shell profiles. On Windows, it adds `$INSTDIR`, `$INSTDIR/Scripts`, `$INSTDIR/Library/bin`
  to `PATH`. This is the default.
- `condabin`: only adds `$INSTDIR/condabin` to `PATH`. On Unix, `conda>=25.5.0` is required
  in `base`.
- `False`: the installer doesn't perform initialization.

See also `initialize_by_default`.

### `initialize_by_default`

Default value for the option added by `initialize_conda`. The default is
true for PKG installers, and false for EXE (GUI and CLI installations) and shell installers. The user
is able to change the default during interactive installation. NOTE: For Windows,
`AddToPath` is disabled when `InstallationType=AllUsers`.

Only applies if `initialize_conda` is not false.

### `register_python`

If the installer installs a Python instance, offer the user an option to register the installed Python instance as the
system's default Python. (Windows only)

### `register_python_default`

Default choice for whether to register the installed Python instance as the
system's default Python. The user is still able to change this during
interactive installation. (Windows only).
```

Let me know what you think.

news/1105-improve-handling-of-certain-options
lrandersson: Added https://github.com/conda/constructor/pull/1105/commits/666c2e353eab27612ccbf3cc9eac8d4c505ef2d4

constructor/nsis/main.nsi.tmpl
lrandersson: I agree, added here https://github.com/conda/constructor/pull/1105/commits/666c2e353eab27612ccbf3cc9eac8d4c505ef2d4

lrandersson: Update: I had to adjust it accordingly https://github.com/conda/constructor/pull/1105/commits/83ed89a904589d62a245530357d7452dfb230ffb
I missed that `Function .onInit` calls `!insertmacro ParseCommandLineArgs` twice. First, early in the function call and then again (is this a mistake?) `Call onInit_Release` -> Which contains `!insertmacro ParseCommandLineArgs`

Therefore, we now have the following initialization:
```
Function .onInit
    ${LogSet} on
    Push $0
    Push $1
    Push $2
    Push $R1
    Push $R2

    # Initialize INIT_CONDA and REG_PY, note that they to default 0 if the installer
    # is not configured to prompt the user with a choice.
    !if ${INIT_CONDA_OPTION} == 1
        StrCpy $INIT_CONDA ${INIT_CONDA_DEFAULT_VALUE}
    !else
        StrCpy $INIT_CONDA 0
    !endif

    !if ${REGISTER_PYTHON_OPTION} == 1
        StrCpy $REG_PY ${REGISTER_PYTHON_DEFAULT_VALUE}
    !else
        StrCpy $REG_PY 0
    !endif
    ...
```

marcoesters: > First, early in the function call and then again (is this a mistake?)

That is a good question. My guess is that the `mui_AnaCustomOptions_InitDefaults` function maybe overwrote some variables, but not the ones that are needed for all the pre-install checks.

> Therefore, we now have the following initialization:

If I understand your changes correctly, this would re-introduce the inconsistency between shell installers and EXE installers. The shell installers do not initialize anything in a CLI installation. The thought behind that is that we do not want to change shell profiles without explicit user consent. Your current initialization would set the default to the value set in `construct.yaml` (which contradicts the current help text where you write that the default is always `0`). The difference is that in EXE installers, users can still opt out though.

I think initializing everything to `0` is the correct way to do this. In the CLI, `ParseCommandLineArgs` will override the default with user input. For GUI installers, `mui_AnaCustomOptions_InitDefaults` will initialize the correct defaults for the GUI. The easiest and cleanest way is probably:

```
${If} ${Silent}
  !insetmacro ParseCommandLineArgs
${Else}
  Call mui_AnaCustomOptions_InitDefaults
${EndIf}
```

That will, of course, mean that `OnInit_Release` cannot insert the command line parsing macro anymore, which I don't think is a bad thing.

constructor/nsis/OptionsDialog.nsh
marcoesters: > And to be honest, I think we can drop the entire None-situation, True/False should suffice but I assume this was implemented as None because the option might be irrelevant depending on its parent

`None` means "not specified in `constructor.yaml`". Since `initialize_by_default` is different between PKG, EXE, and SH, `None` is still necessary.

lrandersson: Changed back here https://github.com/conda/constructor/pull/1105/commits/16ffe6e7fd468e7a7f6b365293ba64bc5f046984.
I feel like `register_python_default` should also be `None` then instead of default `False`? I'm not planning to change it now but it seems like `initialize_by_default` and `register_python_default` should have the same defaults?

lrandersson: See latest commit https://github.com/conda/constructor/pull/1105/commits/a913928ddcce235a745d2bbab7033966d6000dde

constructor/nsis/main.nsi.tmpl
lrandersson: See latest commit https://github.com/conda/constructor/pull/1105/commits/a913928ddcce235a745d2bbab7033966d6000dde

news/1105-improve-handling-of-certain-options
lrandersson: More adjustments in the latest commit https://github.com/conda/constructor/pull/1105/commits/a913928ddcce235a745d2bbab7033966d6000dde

constructor/nsis/main.nsi.tmpl
```diff
@@ -367,18 +369,45 @@ FunctionEnd
...
+                    ${Print} "/AddToPath=1 is disabled and ignored in 'All Users' installations"
+                    StrCpy $INIT_CONDA 0
+                ${Else}
+                    StrCpy $INIT_CONDA 1
+                ${EndIf}
+            ${ElseIf} $ARGV_AddToPath = "0"
+                StrCpy $INIT_CONDA 0
+            ${EndIf}
+        ${Else}
+            # If we have Errors, i.e. the option is not explicitly set by the user
```
```suggestion
            # If we have Errors, the option is not explicitly set by the user
```

news/1105-improve-handling-of-certain-options
```diff
@@ -1,7 +1,9 @@
 ### Enhancements
 
 * EXE: Improve handling of options `initialize_conda`, `register_python` with their corresponding default values. The behavior of these options
-  with respect to `initialize_conda_default` and `register_python_default` is now consistent with `.sh` and `.pkg` installers. (#1003, #1004 via #1105)
+  with respect to `initialize_by_default` and `register_python_default` is now consistent with `.sh` and `.pkg` installers.
+  Windows CLI installations now don't add `conda` to `PATH` or register Python by default, and, command-line arguments are
```
marcoesters:
```suggestion
  Windows CLI installations now don't add `conda` to `PATH` or register Python by default, and command-line arguments are
```

news/1105-improve-handling-of-certain-options
```diff
@@ -1,7 +1,9 @@
...
+  only parsed when installing in silent mode (enabled with the flag `/S`).(#1003, #1004 via #1105)
```
marcoesters:
```suggestion
  only parsed when installing in silent mode (enabled with the flag `/S`). (#1003, #1004 via #1105)
```

constructor/nsis/main.nsi.tmpl
```diff
@@ -367,18 +369,45 @@ FunctionEnd
...
+        ClearErrors
         ${GetOptions} $ARGV "/RegisterPython=" $ARGV_RegisterPython
         ${IfNot} ${Errors}
             ${If} $ARGV_RegisterPython == "1"
                 StrCpy $REG_PY 1
             ${ElseIf} $ARGV_RegisterPython == "0"
                 StrCpy $REG_PY 0
             ${EndIf}
+        ${Else}
+            # If we have Errors, i.e. the option is not explicitly set by the user
```
```suggestion
            # If we have Errors, the option is not explicitly set by the user
```

constructor/nsis/main.nsi.tmpl
```diff
@@ -763,16 +756,11 @@ Function .onInit
         StrCpy $CheckPathLength "1"
     ${EndIf}
 
-    # Initialize the custom options
-    Call mui_AnaCustomOptions_InitDefaults
-
     StrCpy $CheckPathLength "{{ 1 if check_path_length else 0 }}"
     StrCpy $Ana_ClearPkgCache_State {{ '${BST_UNCHECKED}' if keep_pkgs else '${BST_CHECKED}' }}
     StrCpy $Ana_PreInstall_State {{ '${BST_CHECKED}' if pre_install_exists else '${BST_UNCHECKED}' }}
     StrCpy $Ana_PostInstall_State {{ '${BST_CHECKED}' if post_install_exists else '${BST_UNCHECKED}' }}
```
marcoesters, line 762: Interesting that these are not initialized in `mui_AnaCustomOptions_InitDefaults`. Something to add to #1109 or #1110

constructor/_schema.py
```diff
@@ -692,16 +692,16 @@ class ConstructorConfiguration(BaseModel):
...
     is able to change the default during interactive installation. NOTE: For Windows,
     `AddToPath` is disabled when `InstallationType=AllUsers`.
 
     Only applies if `initialize_conda` is not false.
     """
     register_python: bool = True
     """
-    Whether to offer the user an option to register the installed Python instance as the
+    If the installer installs a Python instance, offer the user an option to register the installed Python instance as the
     system's default Python. (Windows only)
```
marcoesters:
```suggestion
    If the installer installs a Python instance, offer the user an option to register the installed Python instance as the
    system's default Python. Defaults to `true` for GUI and `false` for CLI installations. (Windows only)
```

constructor/_schema.py
```diff
@@ -692,16 +692,16 @@ class ConstructorConfiguration(BaseModel):
     """
     initialize_by_default: bool | None = None
     """
-    Default value for the option added by `initialize_conda`. The default
-    is true for GUI installers (EXE, PKG) and false for shell installers. The user
+    Default value for the option added by `initialize_conda`. The default is
+    true for PKG installers, and false for EXE (GUI and CLI installations) and shell installers. The user
     is able to change the default during interactive installation. NOTE: For Windows,
     `AddToPath` is disabled when `InstallationType=AllUsers`.
```
marcoesters:
```suggestion
    Default value for the option added by `initialize_conda`. The default is
    true for PKG installers, and false for EXE and SH shell installers.
    The user is able to change the default during interactive installations.
    Non-interactive installations are not affected by this value: users must explicitly request
    to add to `PATH` via CLI options.
    NOTE: For Windows, `/AddToPath` is disabled when `/InstallationType=AllUsers`.
```

I tried to make this explicit. Let me know what you think.

marcoesters (commented):
I think the code is there - just a few word changes. Please verify that my suggestions caught the spirit of it.

constructor/nsis/main.nsi.tmpl
lrandersson: https://github.com/conda/constructor/pull/1105/commits/cf93b7a7c4346606f9542f766f8c1bd2aa865344

news/1105-improve-handling-of-certain-options
lrandersson: https://github.com/conda/constructor/pull/1105/commits/cf93b7a7c4346606f9542f766f8c1bd2aa865344

news/1105-improve-handling-of-certain-options
lrandersson: https://github.com/conda/constructor/pull/1105/commits/cf93b7a7c4346606f9542f766f8c1bd2aa865344

constructor/nsis/main.nsi.tmpl
lrandersson: https://github.com/conda/constructor/pull/1105/commits/cf93b7a7c4346606f9542f766f8c1bd2aa865344

constructor/nsis/main.nsi.tmpl
lrandersson: I'm glad you mentioned that. I was also giving them a side-eye yesterday, I have updated #1110 to fix this.

constructor/_schema.py
lrandersson: https://github.com/conda/constructor/pull/1105/commits/cf93b7a7c4346606f9542f766f8c1bd2aa865344

constructor/_schema.py
lrandersson: Yeah I think this is a good change. I feel likely in general it's a bit confusing as a user when reading about these two (four) options. We have one option to just make another option visible, and then the actual option to toggle the value. Easy to get confused.
Now with being this explicit it will be easier to understand. I've applied the suggestion in https://github.com/conda/constructor/pull/1105/commits/cf93b7a7c4346606f9542f766f8c1bd2aa865344.

PR #1108, opened 2025-11-11
Improve logging for EXE installers

### LINKED ISSUE #998: EXE: Improve logging of `python.exe` commands
### Checklist

### What is the idea?

Due to limitations with NSIS, EXE installers rely on `python` to perform certain functions during the installation. If these commands fail, the error messages do not provide any details about why they fail.

### Why is this needed?

To enable users to provide better bug reports and improve user experience even with failed installations.

### What should happen?

There should be additional output in the log or in the error message to indicate what the problem is.

### Additional Context

_No response_

marcoesters:
### Description

The logging experience for EXE installers has several problems that make debugging difficult to impossible:

* Log files are written before the installation starts. This can cause `install.log` files to appear inside `%USERPROFILE%` or even cause the installation to fail because the check for empty directories fail due to the log file.
* Output of conda invocations are only visible in the GUI. not in the CLI or in the log file.
* No error messages are visible if calling binaries (like `icacls.exe` or `python.exe`) fails.

This PR addresses these issues by:

* Only enabling logging after `$INSTDIR` has been created.
* Outputting to file in silent mode and writing the content of that file both to log and to the console.
* Executing commands in a `cmd.exe` subshell to output error messages coming from executing the binary (e.g., if it's not found or if there are permission issues). This makes `pythonw.exe` obsolete and outputs errors from the custom Python script.
* Moving registry entry removal to the beginning of the installation. This doesn't just allow these commands to be logged, it also doesn't make users remove these registry entries manually should the installation fail for other reasons.
* Stopping the logging right before the installation directory is removed so that a user doesn't have to reboot. This led me to making "abort" the default choice for `AbortRetryNSExecWait` so that logs are still available if the uninstallation fails (continuing even though steps fail shouldn't be happening in the first place to avoid undefined behavior, in my opinion).

I also decided to remove unused functions, which resulted in `Utils.nsh` only containing the `IsWritable` function. I could move that function into the main template, or I can move the functions inside the `FunctionTemplates` macro and the `Print` macro into `Utils.nsh`.

Caveats:

* The `conda` commands cannot directly write into `install.log` even if they support `--log-file` because NSIS maintains the file handle on the log file and doesn't allow writing to it outside `LogSet`.
* Since `cmd.exe` does not support the `tee` command, there is a trade-off between live printing for GUI installers and logging for system calls. I prioritized getting real time output into the GUI window over writing to the log file.

This also closes #998 

Xref: https://github.com/conda/conda-standalone/pull/218

### Checklist - did you ...

### DIFF
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -1275,7 +1275,9 @@ FunctionEnd
             # When binaries are executed with Exec or ExectToLog, only the output of these
             # commands are logged. If they do not start successfully (e.g., if they don't exist),
             # no error messages are returned. Using a subshell reveals these kinds of issues.
-            StrCpy $R3 '"$CMD_EXE" /D /C "$R3"'
+            # Enable delayed expansion so that error levels are parsed at runtime instead of
+            # parse time.
+            StrCpy $R3 '"$CMD_EXE" /V:ON /D /C "$R3"'
             ${If} $R1 == "WithLog"
                 nsExec::ExecToLog $R3
             ${ElseIf} $R1 == "NoLog"
@@ -1342,7 +1344,7 @@ FunctionEnd
         StrCpy $R1 "Failed to remove {{ NAME }} from PATH"
     ${EndIf}
     # `type` is used to simulate a `tee`-like output in cmd.exe
-    push '"$INSTDIR\python.exe" -E -s "$INSTDIR\Lib\_nsis.py" $R0 > "${STEP_LOG}" 2>&1 & type "${STEP_LOG}"'
+    push '"$INSTDIR\python.exe" -E -s "$INSTDIR\Lib\_nsis.py" $R0 > "${STEP_LOG}" 2>&1 & SET ERR=!ERRORLEVEL! & type "${STEP_LOG}" & EXIT /B !ERR!'
     push $R1
     push 'WithLog'
     call ${un}AbortRetryNSExecWait
@@ -1579,7 +1581,7 @@ Section "Install"
     IfFileExists "$INSTDIR\pkgs\pre_install.bat" 0 NoPreInstall
         ${Print} "Running pre-install script..."
         # `type` is used to simulate a `tee`-like output in cmd.exe
-        push '"$INSTDIR\pkgs\pre_install.bat" > "${STEP_LOG}" 2>&1 & type "${STEP_LOG}"'
+        push '"$INSTDIR\pkgs\pre_install.bat" > "${STEP_LOG}" 2>&1 & SET ERR=!ERRORLEVEL! & type "${STEP_LOG}" & EXIT /B !ERR!'
         push "Failed to run pre-install script."
         push 'WithLog'
         call AbortRetryNSExecWait
@@ -1641,7 +1643,7 @@ Section "Install"
         ${If} $Ana_PostInstall_State = ${BST_CHECKED}
             ${Print} "Running post-install..."
             # `type` is used to simulate a `tee`-like output in cmd.exe
-            push '"$INSTDIR\pkgs\post_install.bat" > "${STEP_LOG}" 2>&1 & type "${STEP_LOG}"'
+            push '"$INSTDIR\pkgs\post_install.bat" > "${STEP_LOG}" 2>&1 & SET ERR=!ERRORLEVEL! & type "${STEP_LOG}" & call exit !ERR!'
             push "Failed to run post-install script."
             push 'WithLog'
             call AbortRetryNSExecWait
@@ -1707,7 +1709,7 @@ Section "Install"
     # Use icacls because it is much faster than custom NSIS solutions.
     # We continue on error because icacls fails on broken links.
     # `type` is used to simulate a `tee`-like output in cmd.exe
-    push '"$ICACLS_EXE" "$INSTDIR\*" /inheritance:e /T /C /Q > "${STEP_LOG}" 2>&1 & type "${STEP_LOG}"'
+    push '"$ICACLS_EXE" "$INSTDIR\*" /inheritance:e /T /C /Q > "${STEP_LOG}" 2>&1 & SET ERR=!ERRORLEVEL! & type "${STEP_LOG}" & EXIT /B !ERR!'
     push 'Failed to enable inheritance for all files in the installation directory.'
     push 'WithLog'
     call AbortRetryNSExecWait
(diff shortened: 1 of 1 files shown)


constructor/nsis/_nsis.py
```diff
@@ -96,7 +79,7 @@ def add_to_path(pyversion, arch):
     except IOError:
         old_prefixes = []
     for prefix in old_prefixes:
-        out('Removing old installation at %s from PATH (if any entries get found)\n' % (prefix))
+        print('Removing old installation at %s from PATH (if any entries get found)\n' % (prefix))
```
jaimergp: If we are switching to `print()`, may as well remove the % formatting.

constructor/nsis/_system_path.py
jaimergp, line 1: What are we using this module for? I thought we put most of this logic in conda-standalone. Is this for micromamba?

constructor/nsis/main.nsi.tmpl
```diff
@@ -17,6 +17,16 @@ Unicode true
 !define LogSet "!insertmacro LogSetMacro"
 !macro LogSetMacro SETTING
   !ifdef ENABLE_LOGGING
+    ${If} ${SETTING} == "on"
+      ${IfNot} ${FileExists} "$INSTDIR\install.log"
+        # Enforce UTF-16 encoding in the log file
+        # NSIS doesn't write the correct BOM to log files,
+        # so each character will be followed by NUL.
+        FileOpen $R0 "$INSTDIR\install.log" w
+        FileWriteUTF16LE $R0 ""
```
jaimergp: Do you know why we need this type of UTF? I remembering being bitten by this oddity.

constructor/nsis/main.nsi.tmpl
```diff
@@ -1212,49 +1230,98 @@ Function un.OnDirectoryLeave
...
         ${Do}
-            ${If} $1 == "WithLog"
-                nsExec::ExecToLog $3
-            ${ElseIf} $1 == "NoLog"
-                nsExec::Exec $3
+            # Execute command inside a subshell to catch issues with command execution.
+            # When binaries are executed with Exec or ExectToLog, only the output of these
+            # commands are logged. If they do not start successfully (e.g., if they don't exist),
+            # no error messages are returned. Using a subshell reveals these kinds of issues.
+            StrCpy $R3 '"$CMD_EXE" /D /C "$R3"'
```
jaimergp: Do we need to be careful about quotes? I think CMD does not need the entire string to be quoted, only the individual arguments (which presumably were quoted on input). IOW:

```suggestion
            StrCpy $R3 '"$CMD_EXE" /D /C $R3'
```

constructor/nsis/main.nsi.tmpl
```diff
@@ -1578,7 +1649,14 @@ Section "Install"
...
         ${If} $Ana_PostInstall_State = ${BST_CHECKED}
             ${Print} "Running post install..."
-            push '"$CMD_EXE" /D /C "$INSTDIR\pkgs\post_install.bat"'
+            # Output into step log if silent so that the output is visible in the console
+            # and the logs. For GUI installers, there is no good way to have live output
+            # and output into logs.
+            ${If} ${Silent}
+                push '"$INSTDIR\pkgs\post_install.bat" > "${STEP_LOG}" 2>&1"'
+            ${Else}
+                push '"$INSTDIR\pkgs\post_install.bat"'
```
jaimergp: Can't we do something like `<command> > output.txt & type output.txt`? As in "output to file, then print the file contents".

constructor/nsis/main.nsi.tmpl
```diff
@@ -1776,11 +1895,11 @@ Section "Uninstall"
...
-    # When running conda.bat directly, there is a non-fatal error
-    # that DOSKEY (called by conda_hook.bat) is not a valid command.
-    # While the operation still succeeds, this error is confusing.
-    # Calling via cmd.exe fixes that.
-    push '"$CMD_EXE" /D /C "$INSTDIR\condabin\conda.bat" init cmd.exe --reverse --$R0'
+    ${If} ${Silent}
+        push '"$INSTDIR\condabin\conda.bat" init cmd.exe --reverse --$R0 > "${STEP_LOG}" 2>&1'
+    ${Else}
+        push '"$INSTDIR\condabin\conda.bat" init cmd.exe --reverse --$R0'
+    ${EndIf}
```
jaimergp: We don't need the CMD workaround anymore here because all commands are run via CMD anyway, right?

constructor/nsis/_system_path.py
marcoesters: Yes, most of the logic is in `conda-standalone` (at least on `main`). However, this requires a `conda-standalone` version that supports the new logic. The remaining Python modules are for installers that use an older `conda-standalone`.

constructor/nsis/main.nsi.tmpl
marcoesters: That's what we're doing for silent installations. This would mean that we don't see the output in real time though. If you think the file logging is more important, I can change that.

constructor/nsis/main.nsi.tmpl
marcoesters: That's correct

constructor/nsis/main.nsi.tmpl
marcoesters: I just tested this and the extraction command fails when `$INSTDIR` contains spaces. It succeeds when `$R3` is surrounded by quotes.

constructor/nsis/main.nsi.tmpl
marcoesters: I honestly don't remember. I just tested `FileWrite` and it looks okay.

constructor/nsis/_nsis.py
marcoesters: Done

constructor/nsis/main.nsi.tmpl
jaimergp: I don't think we gain too much by having real time output in the post install scripts (they should usually be quite quick to run). The other idea is to use `powershell`'s [`Tee-Object`](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/tee-object?view=powershell-7.5)

jaimergp: Ping here @marcoesters

marcoesters: This wasn't as trivial as it seemed at first because `type` changes the exit code such that commands always succeed. I found a workaround - not pretty, but it works.

jaimergp (approved):

PR #1123, opened 2025-12-01
Prepare 3.14

marcoesters:
✂️ ✂️ ✂️ Snip, snip, release! ✂️ ✂️ ✂️

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,9 @@
 [//]: # (current developments)
 
-## 2025-12-01   4.0.0:
+## 2025-12-02   3.14.0:
 ### Enhancements
 
-* **Breaking change**: Replace custom Python script with `conda-standalone` calls.
+* Replace custom Python script with `conda-standalone` calls.
   This removes Python as an implicit dependency from installers. (#549 via #1089)
 * EXE: Improve handling of options `initialize_conda`, `register_python` with their corresponding default values. The behavior of these options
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,27 @@
 [//]: # (current developments)
 
+## 2025-12-01   3.13.2:
+### Enhancements
+
+* **Breaking change**: Replace custom Python script with `conda-standalone` calls.
```
jaimergp: If this is a breaking change, should we at least bump to 3.14? If we should not, is it really a breaking change?

jezdez: Agreed, breaking changes are usually leading to a bump in the MAJOR version though, not in a MINOR, or even PATCH version.

marcoesters: The reason why I added this to the news file is that under some circumstances (like using `mamba` version 2), Python is not added to the installer anymore unless requested in the `constructor.yaml` file. I think this is a breaking change.

jaimergp: I don't think it's breaking in the sense of requiring v4, imo. It's a change in how the installer is bootstrapped, but not in a way that would impact users. If the installer implicitly required `python` but it was not added as an explicit dependency, it's really unlikely none of the other packages mentioned would have brought it in.

In that sense I'm inclined to call this release 3.14, not 4.0.

marcoesters: That sounds fair to me. Adding Python implicitly was never really a feature, more of a necessary nuisance. It was never really transparent to the user why Python was always added. I removed the breaking change language from the change log.

PR #1124, opened 2025-12-07
Assign default values to uninitialized variables

lrandersson:
### Description
This PR resolves a regression introduced in https://github.com/conda/constructor/commit/227bb6511026059de247aa35b63ecc93f0be4738.
Initially I was planning to fix this only for `ARGV_NoRegistry` but I noticed that there are two errors that are handled as `${If} ${Errors} -> assign`.

### Checklist - did you ...

### DIFF
--- a/news/1124-assign-uninitialized-variables
+++ b/news/1124-assign-uninitialized-variables
@@ -4,7 +4,7 @@
 
 ### Bug fixes
 
-* A regression with uninitialized variables has been fixed, this affected constructor 3.14.0. (#1124)
+* EXE: Fix a regression with uninitialized variables that prevented installations from being added to the "Add/Remove Programs" list. (#1124)
 
 ### Deprecations
 


constructor/nsis/main.nsi.tmpl
```diff
@@ -754,16 +759,6 @@ Function .onInit
         ${EndIf}
     ${EndIf}
 
-    ; Set default value
-    ${If} $CheckPathLength == ""
-        StrCpy $CheckPathLength "1"
-    ${EndIf}
-
-    StrCpy $CheckPathLength "{{ 1 if check_path_length else 0 }}"
```
lrandersson, line 759: I noticed we set `CheckPathLength` to 1 if it is not set, and then we set it to either 0 or 1. Seems odd? The documentation also says default is `False` for Windows regarding `check_path_length`, let me know what you think we should do here.

Note that I moved this code up to `Function InitializeVariables` and removed the "if empty set to 1" part.

Then in `cli-options.md` it says:
```
- `/CheckPathLength=[0|1]`: Check whether the installation path is too long (>46
  characters). The default depends on how the installer was created.
  ```
which I think looks like it is now with the changes in this PR.

File: news/1124-assign-uninitialized-variables [hunk 2]
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* A regression with uninitialized variables has been fixed, this affected constructor 3.14.0. (#1124)
```
marcoesters:
```suggestion
* EXE: Fix a regression with uninitialized variables that prevented installations from being added to the "Add/Remove Programs" list. (#1124)
```

PR #1132, opened 2025-12-08
Fix issue with GetOptions resetting variable and account for KeepPkgCache

lrandersson:
### Description
This PR introduces an issue detected with NSIS `GetOptions` resetting the value of `ARGV_NoRegistry` as well as properly accounted for the CLI arg `KeepPkgCache`, looks like this was ignored before.

### Checklist - did you ...

### DIFF
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -104,10 +104,6 @@ ${Using:StrFunc} StrStr
 var /global INIT_CONDA
 var /global REG_PY
 
-# Introduce an additional variable to handle the value of ARGV_NoRegistry,
-# since when CLI arguments are parsed the NSIS function GetOptions
-# will reset the value of variable if not set,
-# hence, we cannot rely strictly on ARGV_NoRegistry
 var /global NO_REGISTRY
 
 var /global INSTDIR_JUSTME


tests/test_examples.py
```diff
@@ -1423,3 +1466,45 @@ def test_regressions(tmp_path, request):
             check_subprocess=True,
             uninstall=True,
         )
+
+
+@pytest.mark.parametrize("no_registry", (0, 1))
+@pytest.mark.skipif(not ON_CI, reason="CI only")
+@pytest.mark.skipif(not sys.platform.startswith("win"), reason="Windows only")
+def test_not_in_installed_menu_list_(tmp_path, request, no_registry):
+    """Verify the app is Installed Apps Menu (or not), based the CLI arg '/NoRegistry'.
```
lrandersson:
```suggestion
    """Verify the app is in the Installed Apps Menu (or not), based on the CLI arg '/NoRegistry'.
```

news/1132-fix-cli-args
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* EXE: Fixed an issue for silent installers where command-line argument `/KeepPkgCache` was ignored and `/NoRegistry` would reset the default value.
```
marcoesters:
```suggestion
* EXE: Fixed an issue for silent installers where command-line argument `/KeepPkgCache` was ignored and `/NoRegistry` would reset the default value. (#1132)
```

tests/test_examples.py
```diff
@@ -1423,3 +1466,45 @@ def test_regressions(tmp_path, request):
...
+    name = ""
+    version = ""
+    for line in contents:
+        if "name: " in line:
+            name = line.split(":")[-1].strip()
+        if "version: " in line:
+            version = line.split(":")[-1].strip()
+        if name and version:
+            break
+    partial_name = f"{name} {version}"
```
marcoesters, line 1495: This seems brittle. Instead of manually parsing the `construct.yaml` file, I suggest using the `UninstallString` value and see if it starts with `$INSTDIR\Uninstall-`: https://github.com/conda/constructor/blob/7a837bd9d3c94813cc4734ebc6f9879557ddd488/constructor/nsis/main.nsi.tmpl#L1627-L1628

news/1132-fix-cli-args
lrandersson: https://github.com/conda/constructor/pull/1132/commits/19e254d569129fbdf08eca9434bb1d8e39b294b8

tests/test_examples.py
lrandersson: Hmm, what about using the variable `installer` we get from `create_installer`? If I'm not mistaken that should be the path to the `exe` and it is named as `<name>-<version>-<build number>-<platform>-<arch>.exe`. If we split `installer.name` on `-` we have name from `[0]` and version from `[1]`, let me know what you think.

marcoesters: I don't think the installer uses the build number (it's part of the version string), but that could work, too.

lrandersson: Reduced it to
```
    installer_file_name_parts = Path(installer).name.split("-")
    name = installer_file_name_parts[0]
    version = installer_file_name_parts[1]
    partial_name = f"{name} {version}"
```

constructor/nsis/main.nsi.tmpl
```diff
@@ -104,6 +104,12 @@ ${Using:StrFunc} StrStr
 var /global INIT_CONDA
 var /global REG_PY
 
+# Introduce an additional variable to handle the value of ARGV_NoRegistry,
+# since when CLI arguments are parsed the NSIS function GetOptions
+# will reset the value of variable if not set,
+# hence, we cannot rely strictly on ARGV_NoRegistry
+var /global NO_REGISTRY
```
```suggestion
var /global NO_REGISTRY
```

I don't think this explanation is necessary. We want to replace these variables with global values anyway. In the end, I think all `ARGV_` variables should be removed and replaced with unnamed registers that are just used for parsing.

lrandersson: Added the suggestion, and I agree all the ARGV_ should be updated in a similar fashion.

PR #1133, opened 2025-12-09
MSI: Integration tests

lrandersson:
### Description

Opened this one separately, this branch and PR depends on https://github.com/conda/constructor/pull/1084. No need to review this for now since the diff will look strange and is more for book-keeping since these changes were pulled out of PR 1084.

### Checklist - did you ...

### DIFF
--- a/constructor/briefcase.py
+++ b/constructor/briefcase.py
@@ -112,7 +112,7 @@ def get_license(info):
         return {"file": info["license_file"]}
 
     placeholder_license = Path(__file__).parent / "nsis" / "placeholder_license.txt"
-    return {"file": str(placeholder_license)} # convert to str for TOML serialization
+    return {"file": str(placeholder_license)}  # convert to str for TOML serialization
 
 
 # Create a Briefcase configuration file. Using a full TOML writer rather than a Jinja
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -412,7 +412,6 @@ def _run_uninstaller_msi(
     install_dir: Path,
     timeout: int = 420,
     check: bool = True,
-    request: bool = False,
 ) -> subprocess.CompletedProcess | None:
     cmd = [
         "msiexec.exe",
@@ -425,8 +424,8 @@ def _run_uninstaller_msi(
         # TODO:
         # Check log and if there are remaining files, similar to the exe installers
         pass
-    if request:
-        request.addfinalizer(lambda: shutil.rmtree(str(install_dir), ignore_errors=True))
+    # This is temporary until uninstallation works fine
+    shutil.rmtree(str(install_dir), ignore_errors=True)
 
     return process
 
@@ -494,9 +493,7 @@ def _run_installer(
         _sentinel_file_checks(example_path, install_dir)
     if uninstall:
         if installer.suffix == ".msi":
-            _run_uninstaller_msi(
-                installer, install_dir, timeout=timeout, check=check_subprocess, request=request
-            )
+            _run_uninstaller_msi(installer, install_dir, timeout=timeout, check=check_subprocess)
         elif installer.suffix == ".exe":
             _run_uninstaller_exe(install_dir, timeout=timeout, check=check_subprocess)
     return process
@@ -647,7 +644,7 @@ def test_example_extra_envs(tmp_path, request):
 
         if sys.platform.startswith("win"):
             if installer.suffix == ".msi":
-                _run_uninstaller_msi(installer, install_dir, request=request)
+                _run_uninstaller_msi(installer, install_dir)
             else:
                 _run_uninstaller_exe(install_dir=install_dir)
 
@@ -741,7 +738,7 @@ def test_example_miniforge(tmp_path, request, example):
                 assert not list(start_menu_dir.glob("Miniforge*.lnk"))
             elif installer.suffix == ".msi":
                 # TODO: Start menus
-                _run_uninstaller_msi(installer, install_dir, request=request)
+                _run_uninstaller_msi(installer, install_dir)
                 raise NotImplementedError("Test needs to be implemented")
 
 
@@ -907,7 +904,7 @@ def test_example_shortcuts(tmp_path, request):
             else:
                 raise AssertionError("No shortcuts found!")
             if installer.suffix == ".msi":
-                _run_uninstaller_msi(installer, install_dir, request=request)
+                _run_uninstaller_msi(installer, install_dir)
             else:
                 _run_uninstaller_exe(install_dir)
             assert not (package_1 / "A.lnk").is_file()


.github/workflows/main.yml
```diff
@@ -135,6 +135,18 @@ jobs:
...
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          repository: beeware/briefcase
+          ref: efb0e08c2a9ce72dcebd0a62c12f5662fd5eb0e0 # this is a commit with latest necessary changes on the main branch
+          path: briefcase
+          fetch-depth: 1
+
+      - name: Install Briefcase (editable)
+        run: |
+          pip install -e briefcase
```
marcoesters: Is this still needed? Either way, this should only be installed for Windows.

tests/test_examples.py
```diff
@@ -862,8 +986,11 @@ def test_example_from_explicit(tmp_path, request):
 
 
 def test_register_envs(tmp_path, request):
+    """Verify that 'register_envs: False' results in the environment not being registered."""
     input_path = _example_path("register_envs")
     for installer, install_dir in create_installer(input_path, tmp_path):
+        if installer.suffix == ".msi":
+            raise Exception("Test for 'register_envs' not yet implemented for MSI")
```
marcoesters: Should these exceptions be `NotImplementedError`?

.github/workflows/main.yml
lrandersson: You're right, I removed that

tests/test_examples.py
lrandersson: Agree, changed to `NotImplementedError`

constructor/briefcase.py
```diff
@@ -100,6 +105,15 @@ def get_bundle_app_name(info, name):
     return bundle, app_name
 
 
+def get_license(info):
+    """Retrieve the specified license as a dict or return a placeholder if not set."""
+
+    if "license_file" in info:
+        return {"file": info["license_file"]}
+    # We cannot return an empty string because that results in an exception on the briefcase side.
+    return {"text": "TODO"}
```
marcoesters: We do have a placeholder license: https://github.com/conda/constructor/blob/5eecfa0e8b9a46fa96bc19eb1cbf08a91a25fa52/constructor/nsis/placeholder_license.txt

tests/test_examples.py
```diff
@@ -336,6 +338,94 @@ def _sentinel_file_checks(example_path, install_dir):
...
+
+def calculate_msi_install_path(installer: Path) -> Path:
+    """This is a temporary solution for now since we cannot choose the install location ourselves.
+    Installers are named <name>-<version>-Windows-x86_64.msi.
+    """
+    dir_name = installer.name.replace("-Windows-x86_64.msi", "").replace("-", " ")
+    if is_admin():
+        root_dir = Path(os.environ.get("PROGRAMFILES") or r"C:\Program Files")
+    else:
+        local_dir = os.environ.get("LOCALAPPDATA") or str(Path.home() / r"AppData\Local")
```
marcoesters: Are you using `or` instead of default values to account for empty strings?

tests/test_examples.py
```diff
@@ -336,6 +338,94 @@ def _sentinel_file_checks(example_path, install_dir):
...
+        "/qn",
+    ]
+
+    log_path = Path(os.environ.get("TEMP")) / (install_dir.name + ".log")
+    cmd.extend(["/L*V", str(log_path)])
+    try:
+        process = _execute(cmd, installer_input=installer_input, timeout=timeout, check=check)
+    except subprocess.CalledProcessError as e:
+        if log_path.exists():
+            # When running on the CI system, it gets misdecodes a UTF-16 log file as UTF-8,
```
```suggestion
            # When running on the CI system, it tries to decode a UTF-16 log file as UTF-8,
```

tests/test_examples.py
```diff
@@ -381,12 +471,30 @@ def _run_installer(
...
+        print("sentinel_file_checks for MSI installers not yet implemented")
+    elif check_sentinels and not (installer.suffix == ".pkg" and ON_CI):
         _sentinel_file_checks(example_path, install_dir)
-    if uninstall and installer.suffix == ".exe":
-        _run_uninstaller_exe(install_dir, timeout=timeout, check=check_subprocess)
+    if uninstall:
+        if installer.suffix == ".msi":
+            if request:  #  and ON_CI
+                # We always need to do this currently since uninstall doesnt work fully
+                request.addfinalizer(lambda: shutil.rmtree(str(install_dir), ignore_errors=True))
```
marcoesters: Why not add `rmtree` to `_run_uninstaller_msi` then?

tests/test_examples.py
```diff
@@ -528,13 +642,27 @@ def test_example_extra_envs(tmp_path, request):
             assert "@EXPLICIT" in envtxt.read_text()
 
         if sys.platform.startswith("win"):
-            _run_uninstaller_exe(install_dir=install_dir)
+            if installer.suffix == ".msi":
+                if request:
+                    request.addfinalizer(
+                        lambda: shutil.rmtree(str(install_dir), ignore_errors=True)
+                    )
```
marcoesters: Same here - consider adding to `_run_uninstaller_msi`, especially since this is used in other tests later.

constructor/briefcase.py
lrandersson: Thanks, updated!

tests/test_examples.py
lrandersson: Correct, updated to follow the convention we use with `.get("FOO", <default>)`

tests/test_examples.py
lrandersson: Fixed

tests/test_examples.py
lrandersson: Initially I didn't want to the interface to `*_uninstaller_msi` and `*_uninstaller_exe` to differ, but since this is temporary I've updated it now, I agree it looks more concise this way.

tests/test_examples.py
lrandersson: Fixed

tests/test_examples.py
```diff
@@ -410,6 +412,7 @@ def _run_uninstaller_msi(
     install_dir: Path,
     timeout: int = 420,
     check: bool = True,
+    request: bool = False,
```
marcoesters: `request` is not a boolean, but a `FixtureRequest` object: https://docs.pytest.org/en/stable/reference/reference.html#request

However, I would say that it's generally not needed as a finalizer. We can just add `rmtree` as a clean-up step until the uninstallation works well.

lrandersson: Ahh good that you caught that, I should **not** make assumptions based on an expression like `if request:`, I've updated it https://github.com/conda/constructor/pull/1133/changes/496bf86e1af3de04da75a8d6d4082c5cbd96b204, let me know if this is what you were thinking. I think this should be OK now, we always invoke `rmtree`, hopefully no test depended on `rmtree` not being called.

PR #1090, opened 2025-10-16
Preserve name of conda exe

lrandersson:
### Description

Potential fix for https://github.com/conda/constructor/issues/1033

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/constructor/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [x] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #1135, opened 2025-12-11
fix: symbolic linking (`ln -s`) with pre-existing installation

dlaehnemann:
### Description

This is an issue we came across very far downstream, when using the [`setup-miniconda` GitHub Action](https://github.com/conda-incubator/setup-miniconda/tree/main) in multiple steps of the same GitHub Actions job implicitly (through the [`snakemake` GitHub Action that uses it](https://github.com/snakemake/snakemake-github-action/blob/73be4c1951827687d8c1c2eb4f1c29f278618422/action.yml#L76-L83), which we use for two subsequent steps).

Basically, the during first use of the `snakemake` GitHub Action, `setup-miniconda` installs `miniforge` without any issues. During the second install, it throws [this error](https://github.com/snakemake-workflows/rna-seq-star-deseq2/actions/runs/19935872631/job/57477521420#step:6:68):
```
Running installer...
  /usr/bin/bash /home/runner/work/_temp/e83768cc-2602-4a3d-a62b-6a1cc1af9dd4.sh -f -b -p /home/runner/miniconda3
  PREFIX=/home/runner/miniconda3
  Unpacking bootstrapper...
  Warning: ln: failed to create symbolic link '/home/runner/miniconda3/_conda': File exists
  
  ln: failed to create symbolic link '/home/runner/miniconda3/_conda': File exists
Error: The process '/usr/bin/bash' failed with exit code 1
```

Finding out where that linking actually happens, so where the installers actually take their script code from, was a bit of treasure hunt. But here we are. The code I am suggesting to modify was only recently added in #1090 and thus seems very likely to be the cause of this. But if this forcing of the symlinking does not look like the correct fix to you, I'd be glad to stand corrected.

### Checklist - did you ...

- [ ] ~Add / update outdated documentation?~

### DIFF
--- a/news/1135-fix-symbolic-linking
+++ b/news/1135-fix-symbolic-linking
@@ -1,3 +1,3 @@
 ### Bug fixes
 
-* [#1090](https://github.com/conda/constructor/pull/1090) introduced symbolic linking of `_conda`. This fails the whole installer, if the `_conda` file (or link) exists at the target location. [#1135](https://github.com/conda/constructor/pull/1135) fixes this, by forcing (`-f`) the symlink.
+* Force symbolic linking of `_conda` in SH and PKG installers. This fixes a regression introduced by #1090 that made installations fail if `_conda` already exists in the target location. (#1135)


dlaehnemann:
Pinging @marcoesters , as he reviewed the pull request #1090 that brought in this code.

news/1135-fix-symbolic-linking
```diff
@@ -0,0 +1,3 @@
+### Bug fixes
+
+* [#1090](https://github.com/conda/constructor/pull/1090) introduced symbolic linking of `_conda`. This fails the whole installer, if the `_conda` file (or link) exists at the target location. [#1135](https://github.com/conda/constructor/pull/1135) fixes this, by forcing (`-f`) the symlink.
```
marcoesters:
```suggestion
* Force symbolic linking of `_conda` in SH and PKG installers. This fixes a regression introduced by #1090 that made installations fail if `_conda` already exists in the target location. (#1135)
```

I'm reformatting this a little to be more consistent with the style we have in other release notes.

dlaehnemann:
Perfect, that was quick. Many thanks!

PR #1143, opened 2025-12-17
Briefcase uninstall

lrandersson:
### Description
Needs to be rebased once https://github.com/conda/constructor/pull/1133 is merged.

### Checklist - did you ...


constructor/briefcase.py
```diff
@@ -64,53 +60,235 @@ def get_name_version(info):
...
+def get_license(info):
+    """Retrieve the specified license as a dict or return a placeholder if not set."""
+
+    if "license_file" in info:
+        return {"file": info["license_file"]}
+    # We cannot return an empty string because that results in an exception on the briefcase side.
+    return {"text": "TODO"}
+
+
+class UninstallBat:
```
lrandersson, line 102: This approach is a bit different from how we currently handle `run_installation.bat`. I'd like to hear some feedback/thoughts on what you (the reviewer) thinks.

I went with this class-based approach because I think it's better in terms of future maintenance and also makes unit testing very convenient, and easy to read. All the code for it is in one place and easy to follow. I think most of `constructor` is written in a "scripted" approach and I think that works for small projects, but for larger projects it can quickly become cumbersome. I would argue the project could benefit from a more object-oriented approach in general.
Moreover, simplifying the testing would be a good win for the project. One of the major pain points in `constructor` is the difficulty of running the tests locally, and to do it fast and "as close to production" as possible.

If we don't wanna go with this approach, it wouldn't take a lot of effort to simply add a `pre_uninstall.bat` and change the implementation, but I would propose we add something similar also for `run_installation.bat` (I'm happy to do that update if we go with it).

This can also be expanded upon at a later point, for instance by changing the lists of commands to jinja-based templating.

Issue #1103, opened 2025-11-06
MSI: Add a UI dialog with additional installation options


lrandersson:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

Based on the work in https://github.com/conda/constructor/pull/1084, when you build an MSI installer you cannot prompt the user to choose specific installation options. For comparison the `exe` installers have this behavior already via `OptionsDialog.nsh`, where they can choose several optional features for their installation.

### Why is this needed?

To enable the user to customize their installation with additional features/behaviors. For example, when a user runs my installer I want to allow the user to decide if a specific environment variable should be added to `PATH` or not. This is also already available via the `exe` installers.

### What should happen?

A UI dialog similar to the current behavior with `exe` installers should be displayed where the user can customize their installation.

mhsmith:
Briefcase's installer options are documented [here](https://briefcase.beeware.org/en/latest/reference/platforms/windows/#installer-options). Adding at least one option to the pyproject.toml file should cause the options dialog to appear.

lrandersson:
https://github.com/conda/constructor/pull/1179

PR #1144, opened 2025-12-17
Briefcase installer options

lrandersson:
### Description
This does not fully resolve https://github.com/conda/constructor/issues/1103 but is a start.

### Checklist - did you ...

### DIFF
--- a/constructor/briefcase.py
+++ b/constructor/briefcase.py
@@ -132,18 +132,16 @@ def create_install_options_list(info: dict) -> list[dict]:
             has_python = True
             break
 
-    if has_python:
-        register_python = info.get("register_python", True)
-        if register_python:
-            options.append(
-                {
-                    "name": "register_python",
-                    "title": f"Register {info['name']} as my default Python {python_version}.",
-                    "description": "Allows other programs, such as VSCode, PyCharm, etc. to automatically "
-                    f"detect {info['name']} as the primary Python {python_version} on the system.",
-                    "default": info.get("register_python_default", False),
-                }
-            )
+    if has_python and info.get("register_python", True):
+        options.append(
+            {
+                "name": "register_python",
+                "title": f"Register {info['name']} as my default Python {python_version}.",
+                "description": "Allows other programs, such as VSCode, PyCharm, etc. to automatically "
+                f"detect {info['name']} as the primary Python {python_version} on the system.",
+                "default": info.get("register_python_default", False),
+            }
+        )
 
     # Initialize conda
     initialize_conda = info.get("initialize_conda", "classic")
@@ -201,13 +199,9 @@ def create_install_options_list(info: dict) -> list[dict]:
 
         if script:
             script_path = Path(script)
-            if not script_path.is_file():
-                raise FileNotFoundError(
-                    f"Specified {script_type}-install script '{script}' does not exist."
-                )
             if not is_bat_file(script_path):
-                raise OSError(
-                    f"Specified {script_type}-install script '{script}' must be a '.bat' file."
+                raise ValueError(
+                    f"Specified {script_type}-install script '{script}' must be an existing '.bat' file."
                 )
 
         # The UI option is only displayed if a description is set


constructor/briefcase.py
```diff
@@ -100,6 +100,34 @@ def get_bundle_app_name(info, name):
...
+            {
+                "name": "register_python",
+                "title": "Register Python as System Default",
+                "description": "TODO: Register Python description",
+                "default": info.get("register_python_default", False),
+            }
+        )
+    initialize_conda = info.get("initialize_conda", "classic")
+    if initialize_conda:
+        # TODO: How would we distinguish between True/classic in the UI? Same for NSIS
```
marcoesters: `True` and `classic` are synonyms. You would have to distinguish between `classic` and `condabin`. You can currently just take what's in the NSIS template and add that here.

constructor/briefcase.py
```diff
@@ -100,6 +100,34 @@ def get_bundle_app_name(info, name):
...
+def create_install_options_list(info: dict) -> list[dict]:
+    """Returns a list of dicts with data formatted for the installation options page."""
+    options = []
+    register_python = info.get("register_python", True)
+    if register_python:
+        options.append(
+            {
+                "name": "register_python",
+                "title": "Register Python as System Default",
+                "description": "TODO: Register Python description",
```
marcoesters: For now, you can just copy the description in NSIS.

constructor/briefcase.py
lrandersson: Fixed https://github.com/conda/constructor/pull/1144/changes/326be7f9764d9b24509ec0a80a0d330bb3567cd4

constructor/briefcase.py
lrandersson: Fixed https://github.com/conda/constructor/pull/1144/changes/326be7f9764d9b24509ec0a80a0d330bb3567cd4

constructor/briefcase.py
marcoesters: I was thinking about these lines: https://github.com/conda/constructor/blob/97dbba0fc910ee103e84ffc626c63c66884dd79b/constructor/nsis/OptionsDialog.nsh#L167-L178

You can then add the `condabin` UI text if `info.get("initialize_conda", "classic") == "condabin"`, and the other text if not.

lrandersson: Ah, updated https://github.com/conda/constructor/pull/1144/changes/a642278a95ec39a50b17fd63e496ca0e2961b297

constructor/briefcase.py
```diff
@@ -115,6 +115,42 @@ def get_license(info):
     return {"file": str(placeholder_license)}  # convert to str for TOML serialization
 
 
+def create_install_options_list(info: dict) -> list[dict]:
+    """Returns a list of dicts with data formatted for the installation options page."""
+    options = []
+    register_python = info.get("register_python", True)
+    if register_python:
+        python_version = f"{sys.version_info.major}.{sys.version_info.minor}"
```
marcoesters: This will be the Python version of the Python binary used to call `constructor`. This is not necessarily the same as the one bundled in the installer - in fact, `python` may not even be in the installer at all. You will have to use `_dists` in the `info` dictionary to get the Python version bundled in the installer.

constructor/briefcase.py
```diff
@@ -115,6 +115,42 @@ def get_license(info):
...
+                "name": "register_python",
+                "title": f"Register {info['name']} as my default Python {python_version}.",
+                "description": "Allows other programs, such as VSCode, PyCharm, etc. to automatically "
+                f"detect {info['name']} as the primary Python {python_version} on the system.",
+                "default": info.get("register_python_default", False),
+            }
+        )
+    initialize_conda = info.get("initialize_conda", "classic")
+    if initialize_conda:
+        if initialize_conda == "condabin":
```
marcoesters, line 149: You could remove a layer of nesting with this structure:

```
if initialize_conda == "condabin":
    ...
elif initialize_conda:
    ...
```

constructor/briefcase.py
lrandersson: While looking into `info["dists"]`, I found this seemingly old code in `winexe.py`:
```
    make_nsi(
        {
            "name": "Maxi",
            "version": "1.2",
            "_platform": "win-64",
            "_outpath": "dummy.exe",
            "_download_dir": "dummy",
            "_dists": ["python-2.7.9-0.tar.bz2", "vs2008_runtime-1.0-1.tar.bz2"],
        },
        ".",
    )
```
Is it safe to assume that if an installer bundles python, the package is formatted as above? I.e. `python-x.y.z-<build number>.tar.bz2`?

marcoesters: In general, package names are formatted as `<package name>-<version>-<build string>.<extension>`. `<build string>` is not necessarily a number and `<package name>` may contain dashes. `<extension>` is either `.tar.bz2` or `.conda`, but that may change in the future.

lrandersson: See https://github.com/conda/constructor/pull/1144/changes/6087f5ebae6b3d233a28e54b2ec9684ce04c27ea

constructor/briefcase.py
lrandersson: I thought about that, but because `options.add(...)` should only be called `if initialize_conda` evalutes to True, I decided to keep it. Alternatively with your suggestion I would need:
```
if initialize_conda == "condabin":
    ...
elif initialize_conda:
    ...
if initialize_conda:
    options.add(...)
```
Let me know what you think.

constructor/briefcase.py
```diff
@@ -115,6 +115,115 @@ def get_license(info):
...
+        if register_python:
+            options.append(
+                {
+                    "name": "register_python",
+                    "title": f"Register {info['name']} as my default Python {python_version}.",
+                    "description": "Allows other programs, such as VSCode, PyCharm, etc. to automatically "
+                    f"detect {info['name']} as the primary Python {python_version} on the system.",
+                    "default": info.get("register_python_default", False),
+                }
+            )
```
```suggestion
    if has_python and info.get("register_python", True)
        options.append(
            {
                "name": "register_python",
                "title": f"Register {info['name']} as my default Python {python_version}.",
                "description": "Allows other programs, such as VSCode, PyCharm, etc. to automatically "
                f"detect {info['name']} as the primary Python {python_version} on the system.",
                "default": info.get("register_python_default", False),
            }
        )
```

constructor/briefcase.py
```diff
@@ -115,6 +115,115 @@ def get_license(info):
...
+                "default": False,
+            }
+        )
+
+    # Pre/Post install script
+    for script_type in ["pre", "post"]:
+        script_description = info.get(f"{script_type}_install_desc", "")
+        script = info.get(f"{script_type}_install", "")
+        if script_description and not script:
+            raise ValueError(
```
marcoesters, line 196:
```suggestion
            raise KeyError(
```

Technically since it is a key in `construct.yaml`?

constructor/briefcase.py
```diff
@@ -115,6 +115,115 @@ def get_license(info):
...
+                f"{script_type}_install_desc was set, but {script_type}_install was not!"
+            )
+
+        if script:
+            script_path = Path(script)
+            if not script_path.is_file():
+                raise FileNotFoundError(
+                    f"Specified {script_type}-install script '{script}' does not exist."
+                )
+            if not is_bat_file(script_path):
```
marcoesters, line 202: I think you can just check for the suffix here. The only thing `is_bat_file` does additionally is checking whether the file exists - which you do immediately before this.

marcoesters (commented):
Just a few small comments

constructor/briefcase.py
lrandersson: Fixed in 1095957a8783720196a9daa4b4dc128c940c507a

constructor/briefcase.py
lrandersson: I see your point but it seemed a bit odd with a `KeyError` here, to me a `KeyError` always looks like an uncaught exception. I was also thinking of creating a custom exception but since `constructor` only use default exceptions I decided to stick with it.
Do we raise `KeyError` in general for `yaml`-files that are not "proper"?

constructor/briefcase.py
lrandersson: I removed the first file check and updated the error message. Let me know if that works, 1095957a8783720196a9daa4b4dc128c940c507a.
Initially I coped the function `is_bat_file` from from the Uninstallation MSI branch, which is why I forgot to adjust to the double `is_file` check.

PR #1145, opened 2025-12-18
Remove use of Python 3.9, include 3.13 in testing

lrandersson:
### Description
This PR bumps the Python dependencies used in the testing suite.
* Remove Python 3.9
* Include Python 3.9
* Removed a `.txt` file that was written for macOS but not used, and the test `test_example_from_explicit` seems to be Linux only.

### Checklist - did you ...

### DIFF
--- a/news/1145-python-test-deps
+++ b/news/1145-python-test-deps
@@ -16,4 +16,4 @@
 
 ### Other
 
-* Removed Python `3.9` from the testing suite, include Python `3.13`. (#1145)
+* Remove Python `3.9` from the testing suite, include Python `3.13`. (#1145)


marcoesters:
I think we can use the opportunity here to update the minimum Python everywhere (`pyproject.toml`, `meta.yaml`).

lrandersson:
> I think we can use the opportunity here to update the minimum Python everywhere (`pyproject.toml`, `meta.yaml`).

Can we also remove `scripts\run_examples.py`? It looks very old and unused, it also references a few examples that I dont see referenced anywhere else.

marcoesters:
> > I think we can use the opportunity here to update the minimum Python everywhere (`pyproject.toml`, `meta.yaml`).
> 
> Can we also remove `scripts\run_examples.py`? It looks very old and unused, it also references a few examples that I dont see referenced anywhere else.

I'm all for getting rid of things that are not used.

news/1145-python-test-deps
```diff
@@ -0,0 +1,19 @@
...
+
+* <news item>
+
+### Docs
+
+* <news item>
+
+### Other
+
+* Removed Python `3.9` from the testing suite, include Python `3.13`. (#1145)
```
marcoesters:
```suggestion
* Remove Python `3.9` from the testing suite, include Python `3.13`. (#1145)
```

PR #1063, opened 2025-09-14
Use pathlib whenever possible

jaimergp:
### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](https://github.com/conda/constructor/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #1149, opened 2026-01-13
Provide metadata to easily protect environments

Jrice1317:
#### Overview
This PR adds frozen environment metadata support across all installer types, enabling users to easily protect their environments.
#### Changes
- Implemented metadata schema and validation
- Added support for shell (SH) installers
- Added support for macOS package (PKG) installers
- Added support for Windows executable (EXE) installers
- Extended integration test coverage across all installer types
- Updated documentation and examples
#### Related Links
https://conda.org/learn/ceps/cep-0022/

### Checklist - did you ...

### DIFF
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -1639,7 +1639,7 @@ Section "Install"
     File "{{ env.history_abspath }}"
 
 {%- if env.frozen_abspath %}
-    # Add frozen marker file if configured
+    # Add frozen marker file
     SetOutPath "{{ env.conda_meta }}"
     File "{{ env.frozen_abspath }}"
 {%- endif %}


constructor/data/construct.schema.json
```diff
@@ -124,6 +124,26 @@
...
+                "const": "conda"
+              },
+              "type": "object"
+            },
+            {
+              "type": "null"
+            }
+          ],
+          "default": null,
+          "description": "Same as `freeze_base`, but for this conda environment.",
```
lrandersson, line 144: "but for this conda environment", does this mean the same thing as "for the currently activated conda environment"?

constructor/main.py
```diff
@@ -105,6 +105,69 @@ def _win_install_needs_python_exe(conda_exe: str, conda_exe_type: StandaloneExe
     return results.returncode == 2
 
 
+# Validate frozen environments
+def validate_frozen_envs(info, exe_type, exe_version) -> bool:
```
lrandersson: Should we annotate the types for `info, exe_type, exe_version`?

constructor/main.py
```diff
@@ -105,6 +105,69 @@ def _win_install_needs_python_exe(conda_exe: str, conda_exe_type: StandaloneExe
...
+    """Validate frozen environments.
+
+    Checks:
+    - No conflicts between freeze_base/freeze_env and extra_files for same environment
+    - Conda-standalone version if frozen environments exist
+    """
+
+    def get_frozen_env(path) -> str | None:
```
lrandersson:
```suggestion
    def get_frozen_env(path: str) -> str | None:
```
(I'm assuming it's of type `str` so feel free to ignore)

constructor/main.py
```diff
@@ -105,6 +105,69 @@ def _win_install_needs_python_exe(conda_exe: str, conda_exe_type: StandaloneExe
...
+        parts = Path(path).parts
+        if parts == ("conda-meta", "frozen"):
+            return "base"
+        if len(parts) == 4 and parts[0] == "envs" and parts[-2:] == ("conda-meta", "frozen"):
+            return parts[1]
+        return None
+
+    # Collect environments using freeze_base/freeze_env
+    frozen_envs = set()
+    if info.get("freeze_base", {}).get("conda") is not None:
```
lrandersson, line 134: Is there any risk here that `info.get("freeze_base", {}).get("conda")` evaluates to an empty string and results in follow-up errors?

constructor/main.py
```diff
@@ -105,6 +105,69 @@ def _win_install_needs_python_exe(conda_exe: str, conda_exe_type: StandaloneExe
...
+        )
+
+    # Conda-standalone version validation
+    if exe_type == StandaloneExe.CONDA and check_version(
+        exe_version, min_version="25.5.0", max_version="25.7.0"
+    ):
+        sys.exit(
+            "Error: conda-standalone 25.5.x has known issues with frozen environments. "
+            "Please use conda-standalone 25.7.0 or newer."
+        )
```
lrandersson: Why is this a `sys.exit` call instead of an exception?

constructor/preconda.py
```diff
@@ -244,6 +250,14 @@ def write_conda_meta(info, dst_dir, final_urls_md5s, user_requested_specs=None):
         fh.write("\n".join(builder))
 
 
+def write_frozen(freeze_info, dst_dir):
```
lrandersson: Any type annotations needed?

constructor/shar.py
```diff
@@ -173,12 +173,20 @@ def create(info, verbose=False):
     pre_t.addfile(tarinfo=tarfile.TarInfo("conda-meta/history"))
     post_t.add(join(tmp_dir, "conda-meta", "history"), "conda-meta/history")
 
+    if os.path.exists(join(tmp_dir, "conda-meta", "frozen")):
+        post_t.add(join(tmp_dir, "conda-meta", "frozen"), "conda-meta/frozen")
```
lrandersson, line 177: Considering this PR is in draft [1063 ](https://github.com/conda/constructor/pull/1063) I'd suggest we also use `pathlib` here if possible (same for below).

news/1149-add-support-for-frozen-envs
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Add support for installing [protected conda environments](https://conda.org/learn/ceps/cep-0022#specification). (#1149)
```
lrandersson:
```suggestion
* Add support for installing [protected (frozen) conda environments](https://conda.org/learn/ceps/cep-0022#specification). (#1149)
```
Perhaps it would be good to expose the word `frozen` that we use for the feature. Feel free to ignore this.

constructor/data/construct.schema.json
Jrice1317: Not necessarily "currently activated". There's an option to have pre-configured environments be frozen.

lrandersson: Ahh, thanks for the clarification!

constructor/main.py
Jrice1317: It MUST be at the bare minimum:
```
freeze_base:
  conda: {}
```
OR
```
freeze_env:
  conda: {}
```
where the dictionary can be empty or populated with a key and value.

constructor/main.py
Jrice1317: Great question! My memory is a little fuzzy, but IIRC, it's because of _when_ this gets checked and how the workflow operates. Notice other checks below are not exceptions either, but I wondered the same thing.

Jrice1317: Pretty sure it's related to why my test originally failed when trying to expect my `RuntimeError`

constructor/main.py
Jrice1317: We could! However, I don't think it's necessary for `info` and I address the others with logic below. If we do it here, we would need to do it in the entire script to remain consistent and maybe that should be done in a separate PR. What do you think?

lrandersson: It's not a necessity, I added the comment mainly because I wasn't sure if we had a "requirement" or not in the code base. 

It's more that if someone else who has not implemented this function, is trying to use it in their code, it might not be as trivial for them to figure out the requirements of the input parameters - and here the type annotations would greatly help.

constructor/main.py
lrandersson: But can a user end-up in a situation where you have `info['freeze_base'] = {'conda': ''}`? In this situation `info.get("freeze_base", {}).get("conda") is not None` evaluates to `True` and I just want to make sure it doesn't result in strange behavior

constructor/main.py
lrandersson: I see, I also just noticed there are many `sys.exit` in the codebase, it seems a bit odd to me but perhaps there is a reason behind it. Do you know @marcoesters?

Anyway, feel free to resolve if it's working as intended.

constructor/data/construct.schema.json
Jrice1317: No problem!

constructor/main.py
Jrice1317: I completely agree! It took me a while myself to determine these things to implement the code. I'm all for this suggestion. I just don't want my PR to get any bigger than it already is 😅  because I feel we should add annotations everywhere. I'm happy to open a separate PR for this.

lrandersson: Sounds good to me! 😄

marcoesters: The campsite principle may help here: https://github.com/conda/constructor/blob/153753a334db176e5ceeeca7236b9a79d4becf26/HOW_WE_USE_GITHUB.md#should-we-make-unrelated-changes-at-the-same-time

While it would be nice to have it everywhere, we can add annotations where as we find missing annotations. The more annotations we add on-the-fly, the less daunting the extra PR would become.

constructor/main.py
marcoesters: It's definitely legacy/tech debt. This is an old code base. It's one of these things where we've talked about replacing `sys.exit`, but never did because there are so many. Exceptions should probably be the standard moving forward.

> Pretty sure it's related to why my test originally failed when trying to expect my RuntimeError

The reason for why checking for `RuntimeError` doesn't work is that `constructor` is executed as a subprocess: https://github.com/conda/constructor/blob/153753a334db176e5ceeeca7236b9a79d4becf26/tests/test_examples.py#L409-L418

So, the error will always be `subprocess.CalledProcessError`, no matter what the underlying exception is. `sys.exit()` itself raises the `SystemExit` exception.

constructor/main.py
lrandersson: I have not heard of the campsite principle before in software engineering, but I like it, all in favor!

constructor/main.py
Jrice1317: My apologies! I just realized you asked if it could evaluate to an empty string in your original question, but I just read "empty" 😅 No, [this](https://github.com/conda/constructor/pull/1149/changes#diff-59d91bb2afaa59e20826ccdba174a446cbb8d4d81666606c7288a83d8811aeb4R835) won't allow it. 
```
freeze_base: dict[Literal["conda"], dict] | None = None
```

constructor/shar.py
Jrice1317: I think using `os` is better in this case since we're relying on the strings here.

lrandersson: No problem, in the future though I think we should go with `pathlib` for consistency, it's a more modern approach over `os` for handling file-paths.

lrandersson (approved):

constructor/nsis/main.nsi.tmpl
```diff
@@ -1637,6 +1637,12 @@ Section "Install"
     # channels and retain only the first transaction
     SetOutPath "{{ env.conda_meta }}"
     File "{{ env.history_abspath }}"
+
+{%- if env.frozen_abspath %}
+    # Add frozen marker file if configured
```
```suggestion
    # Add frozen marker file
```

Nit: that comment won't be there if it's not configured.

constructor/main.py
marcoesters: I think you can raise a `RuntimeError`. Let's fix things where we find them.

constructor/preconda.py
```diff
@@ -56,7 +56,7 @@
 )
 
 
-def write_index_cache(info, dst_dir, used_packages):
+def write_index_cache(info: dict, dst_dir: str, used_packages):
```
marcoesters:
```suggestion
def write_index_cache(info: dict, dst_dir: str, used_packages: list[str]):
```

I think?

constructor/preconda.py
```diff
@@ -218,9 +222,11 @@ def write_files(info: dict, workspace: str):
         write_channels_txt(info, env_pkgs, env_config)
         # shortcuts
         write_shortcuts_txt(info, env_pkgs, env_config)
+        # frozen marker file
+        write_frozen(env_config.get("freeze_env"), env_conda_meta)
 
 
-def write_conda_meta(info, dst_dir, final_urls_md5s, user_requested_specs=None):
+def write_conda_meta(info: dict, dst_dir: str, final_urls_md5s: tuple, user_requested_specs=None):
```
marcoesters:
```suggestion
def write_conda_meta(info: dict, dst_dir: str, final_urls_md5s: tuple[str, str], user_requested_specs: list[str] | None = None):
```

constructor/preconda.py
```diff
@@ -271,7 +285,7 @@ def write_repodata_record(info, dst_dir):
             json.dump(rr_json, rf, indent=2, sort_keys=True)
 
 
-def write_initial_state_explicit_txt(info, dst_dir, urls):
+def write_initial_state_explicit_txt(info: dict, dst_dir: str, urls: tuple):
```
marcoesters: Just like with the md5 example above, you may want to add the types of the elements in the tuple

constructor/preconda.py
Jrice1317: I wasn't sure myself, which is why I left it hoping someone else would. 😅

PR #1151, opened 2026-01-21
Add environment variable with installer file path

lrandersson:
### Description

This PR adds an environment variable `INSTALLER_PATH` for installers that have `post_install: ...` defined.

### Checklist - did you ...

### DIFF
--- a/news/1151-installer-path-variable
+++ b/news/1151-installer-path-variable
@@ -1,7 +1,7 @@
 ### Enhancements
 
-An environment variable `INSTALLER_PATH` is now defined for `post_install` scripts and set to the path of the installer executable while the installer is running.
-* EXE: An environment variable `INSTALLER_PLUGINSDIR` is now also defined, it serves the same purpose as the NSIS variable `PLUGINSDIR`.
+* An environment variable `INSTALLER_PATH` is now defined for pre-install and post-install scripts, and set to the path of the installer executable while the installer is running. (#1151)
+* EXE: An environment variable `INSTALLER_PLUGINSDIR` is now also defined, it serves the same purpose as the NSIS variable `$PLUGINSDIR`. (#1151)
 
 ### Bug fixes
 


news/1151-installer-path-variable
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* EXE: An environment variable `INSTALLER_PATH` is now defined for `post_install` scripts and set to the path of installer executable while the installer is running.
```
lrandersson:
```suggestion
* EXE: An environment variable `INSTALLER_PATH` is now defined for `post_install` scripts and set to the path of the installer executable while the installer is running.
```

constructor/nsis/main.nsi.tmpl
```diff
@@ -649,6 +649,10 @@ Function .onInit
     Push $R1
     Push $R2
 
+{%- if post_install_exists %}
+    System::Call 'Kernel32::SetEnvironmentVariable(t, t) i("INSTALLER_PATH", "$EXEPATH")'
```
marcoesters: I would add this to where the other variables are defined in the `Install` section, and not make this a condition for post-install. It could also be used in pre-install scripts, maybe.

`PLUGINSDIR` may also be a good variable to add here.

lrandersson: I've moved it and added `INSTALLER_PLUGINSDIR`, see https://github.com/conda/constructor/pull/1151/changes/f8da86303043d538244e036b5884138ce5f4de1e.
I also considered if we should have it as:
```
...
SetEnvironmentVariable INSTALLER_PLUGINSDIR...
{%- if pre_install_exists or post_install_exists %}
SetEnvironmentVariable INSTALLER_PATH...
{%- endif %}
```
Thoughts? It doesn't hurt having `INSTALLER_PATH` always defined, but I also wonder if it is unnecessary,  is it always _safe_?

news/1151-installer-path-variable
lrandersson: https://github.com/conda/constructor/pull/1151/commits/049b6d653151042b78d019568bac666d89f86994 and https://github.com/conda/constructor/pull/1151/commits/64aaf246fcb7b9df03dde94d5b715cd9aee5be79

constructor/nsis/main.nsi.tmpl
marcoesters: Why would it be unsafe?

lrandersson: I can't think of any reason, but I am bringing it up just for consideration.

marcoesters: I don't see an attack vector right now. The environment variable is not available outside the installation process. So, they can only be used through in scripts. Apart from the pre-install or post-install scripts of the installer, those are the pre-link or pre-unlink scripts of the packages.

However, even if someone were to place a malicious package into an installer, this doesn't give the attacker any more privileges that they'd already have. So, the limitation won't reduce the overall exposure.

lrandersson: I agree, good assessment!

constructor/header.sh
```diff
@@ -81,6 +81,21 @@ export INSTALLER_NAME='{{ installer_name }}'
...
+esac
+
+[ -z "$SCRIPT_PATH" ] && {
+  echo "ERROR: Cannot determine installer path"
+  exit 1
+}
+# Resolve the directory to an absolute path
+SCRIPT_DIR=$(cd -- "$(dirname -- "$SCRIPT_PATH")" && pwd)
+INSTALLER_PATH="$SCRIPT_DIR/$(basename -- "$SCRIPT_PATH")"
+export INSTALLER_PATH
```
lrandersson: For visibility, note that this is based on a proposed AI solution and is said to be robust for both linux/mac with a `#!/bin/sh` shebang. I simplified it a bit because I removed an assumption to follow symlinks, I don't think that is necessary in this case?

It also is a bit _extra_ because it is trying to determine if the script was invoked using `.`, `source` or if the script was on `$PATH`.

constructor/osx/run_user_script.sh
```diff
@@ -30,6 +30,11 @@ export INSTALLER_NAME="{{ installer_name }}"
 export INSTALLER_VER="{{ installer_version }}"
 export INSTALLER_PLAT="{{ installer_platform }}"
 export INSTALLER_TYPE="PKG"
+
+# PACKAGE_PATH is poorly documented, and might not exist across all OS versions
+if [ -n "${PACKAGE_PATH:-}" ]; then
+    export INSTALLER_PATH="$PACKAGE_PATH" # The path to the .pkg installer
```
lrandersson, line 36: Do we want to add a default value? What we are trying to achieve here doesn't seem to be well-supported by Apple for `.pkg` installers, however, I found this and it worked for me when I tested locally. However, it's a poorly documented variable and might not always work.

marcoesters: > However, it's a poorly documented variable

That is, unfortunately, not unusual. This code may fail though with `set -u`. Here is how I solved it for another variable: https://github.com/conda/constructor/blob/dccf7d6c81764b6ad9311e7bdc57d76647a4b61c/constructor/osx/run_user_script.sh#L33-L40

constructor/header.sh
marcoesters: The script location is already a local variable, so you can just reuse that: https://github.com/conda/constructor/blob/dccf7d6c81764b6ad9311e7bdc57d76647a4b61c/constructor/header.sh#L89

constructor/osx/run_user_script.sh
marcoesters: I don't think a default variable makes sense here. If the location of the installer cannot be detected because the variable is not defined, then it shouldn't be set.

constructor/header.sh
lrandersson: Thanks, I noticed the entire path was already defined, I removed the earlier implementation and simply reuse the existing one: https://github.com/conda/constructor/pull/1151/changes/8cf48d31d2301da509a957c3c445670f408390e2

constructor/osx/run_user_script.sh
lrandersson: I don't think the `0` is necessary, I used the `:-` syntax as well (`if [ -n "${PACKAGE_PATH:-}" ]; then`) but this situation evaluates to an empty string if unset.

news/1151-installer-path-variable
```diff
@@ -0,0 +1,20 @@
+### Enhancements
+
+An environment variable `INSTALLER_PATH` is now defined for `post_install` scripts and set to the path of the installer executable while the installer is running.
+* EXE: An environment variable `INSTALLER_PLUGINSDIR` is now also defined, it serves the same purpose as the NSIS variable `PLUGINSDIR`.
```
marcoesters:
```suggestion
* An environment variable `INSTALLER_PATH` is now defined for pre-install and post-install scripts, and set to the path of the installer executable while the installer is running. (#1151)
* EXE: An environment variable `INSTALLER_PLUGINSDIR` is now also defined, it serves the same purpose as the NSIS variable `$PLUGINSDIR`. (#1151)
```

Please always add the PR number (and issue number, not applicable here) to the new items.

lrandersson: You're right, I keep forgetting that. Thanks for the reminder!

PR #1152, opened 2026-01-23
Account for state of option before running pre install script

lrandersson:
### Description

I encountered this bug while working on https://github.com/conda/constructor/pull/1151. Even though I unchecked the "Run pre install script" option in the installation options page, I still got an error "Failed to run pre install script" which seems odd. It looked like we were missing a check of the button. Note that we correctly had one for `post install` scripts:
```
    ${If} ${FileExists} "$INSTDIR\pkgs\post_install.bat"
        ${If} $Ana_PostInstall_State = ${BST_CHECKED}
            ${Print} "Running post-install..."
...
```

### Checklist - did you ...

### DIFF
--- a/news/1152-pre-install-script-bug
+++ b/news/1152-pre-install-script-bug
@@ -4,7 +4,7 @@
 
 ### Bug fixes
 
-* EXE: Resolved an issue where a `pre_install` script was invoked even if the option was unchecked from the installation options page.
+* EXE: Resolved an issue where a `pre_install` script was invoked even if the option was unchecked from the installation options page. (#1152)
 
 ### Deprecations
 


news/1152-pre-install-script-bug
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* EXE: Resolved an issue where a `pre_install` script was invoked even if the option was unchecked from the installation options page.
```
marcoesters:
```suggestion
* EXE: Resolved an issue where a `pre_install` script was invoked even if the option was unchecked from the installation options page. (#1152)
```

PR #1153, opened 2026-01-23
Add architecture checks for macOS installers

marcoesters:
### Description

Add architecture checks for macOS installers:

* For SH installers, add a check for `arm64` installers.
* For PKG installers, restrict the architecture to the architecture of the bundled packages.

The latter will lead to a message shown in the screenshot below right when the user executes the installer. I don't think that message can be changed, but it is immediate.

Alternatively, we could add a pre-install check as we do for existing paths to get a more descriptive error message. However, this error message only shows up late in the process.

image

### Checklist - did you ...

### DIFF
--- a/constructor/osxpkg.py
+++ b/constructor/osxpkg.py
@@ -213,7 +213,7 @@ def modify_xml(xml_path, info):
         options.set("customize", "allow")
         options.set("customLocation", "/")
 
-    arch = "arm64" if info["_platform"] == "osx-arm64" else "osx-64"
+    arch = "arm64" if info["_platform"] == "osx-arm64" else "x86_64"
     arch_key = ET.Element("options", hostArchitectures=arch)
     root.append(arch_key)
 


constructor/header.sh
```diff
@@ -244,6 +244,21 @@ then
...
+        printf "    Are sure you want to continue the installation? [yes|no]\\n"
+        printf "[no] >>> "
+        read -r ans
+        ans=$(echo "${ans}" | tr '[:lower:]' '[:upper:]')
+        if [ "$ans" != "YES" ] && [ "$ans" != "Y" ]
+        then
+            printf "Aborting installation\\n"
+            exit 2
+        fi
+    fi
```
lrandersson, line 261:
```suggestion
        if [ "$ans" != "YES" ] && [ "$ans" != "Y" ]; then
            printf "Aborting installation\\n"
            exit 2
        fi
    fi
```
Just a nitpick for consistency with line 248

marcoesters: It's consistent with how the other architecture checks are set up. I will leave it as is because committing this would just re-run the CI for a small cosmetic change.

PR #1159, opened 2026-01-29
Sign Mach-O binaries in `_internal` directory of `conda-standalone`

### LINKED ISSUE #1045: PKG installers with `onedir` variant of `conda-standalone` fail top be notarized
### Checklist

### What happened?

When building a PKG installer with the `onedir` build variant of `conda-standalone`, the installers fails to notarize. Installers containing single-binary `conda-standalone` builds, on the other hand, get notarized successfully.

I suspect that files inside the `_internal` directory need to be signed, too, to successfully notarize the installer.

### Additional Context

_No response_

marcoesters:
### Description

Sign all Mach-O binary files in the `_internal` directory of `conda-standalone` to pass notarization.

PKG installers built with the `onedir` variant of `conda-standalone` currently fail to be notarized because `constructor` only signs the `_conda` file in `conda-standalone`. The `onedir` build variant, however, contains an `_internal` directory with binaries that need to be signed with a trusted application certificate before notarization succeeds.

Closes #1045.

### Checklist - did you ...

### DIFF
--- a/constructor/osxpkg.py
+++ b/constructor/osxpkg.py
@@ -120,8 +120,6 @@ def sign_standalone_binary(
     }
     codesigner.sign_bundle(exe_path, entitlements=entitlements)
     internal_dir = exe_path.parent / "_internal"
-    if not internal_dir.is_dir():
-        return
     for file in internal_dir.glob("**"):
         if is_macho_binary(file):
             codesigner.sign_bundle(file, entitlements=entitlements)
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -724,8 +724,7 @@ def test_macos_signing(tmp_path, self_signed_application_certificate_macos):
     internal_dir = Path(
         expanded_path, "prepare_installation.pkg", "Payload", "osx-pkg-test", "_internal"
     )
-    if internal_dir.is_dir():
-        components.extend([file for file in internal_dir.glob("**") if is_macho_binary(file)])
+    components.extend([file for file in internal_dir.glob("**") if is_macho_binary(file)])
     validated_signatures = []
     for component in components:
         p = subprocess.run(


constructor/osxpkg.py
```diff
@@ -88,6 +107,26 @@ def _detect_mimetype(path: str):
...
+        "com.apple.security.cs.allow-unsigned-executable-memory": True,
+        "com.apple.security.cs.disable-executable-page-protection": True,
+        "com.apple.security.cs.disable-library-validation": True,
+        "com.apple.security.cs.allow-dyld-environment-variables": True,
+    }
+    codesigner.sign_bundle(exe_path, entitlements=entitlements)
+    internal_dir = exe_path.parent / "_internal"
+    if not internal_dir.is_dir():
+        return
+    for file in internal_dir.glob("**"):
```
lrandersson, line 123: Just a nitpick; Is there any risk that `_internal` is a file here? If we guard against `internal_dir` not existing, it is not necessary since `for file in internal_dir.glob('**')` will iterate over nothing (and not raise an error).

tests/test_examples.py
```diff
@@ -722,6 +721,11 @@ def test_macos_signing(tmp_path, self_signed_application_certificate_macos):
         Path(expanded_path, "prepare_installation.pkg", "Payload", "osx-pkg-test", conda_exe_name),
         Path(expanded_path, "Plugins", "ExtraPage.bundle"),
     ]
+    internal_dir = Path(
+        expanded_path, "prepare_installation.pkg", "Payload", "osx-pkg-test", "_internal"
+    )
+    if internal_dir.is_dir():
+        components.extend([file for file in internal_dir.glob("**") if is_macho_binary(file)])
```
lrandersson: Same here, we can remove `if internal_dir.is_dir():` if this is always a directory when it does exist.

constructor/osxpkg.py
```diff
@@ -445,16 +484,16 @@ def pkgbuild_prepare_installation(info):
         shutil.rmtree(f"{pkg}.expanded")
 
 
-def create_plugins(pages: list = None, codesigner: CodeSign = None):
-    def _build_xcode_projects(xcodeporj_dirs: list[Path]):
+def create_plugins(pages: list[str] | str | None = None, codesigner: CodeSign | None = None):
+    def _build_xcode_projects(xcodeproj_dirs: list[Path]):
```
lrandersson, line 486: Can we remove the input arugment `xcodeproj_dirs`? Looks like it is not used, glad though we fixed the spelling from `xcodeporj_dirs`. If we remove the input argument we should be able to remove some of the code also that is further below then where we construct the input argument, i.e.:
```
        if xcodeproj_dirs:
            _build_xcode_projects(xcodeproj_dirs)
```

marcoesters: It is used in the lines below. The reason this hasn't caused an error is because `xcodeproj_dirs` is not undefined outside the function when it's called.

constructor/osxpkg.py
marcoesters: It's even more convenient than that: if it is a file, it will also iterate over nothing. So, yes, the check can be skipped.

constructor/osxpkg.py
lrandersson: Oh, I see what you mean now. Feel free to resolve this conversation.

constructor/osxpkg.py
lrandersson: Ah, I didn't know that, thanks for sharing!

lrandersson (approved):

PR #1160, opened 2026-01-29
MSI: Change installer layout

lrandersson:
### Description

This PR is one of several that is needed to fully change the installer layout. It is expected that there are test failures with this PR since the uninstallation script has not been updated to account for the changes in this PR.

Currently, this PR is targeting `briefcase-integration`, but I propose that I create a temporary new branch `briefcase-integration-2` from `briefcase-integration`, merge this PR into that one, the PRs that will follow will also target `briefcase-integration-2` and there we can add tests and update tests as necessary.

I've intentionally not created any new tests given the state of this branch but I plan to add tests for these changes in the upcoming PRs.

### Checklist - did you ...

### DIFF
--- a/constructor/briefcase.py
+++ b/constructor/briefcase.py
@@ -250,7 +250,7 @@ def write_pyproject_toml(tmp_dir, info):
 
 def create(info, verbose=False):
     if not IS_WINDOWS:
-        raise Exception(f"Invalid platform '{sys.platform}'. MSI installers requires Windows.")
+        raise Exception(f"Invalid platform '{sys.platform}'. MSI installers require Windows.")
 
     tmp_dir = Path(tempfile.mkdtemp())
     write_pyproject_toml(tmp_dir, info)


constructor/briefcase.py
```diff
@@ -250,18 +250,25 @@ def write_pyproject_toml(tmp_dir, info):
 
 def create(info, verbose=False):
     if not IS_WINDOWS:
-        raise Exception(f"Invalid platform '{sys.platform}'. Only Windows is supported.")
+        raise Exception(f"Invalid platform '{sys.platform}'. MSI installers requires Windows.")
```
```suggestion
        raise Exception(f"Invalid platform '{sys.platform}'. MSI installers require Windows.")
```

constructor/briefcase.py
```diff
@@ -250,18 +250,25 @@ def write_pyproject_toml(tmp_dir, info):
...
 
     tmp_dir = Path(tempfile.mkdtemp())
     write_pyproject_toml(tmp_dir, info)
 
     external_dir = tmp_dir / EXTERNAL_PACKAGE_PATH
     external_dir.mkdir()
-    preconda.write_files(info, external_dir)
+
+    # Create the sub-directory "base",
```
marcoesters, line 261: Should this become a schema property then? Or at least, we should make this a global constant so that we can use this as a template variable later to make changes more robust.

lrandersson: Using jinja-templating for bat-files could be one approach, do we have `.bat`-files today with jinja templating?

I was also considering that we move the `.bat`-file into a Python class and generate it on demand. As an example, I did this with `pre_uninstall.bat` in a separate branch that we have not merged: https://github.com/lrandersson/constructor/blob/briefcase-uninstall/constructor/briefcase.py#L102, then instead of copying the `.bat`-file we generate it as [here](https://github.com/lrandersson/constructor/blob/briefcase-uninstall/constructor/briefcase.py#L318). This made it possible to unit test individual parts of the function without having to go through the entire install/uninstall process for every test.

marcoesters: We don't have any batch files with Jinja templating, but we can template any file we like. I think a template file is much more readable than creating a text file with Python code.

lrandersson: Understandable, it's likely that we need templating also in `pre_uninstall.bat` later and we can add that for both simultaneously, especially to account later for user provided post/pre install/uninstall script.

lrandersson: I do think the Python-based approach makes unit testing more convenient ([example](https://github.com/lrandersson/constructor/blob/briefcase-uninstall/tests/test_briefcase.py#L141), but it might be just as easy with jinja-templating as well.

marcoesters: This is getting off topic, but with the Jinja template, you can always write it into a temporary directory and read the file. That will do the same as the unit test you're showing me.

We can kick this question down the road for when extra environments (which will require a for-loop) are implemented.

marcoesters (approved):

PR #1164, opened 2026-02-04
MSI: Reorganize payload preparation

lrandersson:
### Description
This PR proposes a class `Payload` to organize the payload preparation.
I still think more work can be done on this but I tried to keep the changes minimal to simplify the review.
I have also added a couple of unit tests to demonstrate what I think is the biggest gain of using this structure over the old one.

I think this class can be extracted later and also be used for other installers, since we can make it possible (if needed) to override the function that creates the payload layout.

For the review, it might be easier to also open the whole file via "View file" and not only browsing from the "Files changed" tab.

### Checklist - did you ...

### DIFF
--- a/constructor/briefcase.py
+++ b/constructor/briefcase.py
@@ -243,7 +243,7 @@ class Payload:
     def remove(self, *, ignore_errors: bool = True) -> None:
         """Remove the root of the payload.
 
-        This function requires some extra care due to the root being a cached property.
+        This function requires some extra care due to the root directory being a cached property.
         """
         root = getattr(self, "root", None)
         if root is None:
@@ -387,6 +387,9 @@ class Payload:
 def create(info, verbose=False):
     if not IS_WINDOWS:
         raise Exception(f"Invalid platform '{sys.platform}'. MSI installers require Windows.")
+
+    if not info.get("_conda_exe_supports_logging"):
+        raise Exception("MSI installers require conda-standalone with logging support.")
 
     payload = Payload(info)
     payload.prepare()
--- a/constructor/briefcase/pre_uninstall.bat
+++ b/constructor/briefcase/pre_uninstall.bat
@@ -46,7 +46,7 @@ if errorlevel 1 (
 "%CONDA_EXE%" --log-file "%LOG%" constructor uninstall --prefix "%BASE_PATH%"
 if errorlevel 1 ( exit /b %errorlevel% )
 
-rem If we reached this far without any errors, remove any log-files.
+rem If we reached this far without any errors, remove any log files.
 if exist "%INSTDIR%\install.log" del "%INSTDIR%\install.log"
 if exist "%INSTDIR%\uninstall.log" del "%INSTDIR%\uninstall.log"
 


lrandersson:
I'm tagging @marcoesters here because I assume you will review this out of the possible reviewers. I just wanted to inform you that there is an additional PR that depends on this one, that also is in need of review. https://github.com/lrandersson/constructor/pull/2, this PR is on my own fork because I need to merge that PR into this one (PR 2 -> PR 1164).

constructor/briefcase.py
```diff
@@ -246,47 +247,98 @@ def write_pyproject_toml(tmp_dir, info):
         config["author"] = info["company"]
 
     (tmp_dir / "pyproject.toml").write_text(tomli_w.dumps({"tool": {"briefcase": config}}))
+    logger.debug(f"Created TOML file at: {tmp_dir}")
 
 
-def create(info, verbose=False):
-    if not IS_WINDOWS:
-        raise Exception(f"Invalid platform '{sys.platform}'. MSI installers require Windows.")
+@dataclass(frozen=True)
```
marcoesters: Why use a `dataclass` over `NamedTuple`?

constructor/briefcase.py
```diff
@@ -246,47 +247,98 @@ def write_pyproject_toml(tmp_dir, info):
...
+        preconda.write_files(self.info, layout.base)
+        preconda.copy_extra_files(self.info.get("extra_files", []), layout.external)
+        self._stage_dists(layout)
+        self._stage_conda(layout)
+        return layout
 
-    tmp_dir = Path(tempfile.mkdtemp())
-    write_pyproject_toml(tmp_dir, info)
+    def remove(self) -> None:
+        shutil.rmtree(self.root)
```
marcoesters: Should there be a guard in case the directory doesn't exist (or an `ignore_errors=True`)?

constructor/briefcase.py
```diff
@@ -246,47 +247,98 @@ def write_pyproject_toml(tmp_dir, info):
...
 
-    # Create the sub-directory "base",
-    # note that the directory name "base" is also explicitly
-    # defined in `run_installation.bat`
-    base_dir = external_dir / "base"
-    base_dir.mkdir()
+    def _ensure_root(self) -> Path:
+        if self.root is None:
+            self.root = Path(tempfile.mkdtemp())
+        return self.root
```
marcoesters: This should be part of `__init__` or we shouldn't allow `None` root. I don't see an advantage having a potentially "unbound" root directory.

constructor/briefcase.py
```diff
@@ -246,47 +247,98 @@ def write_pyproject_toml(tmp_dir, info):
...
+    def prepare(self) -> PayloadLayout:
+        root = self._ensure_root()
+        self._write_pyproject(root)
+        layout = self._create_layout(root)
+
+        preconda.write_files(self.info, layout.base)
+        preconda.copy_extra_files(self.info.get("extra_files", []), layout.external)
+        self._stage_dists(layout)
+        self._stage_conda(layout)
+        return layout
```
marcoesters: This part makes me think that `PayloadLayout` may be overkill. `_stage_dists` and `_stage_conda` only use one element of `PayloadLayout` and only `root` is used outside of that function, so the entire object doesn't need to be returned.

Why not combine `_create_layout` and `prepare` into one function and remove the `PayloadLayout` data class? This seems simpler for now.

If we find that we want to use this class later for other installers, we can create these interfaces later.

constructor/briefcase.py
```diff
@@ -246,47 +247,98 @@ def write_pyproject_toml(tmp_dir, info):
...
 
-    preconda.write_files(info, base_dir)
-    preconda.copy_extra_files(info.get("extra_files", []), external_dir)
+    def _create_layout(self, root: Path) -> PayloadLayout:
+        """The layout is created as:
+        root/
+        └── external/
+            └── base/
+                └── pkgs/
+        """
```
marcoesters: I think this is a good opportunity to elaborate a little more on why this directory structure, specifically the intermediary `external` directory is needed.

constructor/briefcase.py
lrandersson: I'd say that they are very similar but in my opinion this more comes down to how we use this. The class `PayloadLayout` represents one thing, and we can add methods to it as well. I tried to keep it minimal for presentation. For a tuple also positional meaning is primary, but for the frozen data class the fields are primary.

constructor/briefcase.py
lrandersson: I thought about it, but, I think it would make the most sense that the one who implements the interface is the one who should care about potential errors. Silently catching/ignoring errors here could lead to unexpected behavior.

constructor/briefcase.py
lrandersson: That is an option, the reason why I have done as above is to avoid filesystem side effects upon class instantiation. The caller ideally only need to use this as:
```
payload = Payload(info)
payload.prepare()
```

However, if we are fine with above the side effects, let me know and I will update it. It is a minor thing to change in terms of code.

constructor/briefcase.py
lrandersson: I see your point, but I think it'll be a bit more "clear" after reviewing the follow-up PR https://github.com/lrandersson/constructor/pull/2/changes where more code has been added.

I like to think about it this way, alternative 1 (which we are very close to in the PR I linked):
```
self.render_templates(layout)
self.write_pyproject_toml(layout)
self.archive(layout)
self._stage_dists(layout)
self._stage_conda(layout)
```
Compared to alternative 2:
```
external = root / "external"
base = external / "base"
pkgs = base / "pkgs"

self.render_templates(root, external)
self.write_pyproject_toml(root, external)
self.archive(base, external)
self._stage_dists(pkgs)
self._stage_conda(external)
```
And if for some reason we need to add more directories as arguments to any of the functions, one does not need to change the function definitions, only the implementation of each function.
To me it doesn't seem overkill but for maintenance and ease of implementation, alternative 1 is much simpler. Please let me know your thoughts.

constructor/briefcase.py
lrandersson: Added here https://github.com/conda/constructor/pull/1164/changes/c4b11e88b1b564bc438202de76880c31cd3c0bec, note that there is also a path to `external` in the `pyproject.toml` that is generated, but I didn't explicitly mention that in the docstring.

constructor/briefcase.py
marcoesters: I'm suggesting a `NamedTuple`, not a `tuple`. A `NamedTuple` also goes by fields and is immutable. I don't see that we will need methods for a structure that just describes a layout.

lrandersson: I see, what would the gain be to use a `NamedTuple` over a `dataclass` in this situation? I agree that we still get the data organized in a similar fashion but they signal different intent, and my thought process is:

1. I'd use a `NamedTuple` when the thing is a tuple.
2. I'd use frozen dataclass when the thing represents a thing.

constructor/briefcase.py
marcoesters: Alternative 1 already lists a few MSI-specific function calls, like preparing the pyproject.toml file (which is technically not part of the installer payload anyway). The SH format structures its payload differently (preconda, postconda), and EXE installers have `temp_extra_files`.

I am overall in favor of a more object-oriented way rather than the current scripting appraoch `constructor` have, and alternative 1 looks reasonable at first glance.

However, I am concerned about introducing premature abstraction now that we may have to undo later. I think for now, alternative 2 is good enough for MSI. When there is earnest effort in making `constructor` more OOP-based, we can take a holistic look at all payloads and find what an appropriate level of abstraction is.

constructor/briefcase.py
marcoesters: Overall, a `NamedTuple` is a smaller object. I use a `NamedTuple` when I just need a simple immutable data container. However, I'm still not fully convinced we need this (yet). See my comment on the `perpare` function below.

constructor/briefcase.py
lrandersson: By the way I have now guards for this since `Payload.root` is now a cached property in https://github.com/lrandersson/constructor/pull/2

constructor/briefcase.py
lrandersson: I see your point. Do you mind if we take this action as the last thing after merging https://github.com/lrandersson/constructor/pull/2 into this PR? That way we can also see a simple diff between the two layouts given the current implementation.
(It will also be easier for me to merge the second PR that way).

lrandersson: @marcoesters I've now added a commit that basically is the same functionality but without the class `PayloadLayout` for comparison: https://github.com/conda/constructor/pull/1164/changes/6af091000598e7597c8dd4d4524b7888fd578df3, let me know what you think.

lrandersson: Ping @marcoesters this should be the last open thread of this PR

constructor/briefcase/pre_uninstall.bat
```diff
@@ -0,0 +1,53 @@
...
+rem MSI installer expects it to exist.
+type nul > "%PAYLOAD_TAR%"
+if errorlevel 1 (
+  {{ error_block('Failed to create "%PAYLOAD_TAR%"', '%errorlevel%') }}
+)
+
+"%CONDA_EXE%" --log-file "%LOG%" constructor uninstall --prefix "%BASE_PATH%"
+if errorlevel 1 ( exit /b %errorlevel% )
+
+rem If we reached this far without any errors, remove any log-files.
```
marcoesters:
```suggestion
rem If we reached this far without any errors, remove any log files.
```

constructor/briefcase.py
```diff
@@ -218,75 +222,189 @@ def create_install_options_list(info: dict) -> list[dict]:
...
+
+    @functools.cached_property
+    def root(self) -> Path:
+        """Create root upon first access and cache it."""
+        return Path(tempfile.mkdtemp(prefix="payload-"))
+
+    def remove(self, *, ignore_errors: bool = True) -> None:
+        """Remove the root of the payload.
+
+        This function requires some extra care due to the root being a cached property.
```
```suggestion
        This function requires some extra care due to the root directory being a cached property.
```

constructor/briefcase.py
```diff
@@ -218,75 +222,189 @@ def create_install_options_list(info: dict) -> list[dict]:
...
+        """
+        root = getattr(self, "root", None)
+        if root is None:
+            return
+        shutil.rmtree(root, ignore_errors=ignore_errors)
+        # Now we drop the cached value so next access will recreate if desired
+        try:
+            delattr(self, "root")
+        except Exception:
+            # delattr on a cached_property may raise on some versions / edge cases
```
marcoesters, line 256: For example? And why is passing the appropriate course of action?

constructor/briefcase.py
marcoesters: I can't say that I miss the `PayloadLayout` class, honestly.

constructor/briefcase/pre_uninstall.bat
```diff
@@ -0,0 +1,53 @@
...
+)
+
+rem Recreate an empty payload tar. This file was deleted during installation but the
+rem MSI installer expects it to exist.
+type nul > "%PAYLOAD_TAR%"
+if errorlevel 1 (
+  {{ error_block('Failed to create "%PAYLOAD_TAR%"', '%errorlevel%') }}
+)
+
+"%CONDA_EXE%" --log-file "%LOG%" constructor uninstall --prefix "%BASE_PATH%"
```
marcoesters, line 46: Since this assumes the `--log-file` feature, I think we need to add a check to ensure that it exists. That information exists in `info`, we just need to raise when it's `False`

constructor/briefcase.py
lrandersson: There are a few ways to look at it:
1. If this fails, the temporary directory is already removed which is the primary goal.
2. The `payload.remove()` function here is intended to be called as the final step (currently inside `create(...)`)
It is likely that we could skip the entire `delattr` shenanigan, but to me it looks more complete to avoid a reference to a directory that might not exist.
I was also thinking of writing `Payload` to be used as a context-manager to achieve automatic clean-up but I thought it might be a bit overkill right now.

constructor/briefcase/pre_uninstall.bat
lrandersson: Great idea, see https://github.com/conda/constructor/pull/1164/changes/1ed9fe066f213d7f0c335742b4d36a81b2f7bd9e

constructor/briefcase.py
lrandersson: https://github.com/conda/constructor/pull/1164/changes/1ed9fe066f213d7f0c335742b4d36a81b2f7bd9e

constructor/briefcase/pre_uninstall.bat
lrandersson: https://github.com/conda/constructor/pull/1164/changes/1ed9fe066f213d7f0c335742b4d36a81b2f7bd9e

PR #1174, opened 2026-02-27
Remove unnecessary .r0 uses and fix issue with intermittent uninstall failures

lrandersson:
### Description
There have been observations of a corrupted `PATH`/`INSTDIR` during Uninstallation which have been occurring intermittently. It is not clear yet why this happens but, this PR intends to add some additional logging in order to debug this more effectively the next time it occurs, by:
1. Add additional logging output of `INSTDIR` near code that is suspect.
2. Remove unnecessary writes to `r0` which affects `$0` which itself is used frequently in the code (for example `ReadEnvStr $0 PATH`)

### Checklist - did you ...

### DIFF
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -1749,11 +1749,6 @@ Section "Uninstall"
 
     # ensure that MSVC runtime DLLs are on PATH during uninstallation
     ReadEnvStr $R7 SystemRoot
-    StrLen $9 "$R7"
-    ${If} $9 > 1024
-        ${Print} "::warning:: System PATH is very long ($9 chars) and may be truncated. \
-                Consider shortening PATH before running this uninstaller."
-    ${EndIf}
     # set PATH for the installer process, so that MSVC runtimes get found OK
     System::Call 'kernel32::SetEnvironmentVariable(t,t)i("PATH", \
                  "$INSTDIR;$INSTDIR\Library\mingw-w64\bin;$INSTDIR\Library\usr\bin;$INSTDIR\Library\bin;$INSTDIR\Scripts;$INSTDIR\bin;$R7;$R7\system32;$R7\system32\Wbem")'


constructor/nsis/main.nsi.tmpl
```diff
@@ -1833,6 +1839,7 @@ Section "Uninstall"
         call un.AbortRetryNSExecWait
     ${EndIf}
     ${If} ${FileExists} "$INSTDIR\.nonadmin"
+        ${Print} "About to modify PATH with INSTDIR: $INSTDIR"
```
lrandersson: TODO: Remove

constructor/nsis/main.nsi.tmpl
```diff
@@ -1743,24 +1743,30 @@ Section "Uninstall"
...
+    StrLen $9 "$R7"
+    ${If} $9 > 1024
+        ${Print} "::warning:: System PATH is very long ($9 chars) and may be truncated. \
+                Consider shortening PATH before running this uninstaller."
+    ${EndIf}
     # set PATH for the installer process, so that MSVC runtimes get found OK
     System::Call 'kernel32::SetEnvironmentVariable(t,t)i("PATH", \
-                 "$INSTDIR;$INSTDIR\Library\mingw-w64\bin;$INSTDIR\Library\usr\bin;$INSTDIR\Library\bin;$INSTDIR\Scripts;$INSTDIR\bin;$0;$0\system32;$0\system32\Wbem").r0'
+                 "$INSTDIR;$INSTDIR\Library\mingw-w64\bin;$INSTDIR\Library\usr\bin;$INSTDIR\Library\bin;$INSTDIR\Scripts;$INSTDIR\bin;$R7;$R7\system32;$R7\system32\Wbem")'
+    ${Print} "Path has been set with INSTDIR: $INSTDIR"
```
lrandersson: TODO: Remove

constructor/nsis/main.nsi.tmpl
```diff
@@ -1743,24 +1743,30 @@ Section "Uninstall"
     ${If} ${Silent}
         !insertmacro un.ParseCommandLineArgs
     ${EndIf}
-
+    ${Print} "Uninstallation started for INSTDIR: $INSTDIR"
```
lrandersson: TODO: Remove

news/1174-uninstall-fix
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* <news item>
+
+### Bug fixes
+
+* EXE: Fix an issue where uninstallation would sometimes fail for installations that have been added to `PATH`.
```
marcoesters: I think this needs a more detailed description. The bug here is that the wrong environment variable was read. This could lead to a truncated `PATH`, which for some reason poisoned `$INSTDIR`.

constructor/nsis/main.nsi.tmpl
```diff
@@ -1494,12 +1494,12 @@ Section "Install"
     ${Print} "${NAME} will now be installed into this location:"
     ${Print} "$INSTDIR$\n"
 
-    ReadEnvStr $0 SystemRoot
+    ReadEnvStr $R7 SystemRoot
```
marcoesters, line 1497: Are you sure we have to change the register here?

news/1174-uninstall-fix
lrandersson: I was thinking that the changelog above was more user-oriented, but let me update it, considering users of the installers will not read the `constructor` changelog.

lrandersson: See https://github.com/conda/constructor/pull/1174/changes/983d7232954f8eeedf0b00c1ce529aaf55335be2

constructor/nsis/main.nsi.tmpl
lrandersson: I don't think we have to, this was just recommended best practices to spread our use of global registers.

marcoesters: Where is that recommended/documented? The way `$0` is used at the moment seems safe enough.

lrandersson: There is no "official" statement as far as I know more than use it with caution, in particular because macros and plugins etc make use of shared registers, and in particular `$R0-$R3` are frequently used. For example in our code the comment:
```
# However, that also means that registers must be handled with caution because
# they will be overwritten in the macro. It is best to use $R* registers if
# temporary variables need to be used.
```
and I'd argue and say this is true for many of our macros such as `setInstdirPermissions`, `AddRemovePath` and this is not inherently wrong, but something that always must be done carefully.

That being said we don't have to change from register `0` to `7` here. I also thought this was an interesting related thread: https://nsis-dev.github.io/NSIS-Forums/html/t-233795.html

constructor/nsis/main.nsi.tmpl
```diff
@@ -1743,24 +1743,28 @@ Section "Uninstall"
...
+    System::Call 'kernel32::SetEnvironmentVariable(t,t)i("CONDA_ROOT_PREFIX", "$INSTDIR")'
 
     # ensure that MSVC runtime DLLs are on PATH during uninstallation
-    ReadEnvStr $0 PATH
+    ReadEnvStr $R7 SystemRoot
+    StrLen $9 "$R7"
+    ${If} $9 > 1024
+        ${Print} "::warning:: System PATH is very long ($9 chars) and may be truncated. \
+                Consider shortening PATH before running this uninstaller."
+    ${EndIf}
```
```suggestion
```

This looks like leftover code since `$R7` is `SystemRoot` now.

PR #1179, opened 2026-03-04
MSI: Implement install options

lrandersson:
### Checklist - did you ...

### DIFF
--- a/constructor/briefcase/pre_uninstall.bat
+++ b/constructor/briefcase/pre_uninstall.bat
@@ -60,7 +60,7 @@ if not exist "%CONDA_EXE%" (
 )
 if "%ALLUSERS%"=="0" (
   if not exist "%BASE_PATH%\.nonadmin" (
-    {{ error_block('The installation appears to be corrupted. Please reinstall the application.', 11) }}
+    {{ error_block('Insufficient permissions. Please re-run the uninstallation as administrator.', 11) }}
   )
 )
 


examples/protected_base/construct.yaml
```diff
@@ -3,7 +3,7 @@
 
 name: ProtectedBaseEnv
 version: 1.0.0
-installer_type: {{ "exe" if os.name == "nt" else "all" }}
+installer_type: all
```
lrandersson, line 6: This I changed because, after I rebased the feature branch towards main, this test started to fail because the `yaml.load` in the test assumes there is no need to perform any jinja rendering of the `construct.yaml` file. I made some updates also in the test to account for this.

constructor/briefcase/pre_uninstall.bat
```diff
@@ -43,7 +57,70 @@ if errorlevel 1 (
   {{ error_block('Failed to create "%PAYLOAD_TAR%"', '%errorlevel%') }}
 )
 
-"%CONDA_EXE%" --log-file "%LOG%" constructor uninstall --prefix "%BASE_PATH%"
+rem Remove shortcuts unconditionally
+echo Removing shortcuts...
+>> "%LOG%" echo Removing shortcuts...
```
marcoesters: Seems like this could be a macro.

constructor/briefcase/pre_uninstall.bat
```diff
@@ -43,7 +57,70 @@ if errorlevel 1 (
...
+"%CONDA_EXE%" constructor --prefix "%BASE_PATH%" --rm-menus --log-file "%LOG%"
```
marcoesters: This is only needed if `uninstall_with_conda_exe` is false. This is part of the regular uninstallation routine of `conda-standalone`.

constructor/briefcase/pre_uninstall.bat
```diff
@@ -43,7 +57,70 @@ if errorlevel 1 (
...
+{%- set pathflag = "--condabin" if initialize_conda == "condabin" else "--classic" %}
+if "%REG_HIVE%"=="HKCU" (
+    echo Removing from PATH...
+    >> "%LOG%" echo Removing from PATH...
+    "%CONDA_EXE%" constructor windows path --remove=user --prefix "%INSTDIR%" {{ pathflag }} --log-file "%LOG%"
+    if errorlevel 1 ( exit /b %errorlevel% )
+)
+
+{%- if has_python %}
+rem Remove Python registry entries only if InstallPath matches BASE_PATH.
```
marcoesters, line 83: I wonder if that should be in `conda-standalone` instead of using batch scripting, but it would be even more of a vendor lock. Batch scripting something this involved just never looks pretty. 😬

constructor/briefcase/run_installation.bat
```diff
@@ -46,17 +56,34 @@ if not exist "%PAYLOAD_TAR%" (
 )
 
 echo Unpacking payload...
-"%CONDA_EXE%" --log-file "%LOG%" constructor extract --prefix "%INSTDIR%" --tar-from-stdin < "%PAYLOAD_TAR%"
+>> "%LOG%" echo Unpacking payload...
```
marcoesters: Same here - this could be a macro.

constructor/briefcase/run_installation.bat
```diff
@@ -46,17 +56,34 @@ if not exist "%PAYLOAD_TAR%" (
...
+"%CONDA_EXE%" constructor extract --prefix "%BASE_PATH%" --conda-pkgs --log-file "%LOG%"
 if errorlevel 1 ( exit /b %errorlevel% )
 
 if not exist "%BASE_PATH%" (
   {{ error_block('"%BASE_PATH%" not found!', 12) }}
 )
 
-"%CONDA_EXE%" --log-file "%LOG%" install --offline --file "%BASE_PATH%\conda-meta\initial-state.explicit.txt" -yp "%BASE_PATH%"
+rem Create .nonadmin marker file for user-scoped installs
+rem This is used by the uninstaller to determine the install mode
```
marcoesters: This needs to go into the `base` environment so that `menuinst` knows, too.

tests/test_examples.py
```diff
@@ -1742,10 +1743,16 @@ def test_frozen_environment(tmp_path, request, has_conflict):
     with open(input_path / "construct.yaml") as f:
         config = yaml.load(f)
 
+    # Since the above yaml.load does not rely on jinja rendering,
```
marcoesters, line 1746: It sounds like we need a `render` function at some point.

constructor/briefcase/pre_uninstall.bat
lrandersson: https://github.com/conda/constructor/pull/1179/changes/26f3fc98639a968dbdfa623415c7f05311f90644

constructor/briefcase/pre_uninstall.bat
lrandersson: Good catch, I wrote it initially as it always would be the case. I had to add some placeholders to be able to mirror what we do in NSIS, please see https://github.com/conda/constructor/pull/1179/changes/26f3fc98639a968dbdfa623415c7f05311f90644.

constructor/briefcase/pre_uninstall.bat
lrandersson: Yeah, it would be nice to "hide" it from the bat-file situation. I think it is something we should at least consider for future changes.

constructor/briefcase/run_installation.bat
lrandersson: https://github.com/conda/constructor/pull/1179/changes/26f3fc98639a968dbdfa623415c7f05311f90644

constructor/briefcase/run_installation.bat
lrandersson: Good catch, added a comment also https://github.com/conda/constructor/pull/1179/changes/26f3fc98639a968dbdfa623415c7f05311f90644

constructor/briefcase/pre_uninstall.bat
```diff
@@ -43,8 +62,88 @@ if errorlevel 1 (
...
+
+:after_remove_python_registry
+{%- endif %}
+
+rem Remove .nonadmin marker file if it exists
+if exist "%BASE_PATH%\.nonadmin" (
+    {{ tee("Removing .nonadmin marker file...") }}
+    del "%BASE_PATH%\.nonadmin"
+    if errorlevel 1 ( exit /b %errorlevel% )
+)
```
marcoesters: This should be done by the uninstallation, not manually or you'll change the scope for `menuinst`.

constructor/briefcase/pre_uninstall.bat
```diff
@@ -43,8 +62,88 @@ if errorlevel 1 (
...
+
+{%- if uninstall_with_conda_exe %}
+rem Run constructor uninstall, conditionally passing optional flags
+set "UNINST_ARGS="
+if "%OPTION_REMOVE_USER_DATA%"=="1" (
+    set "UNINST_ARGS=!UNINST_ARGS! --remove-user-data"
+)
+if "%OPTION_REMOVE_CACHES%"=="1" (
+    set "UNINST_ARGS=!UNINST_ARGS! --remove-caches"
+)
```
marcoesters, line 119: We also need `--remove-config-files`. Compare here: https://github.com/conda/constructor/blob/74cb5eb8c09a23ef7d8e874d96740a83a7d54453/constructor/nsis/main.nsi.tmpl#L1842-L1851

constructor/briefcase/pre_uninstall.bat
```diff
@@ -43,8 +62,88 @@ if errorlevel 1 (
...
+{{ tee("Running constructor uninstall...") }}
+"%CONDA_EXE%" constructor uninstall --prefix "%BASE_PATH%"!UNINST_ARGS! --log-file "%LOG%"
 if errorlevel 1 ( exit /b %errorlevel% )
+{%- else %}
+rem Remove menus for each environment.
+rem TODO: loop over extra_envs when extra_envs support is implemented for MSI.
+{%- for env in setup_envs %}
+{{ tee("Removing menus for " + env.name + "...") }}
+"%CONDA_EXE%" constructor --prefix "{{ env.prefix }}" --rm-menus --log-file "%LOG%"
```
marcoesters, line 136: You should double-check, but I think `--root-prefix %BASE_PATH%` needs to be set if you don't set `%CONDA_ROOT_PREFIX%`, as is done here: https://github.com/conda/constructor/blob/74cb5eb8c09a23ef7d8e874d96740a83a7d54453/constructor/nsis/main.nsi.tmpl#L1544

This may also be necessary for the uninstallation with `conda-standalone`.

constructor/briefcase/pre_uninstall.bat
```diff
@@ -43,8 +62,88 @@ if errorlevel 1 (
...
+) else (
+    set "CONDA_INIT_SCOPE=system"
+)
+{{ tee("Reversing conda shell initialization...") }}
+"%BASE_PATH%\condabin\conda.bat" init cmd.exe --reverse --!CONDA_INIT_SCOPE! --log-file "%LOG%"
+if errorlevel 1 ( exit /b %errorlevel% )
+{%- endif %}
+
+rem Remove the base conda environment. INSTDIR itself is cleaned up by the MSI engine.
+{{ tee("Removing base environment...") }}
```
marcoesters:
```suggestion
{{ tee("Removing environments...") }}
```

Since this also removes all sub-environments

constructor/briefcase/pre_uninstall.bat
```diff
@@ -43,8 +62,88 @@ if errorlevel 1 (
...
+)
+if "%OPTION_REMOVE_CACHES%"=="1" (
+    set "UNINST_ARGS=!UNINST_ARGS! --remove-caches"
+)
+{{ tee("Running constructor uninstall...") }}
+"%CONDA_EXE%" constructor uninstall --prefix "%BASE_PATH%"!UNINST_ARGS! --log-file "%LOG%"
 if errorlevel 1 ( exit /b %errorlevel% )
+{%- else %}
+rem Remove menus for each environment.
+rem TODO: loop over extra_envs when extra_envs support is implemented for MSI.
```
marcoesters:
```suggestion
```

The `extra_envs` are already in `setup_envs` eventually. I would add a TODO item to the installation though.

constructor/briefcase/pre_uninstall.bat
lrandersson: So I am realizing we will not be able to mimic the exact same behavior here for the MSI installer. Correct me if I am wrong but in the NSIS situation, there are actually two-checkboxes but one is hidden based on the install scope? For the MSI installer it doesn't seem like we can conditionally display checkboxes during runtime and we will be forced to display two buttons regardless of the install scope.
It will look a bit ugly, any thoughts on this? I was thinking we will have to display two buttons but one will do nothing (of course depending on the install scope).

constructor/briefcase/pre_uninstall.bat
lrandersson: Good catch, added here https://github.com/conda/constructor/pull/1179/changes/dc06006d368161c0d5beb87ffdf9747c8f9249d8

constructor/briefcase/pre_uninstall.bat
lrandersson: https://github.com/conda/constructor/pull/1179/changes/5c90129d479263440b888542f5e9ceb94e3100f5

constructor/briefcase/pre_uninstall.bat
lrandersson: https://github.com/conda/constructor/pull/1179/changes/5c90129d479263440b888542f5e9ceb94e3100f5

constructor/briefcase/pre_uninstall.bat
lrandersson: Also after some further reading, it _should_ be possible to implement conditionally shown buttons on the `briefcase` side.

marcoesters: > Correct me if I am wrong but in the NSIS situation, there are actually two-checkboxes but one is hidden based on the install scope?

That's correct. This is because a regular user shouldn't even attempt to delete system-wide files.

lrandersson: Thanks for confirming that @marcoesters, as it is now we will not able to conditionally display the buttons, we will have to display both buttons regardless of install scope. Are we OK with this? I'm thinking that we can at least add text to one of them that it is only applicable if the installation was set "for everyone".  Let me know if you have any other idea.

marcoesters: I think it's easier to only display one option where admins remove both and users only remove one. That prevents accidental clicks and permission problems.

lrandersson: See https://github.com/conda/constructor/pull/1179/changes/6b61ea26387f1715436961f666f5b45ad93c28b8

constructor/briefcase/pre_uninstall.bat
lrandersson: You are right that was at least part of the NSIS code, I have also added it here now.
https://github.com/conda/constructor/pull/1179/changes/6b61ea26387f1715436961f666f5b45ad93c28b8
It also made it more consistent with `run_installation.bat` because we did also set it there.

lrandersson: Also changed the check for consistency:
https://github.com/conda/constructor/pull/1179/changes/9dd5fb9acfa68a6007631802057a742da6bbf7c5

constructor/briefcase/pre_uninstall.bat
lrandersson: @marcoesters one more consistency check https://github.com/conda/constructor/pull/1179/changes/ab490f5b9cbb47ba79206058d62c85d8858811bb

constructor/briefcase/pre_uninstall.bat
```diff
@@ -15,26 +24,45 @@ set "BASE_PATH=%INSTDIR%\base"
...
 "%CONDA_EXE%" --version >> "%LOG%" 2>&1
 {%- endif %}
 
 rem Consistency checks
 if not exist "%CONDA_EXE%" (
   {{ error_block('CONDA_EXE not found: "%CONDA_EXE%"', 10) }}
 )
+if "%ALLUSERS%"=="0" (
+  if not exist "%BASE_PATH%\.nonadmin" (
+    {{ error_block('The installation appears to be corrupted. Please reinstall the application.', 11) }}
```
marcoesters:
```suggestion
    {{ error_block('Insufficient permissions. Please re-run the uninstallation as administrator.', 11) }}
```

PR #1182, opened 2026-03-13
MSI: Add support for virtual_specs, write_condarc and condarc

lrandersson:
### Changes

- Extracted `get_condarc_content()` function from `add_condarc()` to allow reuse by MSI installers
- Added `virtual_specs`, `virtual_specs_debug`, and `condarc_content` to template context
- Added virtual specs compatibility check (dry-run with `CONDA_SOLVER=classic` workaround) and `.condarc` file writing block
- Added tests for `virtual_specs` and `condarc` rendering
- Added tests for `get_condarc_content()` and `add_condarc()`
  - I thought this could be good since there were basically no tests for this.

### DIFF
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -1648,9 +1648,10 @@ Section "Install"
 {%- endif %}
 {%- endfor %}
 
-{%- for condarc in WRITE_CONDARC %}
-    {{ condarc }}
-{%- endfor %}
+{%- if condarc_file %}
+    SetOutPath "$INSTDIR"
+    File {{ condarc_file }}
+{%- endif %}
 
     AddSize {{ SIZE }}
 
--- a/constructor/preconda.py
+++ b/constructor/preconda.py
@@ -343,7 +343,7 @@ def write_condarc(info: dict, dst_dir: str):
     to the correct location during installation.
     """
     condarc = get_condarc_content(info)
-    if condarc is None:
+    if not condarc:
         return
     with open(join(dst_dir, ".condarc"), "w") as f:
         f.write(condarc)
--- a/constructor/utils.py
+++ b/constructor/utils.py
@@ -174,30 +174,6 @@ def get_condarc_content(info) -> str | None:
     return condarc
 
 
-def add_condarc(info):
-    """
-    Generate platform-specific commands to write the .condarc file.
-
-    For NSIS (Windows EXE), yields NSIS script commands.
-    For shell scripts, yields shell commands.
-    """
-    condarc = get_condarc_content(info)
-    if condarc is None:
-        return
-    yield "# ----- add condarc"
-    if info["_platform"].startswith("win"):
-        yield "Var /Global CONDARC"
-        yield 'FileOpen $CONDARC "$INSTDIR\\.condarc" w'
-        for line in condarc.splitlines():
-            yield 'FileWrite $CONDARC "%s$\\r$\\n"' % line
-        yield "FileClose $CONDARC"
-    else:
-        yield 'cat <<EOF >"$PREFIX/.condarc"'
-        for line in condarc.splitlines():
-            yield line
-        yield "EOF"
-
-
 def ensure_transmuted_ext(info, url):
     """
     If transmuting, micromamba won't find the dist in the preconda tarball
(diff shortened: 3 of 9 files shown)


constructor/briefcase/run_installation.bat
```diff
@@ -71,6 +71,19 @@ if not exist "%BASE_PATH%" (
   {{ error_block('"%BASE_PATH%" not found!', 12) }}
 )
 
+{%- if virtual_specs %}
+rem Check virtual specs compatibility before proceeding with installation.
+rem We need to specify CONDA_SOLVER=classic to work around this bug:
+rem https://github.com/conda/conda-libmamba-solver/issues/480
```
lrandersson, line 77: This comment is a copy paste from the NSIS code

tests/test_examples.py
```diff
@@ -339,11 +340,14 @@ def _sentinel_file_checks(example_path, install_dir):
...
-    """This is a temporary solution for now since we cannot choose the install location ourselves.
-    Installers are named <name>-<version>-Windows-x86_64.msi.
+def calculate_msi_install_path(config_path: Path) -> Path:
+    """Calculate the MSI install path from the construct.yaml config.
+
+    MSI installers use '<name> <version>' as the install directory name,
+    matching the formal_name set in briefcase.py.
     """
-    dir_name = installer.name.replace("-Windows-x86_64.msi", "").replace("-", " ")
+    config = parse_construct(str(config_path), platform="win-64")
```
lrandersson, line 349: The previous implementation did not work when the version number contained a build number increment, the new approach is intended to be more robust and so far works as expected.

constructor/utils.py
```diff
@@ -132,7 +132,14 @@ def if_repl(match):
     return if_pat.sub(if_repl, data)
 
 
-def add_condarc(info):
+def get_condarc_content(info) -> str | None:
```
marcoesters, line 135: Should this just be added as a file the same we way deliver the `history` file, for example? That way, we don't need to have a separate function.

lrandersson: That's a nice idea but I wonder if we should just go with it as it is and then change this for all installers at once in a separate ticket? That way we will unify the behavior between them as well (if I am not missing anything about how it works today).

lrandersson: @marcoesters I looked into this a bit further, maybe its a smaller change than I expected. Are you thinking perhaps that we in for example `preconda.py` add a `write_condarc` function that writes the file? If that's the case I think it's a good idea, we will get rid of complexity for example in the MSI bat-files (and more).

I could make a separate PR for the main branch, once merged I can add it also for the MSI context.

marcoesters: That's what I'm thinking, yes.

lrandersson: I did this in two steps:
1. Add the new approach for MSI: https://github.com/conda/constructor/pull/1182/changes/673aa974629c4a52735f24fbb75aa1276834c01b
2. Update the other installer types to account for the changes in 1): https://github.com/conda/constructor/pull/1182/changes/26f87f81a8cd4f13c5a6a9ad02463a66937ac338
Let me know if this is what you had in mind or if I missed something.

constructor/nsis/main.nsi.tmpl
```diff
@@ -1510,6 +1510,9 @@ Section "Install"
     SetOutPath "$INSTDIR"
     File {{ conda_exe }}
     File {{ pre_uninstall }}
+{%- if condarc_file %}
+    File {{ condarc_file }}
+{%- endif %}
```
marcoesters: Let's keep it where it is. The installation doesn't need the `.condarc` file.

constructor/preconda.py
```diff
@@ -332,6 +336,19 @@ def write_shortcuts_txt(info: dict, dst_dir: str, env_config: dict):
...
 
 
+def write_condarc(info: dict, dst_dir: str):
+    """Write .condarc file to the workspace if configured.
+
+    The file will be included in the payload tarball and extracted
+    to the correct location during installation.
+    """
+    condarc = get_condarc_content(info)
+    if condarc is None:
```
marcoesters: Or `if not condarc`? We don't want empty strings either.

lrandersson: Good catch, see https://github.com/conda/constructor/pull/1182/changes/9dd4ad2db602fe4beae8636b04d289308ad71523

constructor/nsis/main.nsi.tmpl
lrandersson: See https://github.com/conda/constructor/pull/1182/changes/26f87f81a8cd4f13c5a6a9ad02463a66937ac338

PR #1185, opened 2026-03-19
Do not error when PKG installer cannot change file ownership in HOME

### LINKED ISSUE #1177: Alert user when `~/.conda` has incompatible permissions.
### Checklist

### What is the idea?

The idea is to report to the user that the permissions for the `~/.conda` directory and/or its contents need to be corrected before installation begins. This would be similar to notifying the user that the destination directory already exists and aborting the install.

### Why is this needed?

To make more visible the recommended action to the end user and reduce unnecessary issue tickets to developers.

### What should happen?

Users should be notified of incorrect permissions for `~/.conda` and/or its contents and the recommended action.

### Additional Context

If `~/.conda` has incompatible permissions, the `constructor`-based installer fails on installing a secondary environment without providing sufficient information to an end user. The install log does report (for example):
```
Feb 28 14:01:24 ... ./postinstall: chown: /Users/<user>/.conda/environments.txt: Operation not permitted
Feb 28 14:01:24 ... ./postinstall: chown: /Users/<user>/.conda: Operation not permitted
Feb 28 14:01:24 ... installd[23800]: PackageKit: Install Failed: Error Domain=PKInstallErrorDomain Code=112 "An error occurred while running scripts from the package “Spyder-macOS-arm64.pkg”."
```
However, non-expert users will not be aware of the error or the recommended remediation.

Currently, `constructor` reports to the user when the destination installation directory already exists and then aborts the installation. A similar notification regarding the permission state of `~/.conda` directory and/or contents is desirable.

See spyder-ide/spyder#25767

marcoesters:
### Description

Due to a previous bug in past PKG installers, it is possible that the owner on some of the installer-generated files in the `$HOME` directory is `root`. This must be fixed by hand and cannot be done by the installer. However, at the point where ownership is being changed, the installation of all files is already finished. The installer should not error out at this point anymore.

Instead, it should just raise a warning in the logs.

Closes #1177

### Checklist - did you ...

### DIFF
--- a/constructor/osx/run_conda_init.sh
+++ b/constructor/osx/run_conda_init.sh
@@ -41,7 +41,7 @@ if [[ "${USER}" != "root" ]]; then
                 OWNER=$(stat -f "%u" "${file}" | id -un)
                 if [[ "${OWNER}" == "root" ]]; then
                     if ! chown "${USER}" "${file}"; then
-                        MSG="WARNING! Unable to change ownership of ${file}."
+                        MSG="WARNING! Unable to change ownership of ${file} (owned by ${OWNER})."
                         logger -p "install.warning" "${MSG}" || echo "${MSG}"
                     fi
                 fi
--- a/constructor/osx/run_installation.sh
+++ b/constructor/osx/run_installation.sh
@@ -125,14 +125,16 @@ chown -R "${USER}" "$PREFIX"
 CONDA_DIR="${HOME}/.conda"
 if [[ -d "${CONDA_DIR}" ]]; then
     if ! chown -R "${USER}" "${CONDA_DIR}"; then
-        MSG="WARNING: Unable to change ownership of ${CONDA_DIR}."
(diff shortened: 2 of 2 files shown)


constructor/osx/run_conda_init.sh
```diff
@@ -40,7 +40,10 @@ if [[ "${USER}" != "root" ]]; then
             if [[ -f "${file}" ]] || [[ -d "${file}" ]]; then
                 OWNER=$(stat -f "%u" "${file}" | id -un)
                 if [[ "${OWNER}" == "root" ]]; then
-                    chown "${USER}" "${file}"
+                    if ! chown "${USER}" "${file}"; then
+                        MSG="WARNING! Unable to change ownership of ${file}."
```
lrandersson: I wonder if we should add a "why" also to the message? For example by appending "... not owned by ${USER}"?

Changes look good to me, but if I saw that message I would like to know why.

PR #1186, opened 2026-03-20
MSI: Add support for extra_envs

lrandersson:
### Description

Added MSI support for `extra_envs` installations.

`constructor/briefcase.py`
- Added _setup_envs_commands(info): builds environment configs (base + extra_envs) with name, prefix, lockfile, channels, and shortcuts
- Updated _stage_dists() to include packages from _extra_envs_info
- Updated render_templates() to use _setup_envs_commands()

`constructor/briefcase/run_installation.bat`
- Added install_env(env) Jinja macro for per-environment install logic
- Replaced single install block with loop over setup_envs

### Checklist - did you ...

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -25,7 +25,7 @@ requirements:
     - conda >=4.6
     - python  # >=3.10
     - ruamel.yaml >=0.11.14,<0.19
-    - conda-standalone >=24.1.2
+    - conda-standalone >=24.11.0
     - jinja2
     - jsonschema >=4
     - pillow >=3.1     # [win or osx]


constructor/briefcase/run_installation.bat
```diff
@@ -15,6 +15,17 @@ echo {{ message }}
 >> "%LOG%" echo {{ message }}
 {%- endmacro %}
 
+{% macro install_env(env) %}
```
lrandersson, line 18: I went with the macro approach because it seems easier to read, slightly.

constructor/main.py
```diff
@@ -48,6 +48,19 @@ def get_installer_type(info: dict):
         return os_allowed[osname][:1]
     elif itype == "all":
         return os_allowed[osname]
+    elif isinstance(itype, (list, tuple)):
```
lrandersson, line 51: This I added because when I first added `msi` to `installer_types` to the example `outputs` (further down), an uncaughted exception was raised.

constructor/briefcase/pre_uninstall.bat
```diff
@@ -136,31 +135,6 @@ if "%OPTION_REMOVE_CONFIG_FILES%"=="1" (
 {{ tee("Running constructor uninstall...") }}
 "%CONDA_EXE%" constructor uninstall --prefix "%BASE_PATH%"!UNINST_ARGS! --log-file "%LOG%"
 if errorlevel 1 ( exit /b %errorlevel% )
-{%- else %}
```
lrandersson, line 139: Removed all of this because I enforced `info["uninstall_with_conda_exe"] = True` in `briefcase.py`.

constructor/briefcase.py
```diff
@@ -243,7 +283,7 @@ def create_install_options_list(info: dict) -> list[dict]:
                 "name": "enable_shortcuts",
                 "title": "Create shortcuts",
                 "description": "Create shortcuts (supported packages only).",
-                "default": False,
+                "default": True,
```
lrandersson, line 286: Now it matches NSIS.

constructor/briefcase.py
```diff
@@ -489,6 +528,11 @@ def create(info, verbose=False):
     if not info.get("_conda_exe_supports_logging"):
         raise Exception("MSI installers require conda-standalone with logging support.")
 
+    # MSI installers always use conda-standalone for uninstallation.
+    # This ensures proper cleanup of conda init, environments, and shortcuts
+    # via the `conda constructor uninstall` command.
+    info["uninstall_with_conda_exe"] = True
```
marcoesters, line 533: It needs to be verified that the `conda-standalone` version supports this, or we need to bump the minimum required version for `conda-standalone`.

constructor/briefcase.py
```diff
@@ -475,7 +509,12 @@ def write_pyproject_toml(self, root: Path, external: Path) -> None:
 
     def _stage_dists(self, pkgs_dir: Path) -> None:
         download_dir = Path(self.info["_download_dir"])
-        for dist in self.info["_dists"]:
+        # Collect dists from base and extra_envs, de-duplicated
+        dists = self.info["_dists"].copy()
+        for env_info in self.info.get("_extra_envs_info", {}).values():
+            dists += env_info["_dists"]
+        dists = list({dist: None for dist in dists})  # de-duplicate
```
marcoesters: Why not use a `set` instead?

constructor/main.py
```diff
@@ -48,6 +48,19 @@ def get_installer_type(info: dict):
...
+        # Handle list of installer types, e.g. [exe, msi]
+        for t in itype:
+            if t not in all_allowed:
+                all_allowed_str = ", ".join(sorted(all_allowed))
+                sys.exit("Error: invalid installer type '%s'; allowed: %s" % (t, all_allowed_str))
```
marcoesters, line 56: I see that this is just following what was set below, but is the distinction between `all_allowed` and `os_allowed` really useful?

lrandersson: I think this only depends on how accurate we want the error messages to be, for example:
1. `if t not in all_allowed`: "invalid installer type 'foo'" (not a valid type anywhere)
2. `if t not in os_allowed[osname]`: "invalid installer type 'pkg' for win" (valid type, wrong OS)

I think both error messages do the job but I'd say the distinction makes it a little bit more helpful.

constructor/briefcase.py
lrandersson: We could definitely use a set here, but a dict here preserves the insertion order. The functionality is the same but for debugging purposes and reproducible builds/logs I'd say a dict provides with a slight advantage.

constructor/briefcase.py
lrandersson: Hmm, do you have any preference on which approach would be the best here? I'm leaning towards bumping the minimum required version and take advantage of requiring all the new nice features.

constructor/briefcase.py
marcoesters: Does insertion order matter here?

For debugging or reproducible logs, I would find a sorted list (which you can get from a set) more useful and would make the code more legible/semantic.

constructor/briefcase.py
marcoesters: What is the minimum version that we need to bump it to?

lrandersson: `24.11.0` and I think this would be reasonable considering we have this already https://github.com/conda/constructor/blob/44ba18101752d25e26efd10d25323fe202a0accc/constructor/main.py#L218

constructor/briefcase.py
lrandersson: Good idea, see ee96eec

constructor/briefcase.py
marcoesters: That version check is for exactly that feature, but requiring at least `24.11.0` seems reasonable enough.

constructor/briefcase.py
```diff
@@ -475,7 +509,11 @@ def write_pyproject_toml(self, root: Path, external: Path) -> None:
 
     def _stage_dists(self, pkgs_dir: Path) -> None:
         download_dir = Path(self.info["_download_dir"])
-        for dist in self.info["_dists"]:
+        # Collect dists from base and extra_envs, de-duplicated
+        dists = self.info["_dists"].copy()
+        for env_info in self.info.get("_extra_envs_info", {}).values():
+            dists += env_info["_dists"]
+        for dist in sorted(set(dists)):
```
```suggestion
        dists = set(self.info["_dists"])
        for env_info in self.info.get("_extra_envs_info", {}).values():
            dists.update(env_info.get("_dists", []))
        for dist in sorted(dists):
```

constructor/briefcase.py
lrandersson: See e1911f5

PR #1197, opened 2026-04-03
Expose uninstallation options as environment variables

marcoesters:
### Description

Expose uninstallation options as environment variables for EXE installers. This is useful for when packages write into locations outside of the installation directory and outside directories associated with `conda` (e.g., the `%USERPROFILE%\.anaconda` directory). Exposing these environment variables allows for pre-uninstall scripts to clean up these directories.

Currently only implemented for EXE installers since they are the only installers that allow pre-uninstall scripts. 

### Checklist - did you ...

### DIFF
--- a/constructor/data/construct.schema.json
+++ b/constructor/data/construct.schema.json
@@ -1072,7 +1072,7 @@
         }
       ],
       "default": null,
-      "description": "Path to a pre uninstall script. This is only supported on Windows, and must be a `.bat` file. Installation path is available as `%PREFIX%`. Metadata about the installer can be found in the `%INSTALLER_NAME%`, `%INSTALLER_VER%`, `%INSTALLER_PLAT%` environment variables. `%INSTALLER_TYPE%` is set to `EXE`.",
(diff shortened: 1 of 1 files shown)


constructor/nsis/main.nsi.tmpl
```diff
@@ -1842,19 +1842,24 @@ Section "Uninstall"
...
     ${If} $UninstRemoveUserData_State == ${BST_CHECKED}
-    StrCpy $R0 "$R0 --remove-user-data"
+        StrCpy $R0 "$R0 --remove-user-data"
+        System::Call 'kernel32::SetEnvironmentVariable(t,t)i("UNINSTALLER_REMOVE_USER_DATA", "1")'
     ${EndIf}
 
     ${If} $UninstRemoveCaches_State == ${BST_CHECKED}
-    StrCpy $R0 "$R0 --remove-caches"
+        StrCpy $R0 "$R0 --remove-caches"
+        System::Call 'kernel32::SetEnvironmentVariable(t,t)i("UNINSTALLER_REMOVE_CACHES", "1")'
```
lrandersson, line 1862: Is there any need to set these to 0 if the condition is not fulfilled or does it suffice to have it as unset?

marcoesters: I think a check for `== '1'` or `!= '1'` is sufficient here. Since it involves removing files, I think only checking for the affirmative is good practice.

marcoesters:
> > > Looks good, should we document that these environment variables exist? Perhaps we can do this separately when the MSI work is merged because they will exist but with different names. With that being said, I wonder if we should name these options with a similar pattern as in the MSI scenario:
> > > https://github.com/conda/constructor/blob/d2a6fa2f42c59db31fb75d554dc3d9a288698bc4/constructor/briefcase/pre_uninstall.bat#L67
> > 
> > 
> > They should have the same name for simplicity. I personally prefer the `UNINSTALLER_*` because it shows the scope of the variables. I added them to the documentation.
> 
> I agree with that, unfortunately the naming pattern is part of `briefcase` and not something we can modify within `constructor`.

Then we have two options: call the variables the same in NSIS as they are in `briefcase`, or set `UNINSTALLER_REMOVE_CACHES=%OPTION_REMOVE_CACHES%`. I think the latter is actually needed because of `--remove-config-files`: https://github.com/conda/constructor/blob/d2a6fa2f42c59db31fb75d554dc3d9a288698bc4/constructor/briefcase/pre_uninstall.bat#L142-L150

Issue #1213, opened 2026-04-13
"pkgs" folder persists with conda_trash files after installation with "Clear Package Cache" enabled for Windows 2022


heksein:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

STR:
1. Install Miniconda/Anaconda with the `clear package cache` advanced option selected
2. Once installation is complete, check the install folder for the `pkgs` folder

ER: `pkgs` folder should not be present at all
AR: `pkgs` folder persists with `conda_trash` files 

<img width="1135" height="474" alt="Image" src="https://github.com/user-attachments/assets/3a961497-8306-46b9-98de-db4d480dc898" />
<img width="752" height="433" alt="Image" src="https://github.com/user-attachments/assets/222e3c24-c6c9-4a32-8bb6-3d5365014504" />

### Additional Context

Issue is reproducing on Windows 2022 only.

PR #1214, opened 2026-04-14
EXE: Add additional clean-up of pkgs dir if needed

lrandersson:
### Description

Proposed fix for https://github.com/conda/constructor/issues/1213, this fix relies on:
- `RMDIR /Q /S` - CMD-based recursive deletion, where: `/Q` = quiet, `/S` = recursive

### Checklist - did you ...

### DIFF
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -1679,7 +1679,7 @@ Section "Install"
         push 'WithLog'
         call AbortRetryNSExecWait
         # conda clean may leave behind .conda_trash files or locked files (e.g. vc_runtime DLLs).
-        # Force remove any leftovers using the same two-phase approach as uninstallation.
+        # Force remove any leftovers, note that locked files may still remain
         # See https://github.com/conda/constructor/issues/1213
         ${If} ${FileExists} "$INSTDIR\pkgs"
             nsExec::Exec 'cmd.exe /D /C RMDIR /Q /S "$INSTDIR\pkgs"'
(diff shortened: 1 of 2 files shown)


constructor/nsis/main.nsi.tmpl
```diff
@@ -1678,6 +1678,14 @@ Section "Install"
         push 'Failed to clear package cache'
         push 'WithLog'
         call AbortRetryNSExecWait
+        # conda clean may leave behind .conda_trash files or locked files (e.g. vc_runtime DLLs).
+        # Force remove any leftovers using the same two-phase approach as uninstallation.
+        # See https://github.com/conda/constructor/issues/1213
+        ${If} ${FileExists} "$INSTDIR\pkgs"
+            nsExec::Exec 'cmd.exe /D /C RMDIR /Q /S "$INSTDIR\pkgs"'
+            # Fallback: schedule any remaining locked files for deletion on reboot
+            RMDir /r /REBOOTOK "$INSTDIR\pkgs"
```
```suggestion
```

I'm not sure we want that fallback. Asking for a reboot after an installation sounds like bad UX.

lrandersson: We don't ask for a reboot with this, the user (who runs the installer) will not notice this at all. It only flags the files that were not possible to remove to be removed at the next reboot. From the documentation:
```
If /REBOOTOK is specified, any file or directory which could not have been removed during the process will be removed on reboot -- if any file or directory will be removed on a reboot, the reboot flag will be set.
```
But let me know if you still think it's a bad idea, I thought it was good with the reasoning above.

marcoesters: For the uninstaller at least, the user will still get the prompt to reboot the computer at the end. For a clean-up operation that may just clean up a file or two, I'm not sure this is a good workflow. I'm not sure this actually works for single-user installations since the registry entry is in `HKLM`.

lrandersson: Hmm, I see, thanks for the input. I'll go ahead and add the suggestion!

marcoesters (approved):

PR #1215, opened 2026-04-15
fix: skip mkdir ~/.conda when register_envs is false

chrisburr:
### Description

In https://github.com/DIRACGrid/DIRACOS2/issues/174 we saw that the unconditional `mkdir ~/.conda` at the top of the base-environment install section fails under read-only `$HOME`. As this installer disables env registration it does not need `~/.conda/`, so gate the `mkdir` on `register_envs` being `true`.

### Checklist - did you ...

### DIFF
--- a/constructor/header.sh
+++ b/constructor/header.sh
@@ -599,7 +599,9 @@ export FORCE
 # Avoid silent errors when $HOME is not writable
 # https://github.com/conda/constructor/pull/669
 #}
+{%- if register_envs == "true" %}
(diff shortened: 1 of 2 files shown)


lrandersson (approved):
Thanks for working on this, I think it looks good! I'll let @marcoesters have the final word in case there is something that I might have missed.

marcoesters (approved):

chrisburr:
Thanks for the prompt review!

PR #1220, opened 2026-04-23
MSI: Enable protected base

lrandersson:
### Checklist - did you ...

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -1795,9 +1795,12 @@ def test_frozen_environment(tmp_path, request, has_conflict):
         for installer, install_dir in create_installer(input_path, tmp_path):
             _run_installer(input_path, installer, install_dir, request=request, uninstall=False)
 
+            # MSI installers use a 'base' subdirectory for the conda environment
+            prefix = install_dir / "base" if installer.suffix == ".msi" else install_dir
+
             expected_frozen = {
-                install_dir / "conda-meta" / "frozen": config["freeze_base"]["conda"],
-                install_dir / "envs" / "env1" / "conda-meta" / "frozen": config["extra_envs"][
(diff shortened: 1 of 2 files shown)


tests/test_examples.py
```diff
@@ -1776,11 +1776,6 @@ def test_frozen_environment(tmp_path, request, has_conflict):
     with open(input_path / "construct.yaml") as f:
         config = yaml.load(f)
 
-    # Since the above yaml.load does not rely on jinja rendering,
```
lrandersson: After analyzing the code and enabling the test we could see that it should work "out of the box" ever since the support for `extra_envs` was added. Since `preconda.py:write_files` handles the necessary logic for protected base.

marcoesters: I'm surprised this works since you don't use `--override-frozen` in your install command: https://github.com/conda/constructor/blob/c9cb1bd0161ccaee89599fd57b237feffaeb1c8e/constructor/briefcase/run_installation.bat#L21-L25

Is the `frozen` file extracted later?

lrandersson: Hmm, maybe I have missed something (?), as I understand `preconda.write_files()` creates all the metadata files (including the frozen file (freeze marker file?)) in the workspace so the frozen file is already part of the file `payload.tar.gz`, but in `run_installation.bat` we set `set CONDA_PROTECT_FROZEN_ENVS=0` during install time, so I assume this is why it currently works.
Should I remove `set CONDA_PROTECT_FROZEN_ENVS=0` and instead explicitly set `--override-frozen`? And are they equivalent?

marcoesters: Right, we use the environment variable, thanks for pointing that out!

marcoesters (approved):

PR #1224, opened 2026-04-24
MSI: Add support for signing of installers 

lrandersson:
### Description

Adds code signing support for MSI installers built via Briefcase, reusing the existing EXE signing infrastructure.

- Add `sign()` method to `WindowsSignTool` and `AzureSignTool` for direct file signing (post-build)
- Extract signing tool initialization into `create_signing_tool()` factory to reduce duplication
- Sign MSI after Briefcase builds it, using the same `windows_signing_tool`/`signing_certificate` config as EXE
- Add signature verification to `test_example_signing`

### Checklist - did you ...

### DIFF
--- a/constructor/briefcase.py
+++ b/constructor/briefcase.py
@@ -22,7 +22,7 @@ else:
 
 from . import preconda
 from .jinja import render_template
-from .signing import create_signing_tool
+from .signing import create_windows_signing_tool
 from .utils import (
     DEFAULT_REVERSE_DOMAIN_ID,
     bat_echo_esc,
@@ -572,7 +572,7 @@ def create(info, verbose=False):
     # via the `conda constructor uninstall` command.
     info["uninstall_with_conda_exe"] = True
 
-    signing_tool = create_signing_tool(info)
+    signing_tool = create_windows_signing_tool(info)
 
     payload = Payload(info)
     try:
@@ -590,13 +590,13 @@ def create(info, verbose=False):
         if len(msi_paths) != 1:
             raise RuntimeError(f"Found {len(msi_paths)} MSI files in {dist_dir}, expected 1.")
 
+        if signing_tool:
+            signing_tool.sign(msi_paths[0])
+            signing_tool.verify_signature(msi_paths[0])
+
         outpath = Path(info["_outpath"])
         outpath.unlink(missing_ok=True)
         shutil.move(msi_paths[0], outpath)
-
-        if signing_tool:
-            signing_tool.sign(outpath)
-            signing_tool.verify_signature(outpath)
     finally:
         if not info.get("_debug"):
             payload.remove()
--- a/constructor/briefcase.py
+++ b/constructor/briefcase.py
@@ -22,6 +22,7 @@ else:
 
 from . import preconda
 from .jinja import render_template
+from .signing import create_signing_tool
 from .utils import (
     DEFAULT_REVERSE_DOMAIN_ID,
     bat_echo_esc,
@@ -571,6 +572,8 @@ def create(info, verbose=False):
     # via the `conda constructor uninstall` command.
     info["uninstall_with_conda_exe"] = True
 
+    signing_tool = create_signing_tool(info)
+
     payload = Payload(info)
     try:
         payload.prepare()
@@ -590,6 +593,10 @@ def create(info, verbose=False):
         outpath = Path(info["_outpath"])
         outpath.unlink(missing_ok=True)
         shutil.move(msi_paths[0], outpath)
+
+        if signing_tool:
+            signing_tool.sign(outpath)
+            signing_tool.verify_signature(outpath)
     finally:
         if not info.get("_debug"):
             payload.remove()
--- a/constructor/signing.py
+++ b/constructor/signing.py
@@ -67,6 +67,10 @@ class SigningTool:
         """Verify the signed installer."""
         raise NotImplementedError("Signature verification not implemented for base class.")
 
+    def sign(self, file_path: str | Path):
+        """Sign the specified file."""
+        raise NotImplementedError("Signing not implemented for base class.")
+
 
 class WindowsSignTool(SigningTool):
     def __init__(self, certificate_file=None):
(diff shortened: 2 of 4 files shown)


constructor/signing.py
```diff
@@ -75,21 +79,49 @@ def __init__(self, certificate_file=None):
             certificate_file=certificate_file,
         )
 
+    def _get_signing_params(self):
```
lrandersson, line 82: About the changes in this file to simplify review: The new code is based on existing patterns, `_get_signing_params()` extracts the same environment variables already used in `get_signing_command()`, and `sign()` builds the same command structure but as a list for direct execution. The `create_signing_tool()` function is essentially the same logic that was already in `winexe.py`, just moved to reduce duplicated code, but reuses existing logic to enable MSI signing.

tests/test_examples.py
```diff
@@ -1046,6 +1064,7 @@ def test_example_signing(tmp_path, request):
         CONSTRUCTOR_SIGNING_CERTIFICATE=str(cert_path),
         CONSTRUCTOR_PFX_CERTIFICATE_PASSWORD=cert_pwd,
     ):
+        _verify_windows_signature(installer)
```
lrandersson, line 1067: I added this because I thought it was odd that this test was even passing before signing of installers was supported for MSI Installers. This example was already running for EXE and MSI installers before this PR, and passed for MSI due to lack of actual verification during the testing.

constructor/briefcase.py
```diff
@@ -591,6 +594,10 @@ def create(info, verbose=False):
         outpath = Path(info["_outpath"])
         outpath.unlink(missing_ok=True)
         shutil.move(msi_paths[0], outpath)
+
+        if signing_tool:
```
marcoesters: Should we sign before moving the installer out of the staging directory?

constructor/signing.py
```diff
@@ -75,21 +79,49 @@ def __init__(self, certificate_file=None):
...
+            f"/tr {win_str_esc(params['timestamp_server'])} /td {params['timestamp_digest']} "
+            f"/fd {params['file_digest']}"
         )
-        if "CONSTRUCTOR_PFX_CERTIFICATE_PASSWORD" in os.environ:
+        if params["password"]:
             # signtool can get the password from the env var on its own
             command += ' /p "%CONSTRUCTOR_PFX_CERTIFICATE_PASSWORD%"'
         return command
 
+    def sign(self, file_path: str | Path):
```
marcoesters, line 105: There is still some duplication here since `get_signing_command` is just the `" ".join()` on the list you're building inside `sign()`

constructor/signing.py
```diff
@@ -120,6 +152,75 @@ class AzureSignTool(SigningTool):
...
+            ),
+            "timestamp_digest": os.environ.get("AZURE_SIGNTOOL_TIMESTAMP_DIGEST", "sha256"),
+            "file_digest": os.environ.get("AZURE_SIGNTOOL_FILE_DIGEST", "sha256"),
+            "access_token": os.environ.get("AZURE_SIGNTOOL_KEY_VAULT_ACCESSTOKEN"),
+            "secret": os.environ.get("AZURE_SIGNTOOL_KEY_VAULT_SECRET"),
+            "client_id": os.environ.get("AZURE_SIGNTOOL_KEY_VAULT_CLIENT_ID"),
+            "tenant_id": os.environ.get("AZURE_SIGNTOOL_KEY_VAULT_TENANT_ID"),
+        }
+
+    def sign(self, file_path: str | Path):
```
marcoesters, line 177: Same here: `get_signing_command` and what you need for `sign` are just two different representations of the same data.

constructor/signing.py
```diff
@@ -261,3 +362,23 @@ def sign_bundle(
         else:
             command = self.get_signing_command(bundle, entitlements=entitlements)
             explained_check_call(command)
+
+
+def create_signing_tool(info: dict) -> SigningTool | None:
+    """Create a signing tool based on construct.yaml configuration.
```
marcoesters:
```suggestion
    """Create a signing tool based on construct.yaml.
```

constructor/signing.py
```diff
@@ -261,3 +362,23 @@ def sign_bundle(
...
+    """
+    signing_tool_name = info.get("windows_signing_tool")
+    if not signing_tool_name:
+        return None
+
+    if signing_tool_name == "signtool":
+        signing_tool = WindowsSignTool(certificate_file=info.get("signing_certificate"))
+    elif signing_tool_name == "azuresigntool":
+        signing_tool = AzureSignTool()
+    else:
```
marcoesters, line 380: This function seems to only cover Windows, but this file also contains macOS signing tools.

constructor/signing.py
lrandersson: Yeah this is with some code duplication but its kept intentionally because I tried to keep it simple, here `get_signing_command` and `sign` have very different requirements.
To unify them, we'd need some kind of helper function that returns the command as a list of tuples like (`value`, `needs_escaping`, `is_env_reference`). Then `sign()` would extract just the values, while `get_signing_command()` would iterate through, apply `win_str_esc()` where `needs_escaping=True`, and replace actual values with `%VAR%` syntax where `is_env_reference=True`. For context:
- paths: `sign()` uses actual value, `get_signing_command()` needs `win_str_esc(value)`
- password: sign() uses actual value, `get_signing_command()` emits `"%ENV_VAR%"` literal for NSIS to expand during runtime.
Most elements just need optional escaping, but password is fundamentally different.

I'm happy to do this refactoring if we are OK with the added complexity.

constructor/briefcase.py
lrandersson: Good point I like that idea, 72d0414

constructor/signing.py
lrandersson: 72d0414

constructor/signing.py
lrandersson: I renamed it to make it more clear 72d0414

constructor/signing.py
marcoesters: In that case, I think keeping them separate is good.

marcoesters (approved):

lrandersson:
@marcoesters I just added two more commits
1. https://github.com/conda/constructor/pull/1224/changes/1b2d01c741a84fc1e549ba686a7580559b0acb9e - which wasn't enough
2. https://github.com/conda/constructor/pull/1224/changes/e7771ba4e6ac079745dd9a0f275f718b9ffdaeed
The reason for this is that the `macos-latest, Python 3.12` was failing with a bizarre error and it seems that the `latest` runner system Python has been updated to `Python 3.14`, because the command `pip install -e` was failing. I don't know why (is it intentional) that we use system `pip`? The `pip` command was finding `/Library/Frameworks/Python.framework/Versions/3.14/bin/pip` first in `PATH`, even though the `conda` environment is activated.
With the above mentioned reason I explicitly now invoke `pip` via `python` (that hopefully is the `python` from the conda environment), then the second commit I guess we need ever since the update to `conda-libmamba-solver` with sharded repodata enabled as default (since we saw that `pip` was no longer included as a dependency).
The jobs are still running with these changes so there might be additional changes needed to make it work.

marcoesters:
> 2\. I don't know why (is it intentional) that we use system `pip`?

This is strange. The environment should have been activated (and it is, or `conda install` wouldn't have worked). Maybe it's the way `setup-miniconda` is used? There has been a major release recently.

lrandersson:
> This is strange. The environment should have been activated (and it is, or `conda install` wouldn't have worked). Maybe it's the way `setup-miniconda` is used? There has been a major release recently.

@marcoesters Im not sure, this branch is still pinned to the previous version `conda-incubator/setup-miniconda@fc2d68f6413eb2d87b895e92f8584b5b94a10167 # v3.3.0`. It also only happened on the macos latest runner which is odd so I wonder if its just due to updates in the environment upstream.

lrandersson:
@marcoesters this has also been rebased and I removed the python/pip specific commits after the rebase so it also needs a new approval.

marcoesters (approved):

Issue #1170, opened 2026-02-17
EXE: Implement path length checks based on actual extracted path lengths


marcoesters:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

Current path length checks on Windows have a hard-coded value for the path length. That value is arbitrary and often too short to install into `%LOCALAPPDATA%`. The installers should instead use the correct path length limit by determining how long the longest path will be if long paths are not enabled.

### Why is this needed?

When installers fail because the paths of extracted files are too long, the error does not provide any actionable feedback. Windows refuses to write the file, but the extraction process continues to attempt the writing process, resulting in a `FileNotFoundError`. However, in multiprocessing, this error is not visible and will instead just look like an `InvalidArchiveError`.

This opaque error is confusing to users and difficult to debug (see https://github.com/conda/constructor/issues/1022 and https://github.com/conda/conda/issues/1564). The installer should instead use the correct value and provide proper messaging so that users know how to fix the error.

### What should happen?

`constructor` should detect the maximum path length of the installed files and use that to determine the maximum path length for `$INSTDIR`. The error message should inform users of their options (shorten path or enable long paths).

Issue #1022, opened 2025-07-28
Failed to extract packages error during installation


Tom-van-Woudenberg:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

After https://github.com/conda/constructor/pull/991 was merged, @moorepants began building our Windows installers from the development version of constructor. The first builds worked and we confirmed that the permissions issue was fixed. This first release was built with the development version just after #991 was merged:

https://gitlab.ewi.tudelft.nl/bhmgerritsen/anaconda-dee-config/-/releases/v2025.02 (exe download)

We have successfully installed this version on Windows many times. I recently started testing new installers we are building and reported that the Windows installer no longer works, showing this:

> I get an error "Failed to extract packages", during installation (after uninstall TU Delft conda, restart, reinstall): 

![](https://gitlab.ewi.tudelft.nl/-/project/6594/uploads/d6298186b1d62cb2f80746a3018a3d0f/image.png)

@moorepants has made some changes to the installer build, namely removing the pytorch channel in the construct.yaml file and some minor version bumps. You can see the changes here: 

https://gitlab.ewi.tudelft.nl/bhmgerritsen/anaconda-dee-config/-/compare/v2025.02...master?from_project_id=6594

@moorepants then wondered if the commits to constructor since #991 may have caused the installation failure, so he built an installer using constructor @ db1ea11ef7d2114e083993ad4235ab939fc69b20 (#991) and get this result:

https://github.com/moorepants/anaconda-dee-config/actions/runs/16362487740 (download the exe installer artifact)

and that installed successfully on Windows on his machine. However, it didn't work for me, leading to the same issue

To summarize, it seems like commits post #991 may be causing this failure to extract packages error on some machines. An example exe that should demonstrate this failure is:

https://github.com/moorepants/anaconda-dee-config/actions/runs/16340399472

Tom-van-Woudenberg:
@marcoesters, thank you for your reply!

@moorepants can confirm better which installer (from whih commit) worked.

`.\_conda.exe info` gives:

```
C:\Users\tomvanwoudenbe\AppData\Local\tudelft-conda>.\_conda.exe info

     active environment : None
       user config file : C:\Users\tomvanwoudenbe\.condarc
 populated config files : C:\Users\TOMVAN~1\AppData\Local\Temp\_MEI78602\.condarc
          conda version : 24.11.0
    conda-build version : not installed
         python version : 3.12.8.final.0
[...]

PR #1228, opened 2026-04-28
Improved error handling of long paths for EXE installers

lrandersson:
### Description
This pr covers https://github.com/conda/constructor/issues/1170 which adds a warning for Windows EXE installers.

When users select an installation path that, combined with the longest file path in the packages, may exceed Windows' 260-character limit, the installer now:
- Shows a warning dialog (Yes/No to continue) in interactive mode
- Logs `::warning::` in silent mode
- Appends a hint to extraction errors if path length may be the cause

The warning only appears when Long Paths is disabled in Windows. Computed at build time by tracking the longest relative path across all packages in `check_duplicates_files()`.

### Checklist - did you ...

### DIFF
--- a/constructor/_schema.py
+++ b/constructor/_schema.py
@@ -721,11 +721,11 @@ class ConstructorConfiguration(BaseModel):
 
     Only applies if `register_python` is true.
     """
-    check_path_length: bool = False
+    check_path_length: bool = Field(False, deprecated=True)
     """
-    Check the length of the path where the distribution is installed to ensure nodejs
-    can be installed.  Raise a message to request shorter paths (less than 46 character)
-    or enable long paths on windows > 10 (require admin right). Default is False. (Windows only).
+    _Deprecated_. Path length validation is now always performed using the computed
+    maximum relative path length from the package contents. This option will be
+    removed in a future version. (Windows only).
     """
     check_path_spaces: bool = True
     """
--- a/constructor/fcp.py
+++ b/constructor/fcp.py
@@ -7,17 +7,18 @@
 fcp (fetch conda packages) module
 """
 
+from __future__ import annotations
+
 import logging
 import os
 import shutil
 import sys
 import tempfile
 from collections import defaultdict
-from collections.abc import Iterable
 from itertools import groupby
 from os.path import abspath, expanduser, isdir, join
 from subprocess import check_call
-from typing import TYPE_CHECKING, Literal
+from typing import TYPE_CHECKING
 
 from constructor.utils import filename_dist
 
@@ -39,6 +40,9 @@ from .conda_interface import (
 )
 
 if TYPE_CHECKING:
+    from collections.abc import Iterable
+    from typing import Literal
+
     from .conda_interface import PackageCacheRecord
 
 logger = logging.getLogger(__name__)
@@ -146,10 +150,10 @@ def _fetch(download_dir, precs):
 
 
 def check_duplicates_files(
-    pc_recs: Iterable["PackageCacheRecord"],
+    pc_recs: Iterable[PackageCacheRecord],
     platform: str,
     duplicate_files: Literal["error", "warn", "skip"] = "error",
-    env_prefixes: dict["PackageCacheRecord", str] | None = None,
+    env_prefixes: dict[PackageCacheRecord, str] | None = None,
 ) -> tuple[int, int, int]:
     """
     Check for duplicate files across packages and compute size/path metrics.
@@ -167,6 +171,7 @@ def check_duplicates_files(
             Used to account for extra_envs paths which are installed under
             "envs/<name>/" rather than the base install directory. Records not
             in this dict are assumed to be in the base environment (no prefix).
+            The prefix string should include the trailing slash (e.g., "envs/myenv/").
 
     Returns:
         Tuple of (approx_tarball_size, approx_extracted_size, max_relative_path_length)
--- a/constructor/fcp.py
+++ b/constructor/fcp.py
@@ -153,6 +153,7 @@ def check_duplicates_files(pc_recs, platform, duplicate_files="error"):
     # Keep a min, 50MB buffer size
     total_tarball_size = 52428800
     total_extracted_pkgs_size = 52428800
+    max_relative_path_length = 0
 
     for pc_rec in pc_recs:
         fn = pc_rec.fn
@@ -163,6 +164,7 @@ def check_duplicates_files(pc_recs, platform, duplicate_files="error"):
         paths_data = read_paths_json(extracted_package_dir).paths
         for path_data in paths_data:
             short_path = path_data.path
+            max_relative_path_length = max(max_relative_path_length, len(short_path))
             try:
                 size = path_data.size_in_bytes or getsize(join(extracted_package_dir, short_path))
             except AttributeError:
@@ -176,7 +178,7 @@ def check_duplicates_files(pc_recs, platform, duplicate_files="error"):
             map_members_icase[short_path_lower]["fns"].add(fn)
 
     if duplicate_files == "skip":
-        return total_tarball_size, total_extracted_pkgs_size
+        return total_tarball_size, total_extracted_pkgs_size, max_relative_path_length
 
     logger.info("Checking for duplicate files ...")
     for member in map_members_scase:
@@ -201,7 +203,7 @@ def check_duplicates_files(pc_recs, platform, duplicate_files="error"):
             else:
                 sys.exit(f"Error: {msg_str}")
 
-    return total_tarball_size, total_extracted_pkgs_size
+    return total_tarball_size, total_extracted_pkgs_size, max_relative_path_length
 
 
 def _precs_from_environment(environment, input_dir):
@@ -443,7 +445,7 @@ def _main(
             input_dir=input_dir,
         )
     if dry_run:
-        return None, None, None, None, None, None, None, None
+        return None, None, None, None, None, None, None, None, None
     pc_recs, _urls, dists, has_conda = _fetch_precs(
         precs, download_dir, transmute_file_type=transmute_file_type
     )
@@ -463,8 +465,8 @@ def _main(
         duplicate_files = "skip"
 
     all_pc_recs = list({rec: None for rec in all_pc_recs})  # deduplicate
-    approx_tarballs_size, approx_pkgs_size = check_duplicates_files(
-        pc_recs, platform, duplicate_files=duplicate_files
+    approx_tarballs_size, approx_pkgs_size, max_relative_path_length = check_duplicates_files(
+        all_pc_recs, platform, duplicate_files=duplicate_files
     )
 
     return (
@@ -476,6 +478,7 @@ def _main(
         approx_pkgs_size,
         has_conda,
         extra_envs_data,
+        max_relative_path_length,
     )
 
 
@@ -529,6 +532,7 @@ def main(info, verbose=True, dry_run=False, conda_exe="conda.exe"):
             approx_pkgs_size,
             has_conda,
             extra_envs_info,
+            max_relative_path_length,
         ) = _main(
             name,
             version,
@@ -561,3 +565,4 @@ def main(info, verbose=True, dry_run=False, conda_exe="conda.exe"):
     info["_has_conda"] = has_conda
     # contains {env_name: [_dists, _urls, _records]} for each extra environment
     info["_extra_envs_info"] = extra_envs_info
+    info["_max_relative_path_length"] = max_relative_path_length
--- a/constructor/fcp.py
+++ b/constructor/fcp.py
@@ -13,10 +13,11 @@ import shutil
 import sys
 import tempfile
 from collections import defaultdict
+from collections.abc import Iterable
 from itertools import groupby
 from os.path import abspath, expanduser, isdir, join
 from subprocess import check_call
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Literal
 
 from constructor.utils import filename_dist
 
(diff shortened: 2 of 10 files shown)


constructor/fcp.py
```diff
@@ -463,8 +465,8 @@ def _main(
         duplicate_files = "skip"
 
     all_pc_recs = list({rec: None for rec in all_pc_recs})  # deduplicate
-    approx_tarballs_size, approx_pkgs_size = check_duplicates_files(
-        pc_recs, platform, duplicate_files=duplicate_files
+    approx_tarballs_size, approx_pkgs_size, max_relative_path_length = check_duplicates_files(
+        all_pc_recs, platform, duplicate_files=duplicate_files
```
lrandersson: This change is done to account for `extra_envs` otherwise the error check is only done for the `base` environment. I've also verified this won't cause unintended duplicate file warnings/errors:
1. When `extra_envs` exists, `duplicate_files` is explicitly set to "skip" (line 465), so `check_duplicates_files()` skips all duplicate checking logic entirely and only computes sizes (and now also max path length).
2. When there are no `extra_envs`, `all_pc_recs == pc_recs` (they're the same list), so behavior is unchanged.

constructor/nsis/main.nsi.tmpl
```diff
@@ -1586,9 +1606,18 @@ Section "Install"
     File {{ dist }}
 {%- endfor %}
 
+    # Set hint for extraction error if path may be too long
```
lrandersson: This is not needed for the actual underlying issue but I thought it would be more helpful than "Failed to extract packages", let me know what you think, I'm open to remove it completely.

tests/test_fcp.py
```diff
@@ -1,10 +1,11 @@
 from __future__ import annotations
 
 from contextlib import nullcontext
+from unittest.mock import patch
```
lrandersson: This is only using `unittest` to use the mock module which is part of the standard library. It's also possible to do this via `pyest-mock` but requires adding a new separate plugin. Let me know what you think.

The problem this solution is avoiding is that we need to know the exact `paths.json` format that `conda` expects, and create objects that satisfy conda's `PackageCacheRecord` interface. This couples the test to conda's internal implementation details which I don't like.

constructor/nsis/main.nsi.tmpl
```diff
@@ -1115,6 +1117,24 @@ Function OnDirectoryLeave
         ${EndIf}
     ${EndIf}
 
+    # MAX_PATH check - runs independently of check_path_length
```
marcoesters: I think this should make `check_path_length` obsolete. Now that the check is reliable, we should always check because we *know* the installation will fail.

constructor/nsis/main.nsi.tmpl
```diff
@@ -1115,6 +1117,24 @@ Function OnDirectoryLeave
...
+    ${If} $LongPathsEnabled == "0"
+        IntOp $0 $InstDirLen + ${MAX_RELATIVE_PATH_LENGTH}
+        ${If} $0 > 260
+            ${If} ${Silent}
+                ${Print} "::warning:: The installation path may be too long ($0 chars, limit 260). Some files could fail to extract."
```
marcoesters: Since they *will* (not could) fail to extract, the installation should not continue.

I also think that `$0` is not helpful for the user - `$InstDirLen` (how long their path is) and `260 - ${MAX_RELATIVE_PATH_LENGTH}` (how long the installation directory is allowed to be) are though.

constructor/nsis/main.nsi.tmpl
marcoesters: > This is not needed for the actual underlying issue but I thought it would be more helpful than "Failed to extract packages", let me know what you think, I'm open to remove it completely.

If we let the user continue, maybe. I would not use a global variable for this though, but put the `push` instruction into the if-else construction.

constructor/nsis/main.nsi.tmpl
```diff
@@ -1586,9 +1606,18 @@ Section "Install"
     File {{ dist }}
 {%- endfor %}
 
+    # Set hint for extraction error if path may be too long
+    IntOp $0 $InstDirLen + ${MAX_RELATIVE_PATH_LENGTH}
```
marcoesters: Does this take the (lack of?) trailing backslash for `$InstDir` into account?

constructor/fcp.py
marcoesters: > This change is done to account for `extra_envs` otherwise the error check is only done for the `base` environment. I've also verified this won't cause unintended duplicate file warnings/errors:
> 
>     1. When `extra_envs` exists, `duplicate_files` is explicitly set to "skip" (line 465), so `check_duplicates_files()` skips all duplicate checking logic entirely and only computes sizes (and now also max path length).
> 
>     2. When there are no `extra_envs`, `all_pc_recs == pc_recs` (they're the same list), so behavior is unchanged.

This should be a comment in the source code.

Does the check take into account that files inside `extra_envs` will have a longer base path though? For the `base` environment, the base path is `$INSTDIR`, for extra environments, it's `$INSTDIR\envs\<environment name>`.

tests/test_fcp.py
marcoesters: > This is only using `unittest` to use the mock module which is part of the standard library. It's also possible to do this via `pyest-mock` but requires adding a new separate plugin. Let me know what you think.
> 
> The problem this solution is avoiding is that we need to know the exact `paths.json` format that `conda` expects, and create objects that satisfy conda's `PackageCacheRecord` interface. This couples the test to conda's internal implementation details which I don't like.

`pytest-mock` is a common enough plug-in. I would rather stick to one framework.

constructor/nsis/main.nsi.tmpl
lrandersson: @marcoesters I agree, I propose that I deprecate this option but potentially in a follow-up PR, what do you think?

constructor/nsis/main.nsi.tmpl
lrandersson: I think that's a fair assessment, would this be considered a breaking change though? I agree that an error here makes a lot of sense but I am just worried that making this to an error to begin with would be too intrusive (I dont know if people run these installers with too long paths (hopefully not)). 
Perhaps we could make it into a warning and mention this will become an error in a future release?

lrandersson:
@marcoesters 
I've had made some changes in commit 1552b5d to address some of the review comments:
1. `check_duplicates_files()` - Added `env_prefixes` parameter
- You're right we missed accounting for the additional prefix for `extra_envs`. To fix this we needed to create a mapping because the list of files that were iterating over are not tracking the respective environment they are part of. The function now accepts an optional `env_prefixes` dict that maps package records to their environment prefix (e.g., "`envs/myenv/`"). When computing max path length, the prefix length is added to each file's path.

2. `_main()` - Build `env_prefixes` mapping (needed for above)
- While iterating `extra_envs`, we now build the `env_prefixes` dict mapping each package record to its "`envs/<name>/`" prefix. If a package appears in multiple `extra_envs`, we keep the longest prefix. This mapping is passed to `check_duplicates_files()`.

3. NSIS template - Path separator fix
- Added `+ 1` to both `MAX_PATH` calculations to account for the backslash between install directory and relative path (e.g., `C:\App + \ + lib\foo.py`).

4. Tests - Migrated to `pytest-mock`
- Switched from `unittest.mock.patch` decorator to `pytest-mock` mocker fixture. Added new test `test_check_duplicates_files_with_env_prefixes` to verify the prefix calculation works correctly.

constructor/nsis/main.nsi.tmpl
marcoesters: I think we should do it here - a simple comment in the schema and a warning when parsing the option should suffice.

constructor/nsis/main.nsi.tmpl
marcoesters: The installation will fail anyway because Windows will refuse to extract files that will exceed the maximum path length. The error they see is not very helpful either (see, for example, #1022). With this change, the installation fails earlier with a useful error message that gives the user feedback on how to fix the installation.

Unless you see a circumstance where this error would prevent a successful installation from running, I don't see where this is a breaking change.

lrandersson: Hmm, I agree with what you're saying.
I have changed the warning into an error in this commit: 675b9bf.
The hint has also been removed now since it doesnt serve any purpose with the warning now being an error. Note also the error message has been updated since the last review.

constructor/nsis/main.nsi.tmpl
lrandersson: Added here 675b9bf

examples/miniforge/construct.yaml
```diff
@@ -11,7 +11,6 @@ channels:
 
 write_condarc: True
 keep_pkgs: True
-transmute_file_type: .conda
```
lrandersson, line 14: I believe something changed upstream because now the solver occasionally picks `pybind11-abi` version 4 (a dependency of `mamba 1.5.12`)??? This is odd because in https://github.com/conda/constructor/pull/1224 it picked `pybind11-abi 11`, I wonder if something changed recently upstream. Version 4 only exists as `.tar.bz2` on `conda-forge` - there's no `.conda` artifact. As I understand when `transmute_file_type: .conda` is set, constructor converts the downloaded `.tar.bz2` to `.conda` locally. However, when the installer runs, `libmamba` tries to verify the package URL against the channel and fails with `404` since `conda-forge` has no `.conda` entry for this package in the repodata.

Removing `transmute_file_type` avoids this issue. The transmute feature is still tested by other examples (e.g., customized).

lrandersson:
@marcoesters I rebased the branch with the latest changes from main so this PR needs a new review

constructor/nsis/main.nsi.tmpl
```diff
@@ -1115,6 +1116,26 @@ Function OnDirectoryLeave
         ${EndIf}
     ${EndIf}
 
+    # MAX_PATH check - installation will fail if path exceeds 260 characters
```
marcoesters, line 1073: The path-length handling logic before this section needs to be trimmed. We can get rid of all the `$CheckPathLength` logic.

constructor/_schema.py
```diff
@@ -721,11 +721,11 @@ class ConstructorConfiguration(BaseModel):
 
     Only applies if `register_python` is true.
     """
-    check_path_length: bool = False
+    check_path_length: bool = Field(False, deprecated=True)
```
marcoesters, line 724: If this is deprecated, we should warn about it. In fact, I would remove that template parameter entirely form the NSIS template.

constructor/fcp.py
```diff
@@ -144,15 +145,43 @@ def _fetch(download_dir, precs):
     return list(dict.fromkeys(PrefixGraph(pc.iter_records()).graph))
 
 
-def check_duplicates_files(pc_recs, platform, duplicate_files="error"):
+def check_duplicates_files(
+    pc_recs: Iterable["PackageCacheRecord"],
```
marcoesters: If you add `from __future__ import annotations`, you don't need the strings representation. You can then place `Literal` and `Iterable` into the type checking block.

constructor/fcp.py
```diff
@@ -161,8 +190,12 @@ def check_duplicates_files(pc_recs, platform, duplicate_files="error"):
         total_tarball_size += int(pc_rec.get("size", 0))
 
         paths_data = read_paths_json(extracted_package_dir).paths
+        env_prefix_len = len(env_prefixes.get(pc_rec, ""))
```
marcoesters, line 201: Does the `env_prefix` contain the trailing (back)slash? Might be worth documenting as a comment.

lrandersson:
@marcoesters see commit [28f6004](https://github.com/conda/constructor/pull/1228/changes/28f600401d35a439012da18fa01fb3ca307b012e) for the latest changes
1. "The path-length handling logic before this section needs to be trimmed. We can get rid of all the $CheckPathLength logic."
  - Done. Removed all `$CheckPathLength` and related code.
2. "If this is deprecated, we should warn about it. In fact, I would remove that template parameter entirely from the NSIS template."
  - Done. If I understand this correctly, the deprecation warning is already handled via the JSON schema validator (`deprecated_validator` in `construct.py`). Removed the `check_path_length` template variable from winexe.py so it's no longer passed to the NSIS template.
3. "If you add from __future__ import annotations, you don't need the strings representation. You can then place Literal and Iterable into the type checking block."
  - Done. Added `from __future__ import annotations`, moved `Iterable` and `Literal` to the `TYPE_CHECKING` section, and removed the string quotes from type annotations.
4. "Does the env_prefix contain the trailing (back)slash? Might be worth documenting as a comment."
  - Done. Added to the docstring: "The prefix string should include the trailing slash (e.g., "envs/myenv/")."

constructor/nsis/main.nsi.tmpl
```diff
@@ -1083,34 +1070,22 @@ Function OnDirectoryLeave
...
-            ; If we don't have admin right, we suggest a shorter path or suggest to run with admin right
+    # MAX_PATH check - installation will fail if path exceeds 260 characters
+    ${If} $LongPathsEnabled == "0"
+        IntOp $0 $InstDirLen + ${MAX_RELATIVE_PATH_LENGTH}
+        IntOp $0 $0 + 1  # Account for path separator between install dir and relative path
+        ${If} $0 > 260
+            IntOp $1 260 - ${MAX_RELATIVE_PATH_LENGTH}
+            IntOp $1 $1 - 1  # Account for path separator
+            ${If} ${Silent}
+                ${Print} "::error:: The installation path is too long. Your path is $InstDirLen characters, but must be at most $1 characters."
```
marcoesters: Installers automatically enable long paths if they have administrative privileges. We should keep it that way.

constructor/fcp.py
```diff
@@ -144,15 +149,44 @@ def _fetch(download_dir, precs):
...
+
+    Args:
+        pc_recs: Package cache records to check.
+        platform: Target platform string (e.g., "win-64", "linux-64").
+        duplicate_files: How to handle duplicates - "error", "warn", or "skip".
+        env_prefixes: Optional dict mapping PackageCacheRecord -> path prefix string.
+            Used to account for extra_envs paths which are installed under
+            "envs/<name>/" rather than the base install directory. Records not
+            in this dict are assumed to be in the base environment (no prefix).
+            The prefix string should include the trailing slash (e.g., "envs/myenv/").
```
marcoesters: Is it "should" or "does"? If it's "should", we should make sure that it does and adjust the path length calculation accordingly.

tests/test_fcp.py
```diff
@@ -64,3 +64,101 @@ def test_exclude_packages(values: tuple[..., GenericObject], expected_value):
     with context:
         packages = exclude_packages(*values)
         assert packages == expected_value
+
+
+class MockPathData:
+    """Represents a file path entry."""
+
+    def __init__(self, path):
```
marcoesters: Let's add typing to new code where we can.

marcoesters (commented):
This looks very good - we're almost there, I think.

constructor/nsis/main.nsi.tmpl
lrandersson: Good point, see https://github.com/conda/constructor/pull/1228/changes/5add07a3ab0fffc125e70324ca41d879d5ce732d

constructor/fcp.py
lrandersson: https://github.com/conda/constructor/pull/1228/changes/5add07a3ab0fffc125e70324ca41d879d5ce732d, also added trailing slash validation and updated the docstring

tests/test_fcp.py
lrandersson: Done https://github.com/conda/constructor/pull/1228/changes/5add07a3ab0fffc125e70324ca41d879d5ce732d

constructor/fcp.py
```diff
@@ -171,14 +171,17 @@ def check_duplicates_files(
...
 
     Returns:
         Tuple of (approx_tarball_size, approx_extracted_size, max_relative_path_length)
     """
     assert duplicate_files in ("warn", "skip", "error")
     if env_prefixes is None:
         env_prefixes = {}
+    for prefix in env_prefixes.values():
+        if prefix and not prefix.endswith("/"):
+            raise ValueError(f"env_prefixes values must end with '/': got {prefix!r}")
```
marcoesters: Could we just normalize it instead of erroring out?

lrandersson: Good idea, that makes it a bit more graceful. See dec710e

marcoesters (approved):

PR #1235, opened 2026-05-14
MSI: Add support for installer branding

lrandersson:
### Description

Add branding image support (`welcome_image`, `header_image`, `icon_image`) to MSI installers.

MSI installers now generate branded dialog images and icons, matching the existing EXE behavior. Users can provide custom images or rely on auto-generation.

**Key changes:**

- **`imaging.py`**: Added MSI image dimensions (493x312 welcome, 493x58 header, 256x256 icon) and refactored `os` parameter to `installer_type` for clarity ("exe", "pkg", "msi")
- **`briefcase.py`**: Integrated `write_images()` to generate branding files and configured `pyproject.toml` with `installer_background`, `installer_banner`, and `icon` paths
- **`osxpkg.py`**: Updated to use new `installer_type="pkg"` parameter
- **`_schema.py`**: Added "(Not supported for MSI installers.)" to text branding options (`welcome_file`, `welcome_text`, `readme_file`, `readme_text`, `conclusion_file`, `conclusion_text`)
- **Cleanup**: Removed unused `if __name__ == "__main__"` block from `imaging.py` and deleted `examples/miniconda/` directory

**Note on image dimensions:** MSI welcome/header images have different aspect ratios than EXE (landscape vs portrait). Auto-resize will stretch images. I added a TODO because I will likely need to revisit this (hence also why it's opened as a Draft for now).

### Checklist - did you ...


constructor/imaging.py
```diff
@@ -99,19 +104,37 @@ def add_color_info(info):
         sys.exit("Error: color '%s' not defined" % color_name)
 
 
-def write_images(info, dir_path, os="windows"):
-    if os == "windows":
+def write_images(info, dir_path, installer_type="exe"):
+    if installer_type == "exe":
```
lrandersson, line 128: I also learnt today that `_schema.py` defines `class InstallerTypes(StrEnum)`, I think we should expose this class in `constructor` and update all the installer type checks to use this.

constructor/_schema.py
```diff
@@ -755,31 +755,33 @@ class ConstructorConfiguration(BaseModel):
     File can be plain text (.txt), rich text (.rtf) or HTML (.html). If
     both `welcome_file` and `welcome_text` are provided, `welcome_file` takes precedence.
 
-    If the installer is for Windows and the welcome file type is nsi,
+    If the installer is for Windows EXE and the welcome file type is nsi,
     it will use the nsi script to add in extra pages before the installer
-    begins the installation process.
+    begins the installation process. (Not supported for MSI installers.)
```
marcoesters: I don't think we need to add all the items that are not supported for MSI installers. You are being explicit with the supported formats already.

constructor/briefcase.py
```diff
@@ -36,6 +39,13 @@
 BRIEFCASE_DIR = Path(__file__).parent / "briefcase"
 EXTERNAL_PACKAGE_PATH = "external"
 
+# MSI Branding Limitations:
+# The following EXE branding options are not supported for MSI installers
+# because they require modifications to the WiX template in briefcase-windows-app-template:
+# - welcome_file / welcome_text (custom welcome page text)
+# - readme_file / readme_text (readme page)
```
marcoesters: I think the readme file is for PKG installers only anyway.

constructor/imaging.py
```diff
@@ -99,19 +104,37 @@ def add_color_info(info):
...
+    elif installer_type == "pkg":
         instructions = [
             ("welcome", welcome_size_osx, mk_welcome_image_osx, ".png"),
         ]
+    elif installer_type == "msi":
+        instructions = [
+            (
+                "welcome",
+                welcome_size_msi,
+                lambda info: mk_welcome_image(info, welcome_size_msi),
```
marcoesters: Should the EXE installers also use lambda functions so that we have more explicit names for variables?

lrandersson: The lambdas are only needed for MSI because it passes sizes different from default. EXE uses default parameters, so lambdas would be redundant. I could change it just for consistency but provides with no functional benefit, let me know what you think.

constructor/_schema.py
lrandersson: See a886add

constructor/briefcase.py
lrandersson: a886add

constructor/imaging.py
marcoesters: It's the other way round: the default parameters are the EXE parameters. Making it a lambda would make this more explicit, but that's a minor suggestion.

marcoesters:
@lrandersson, the schema output needs to be refreshed since you changed the descriptions.

lrandersson:
@marcoesters I made some more significant changes because I realized it had some unnecessary code since we can rely on some briefcase/wix default logic, see: 3af4b48
1. No longer always auto-generate images, only when user specifies any of the support branding options.
2. Update docs.
3. Properly scale and pad the welcome image, i.e. rescale to a 164-pixel wide image and pad with white to the right. If "A" is an image, it will be now transformed into "A | white padding".
4. Removed invalid test (since we only create images on demand). Update existing tests to account for this.
5. Remove added lambdas completely due to "reverting" some of the changes.

constructor/_schema.py
```diff
@@ -755,33 +757,31 @@ class ConstructorConfiguration(BaseModel):
     File can be plain text (.txt), rich text (.rtf) or HTML (.html). If
     both `welcome_file` and `welcome_text` are provided, `welcome_file` takes precedence.
 
-    If the installer is for Windows EXE and the welcome file type is nsi,
+    If the installer is for Windows and the welcome file type is nsi,
```
marcoesters: Just EXE, right?

lrandersson: Good catch, shouldn't have made it that implicit. See 97fbdf5.

marcoesters (approved):

PR #1241, opened 2026-05-21
Enable more tests for MSI installers

lrandersson:
### Checklist - did you ...

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -448,11 +448,18 @@ def _run_installer_msi(
     if not sys.platform.startswith("win"):
         raise ValueError("Can only run .msi installers on Windows")
 
-    # Currently we only have 1 test that specifies options, so this is a temporary "fix"
+    # Translate NSIS-style options to MSI properties and collect MSI properties
+    msi_properties = []
+    allusers = False
     if options:
-        allusers = "/InstallationType=AllUsers" in options
-    else:
-        allusers = False
+        for opt in options:
+            if opt == "/InstallationType=AllUsers":
+                allusers = True
+            elif opt == "/InstallationType=JustMe":
+                allusers = False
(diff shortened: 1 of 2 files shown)


tests/test_examples.py
```diff
@@ -448,11 +448,18 @@ def _run_installer_msi(
...
 
-    # Currently we only have 1 test that specifies options, so this is a temporary "fix"
+    # Translate NSIS-style options to MSI properties and collect MSI properties
+    msi_properties = []
+    allusers = False
     if options:
-        allusers = "/InstallationType=AllUsers" in options
-    else:
-        allusers = False
+        for opt in options:
```
marcoesters: You could remove a layer of nesting here:

```
options = options or []
for opt in options:
    ...
```

tests/test_examples.py
```diff
@@ -461,8 +468,9 @@ def _run_installer_msi(
         "ALLUSERS=1"
         if allusers
         else "MSIINSTALLPERUSER=1",  # For some reason tests fail on the CI system if "ALLUSERS=1"
-        "/qn",
     ]
+    cmd.extend(msi_properties)
+    cmd.append("/qn")
```
```suggestion
    *msi_properties,
    "/qn",
]
```

lrandersson: Nice catch, see aa50f90

tests/test_examples.py
lrandersson: Good catch, I changed it slightly further since there is no need for `None` as default. See aa50f90.

marcoesters: `None` is generally better than `[]` because it is not mutable.

lrandersson: See ca5cfa3

marcoesters (approved):

PR #1242, opened 2026-05-21
Specify revision for hook

lrandersson:
### Checklist - did you ...


marcoesters:
I really like this change, but I think we should do this for all hooks.

lrandersson:
> I really like this change, but I think we should do this for all hooks.

See a1c854b

marcoesters (approved):

PR #1247, opened 2026-05-26
 Use `$INSTDIR\tmp` as the temporary directory for EXE to avoid Windows Defender errors

### LINKED ISSUE #1243: Windows install fails due to WDAC
### Checklist

### What happened?

I ran into the error ::error:: Failed to extract packages during the installation. This happened using winget or the downloaded installer. Both were tryinig to install v26.1.1-3.

This error is caused by Windows Defender Application Control (WDAC) preventing the installer from unpacking an unsigned DLL (the python DLL) to %LOCALDATA%\Temp\.... The WDAC is managed by my organization.

I was able to work around the issue by stringing together a set of pwsh commands that makes use of winget. This works by temporarily changing the TEMP directory.

This is a copy of a bug that was reported here: https://github.com/conda-forge/miniforge/issues/863

### Additional Context

_No response_

marcoesters:
### Description

Use `$INSTDIR\tmp` as the temporary directory to avoid Windows Defender errors.

When `conda-standalone` is called, it extracts itself into a [temporary directory](https://pyinstaller.org/en/stable/usage.html#defining-the-extraction-location). On Windows, this is done using the [`GetTempPathW`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppathw) function, which looks for the `%TMP%` environment variable first, which defaults to `%LOCALAPPDATA%\Temp`.

Windows Defender may disallow using the user temporary directory for the expanded binary (see #1243), but does allow them to be extracted into different directories. A natural location for such as "local" temporary directory is the installation directory as is already done in the SH installer:  https://github.com/conda/constructor/blob/7cbad1abd101078059cd54db28a24fa5ba9520c5/constructor/header.sh

Closes #1243 

### Checklist - did you ...

### DIFF
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -1813,7 +1813,7 @@ Section "Uninstall"
     # being extracted into LOCALAPPDATA
     # See https://github.com/conda/constructor/issues/1243
     StrCpy $TmpDir "$INSTDIR\tmp"
-    System::Call 'kernel32::SetEnvironmentVariable(t,t)i("TMP", "${TmpDir}")'
+    System::Call 'kernel32::SetEnvironmentVariable(t,t)i("TMP", "$TmpDir")'
     CreateDirectory "$TmpDir"
 
     # Extra info for pre_uninstall scripts
--- a/constructor/nsis/main.nsi.tmpl
+++ b/constructor/nsis/main.nsi.tmpl
@@ -1812,7 +1812,7 @@ Section "Uninstall"
     # Windows Defender does not get triggered by conda-standalone
     # being extracted into LOCALAPPDATA
     # See https://github.com/conda/constructor/issues/1243
-    StrCpy TmpDir "$INSTDIR\tmp"
+    StrCpy $TmpDir "$INSTDIR\tmp"
     System::Call 'kernel32::SetEnvironmentVariable(t,t)i("TMP", "${TmpDir}")'
     CreateDirectory "$TmpDir"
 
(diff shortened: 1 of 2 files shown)


constructor/nsis/main.nsi.tmpl
```diff
@@ -1460,6 +1461,14 @@ Section "Install"
     # for users even during an all-users installation.
     !insertmacro setInstdirPermissions
 
+    # Create the temporary directory inside $INSTDIR so that
+    # Windows Defender does not get triggered by conda-standalone
+    # being extracted into LOCALAPPDATA
+    # See https://github.com/conda/constructor/issues/1243
+    StrCpy $TmpDir "$INSTDIR\tmp"
+    System::Call 'kernel32::SetEnvironmentVariable(t,t)i("TMP", "$TmpDir")'
```
lrandersson, line 1469: Just for consistency here we write `$TmpDir` but further down we write `${TmpDir}`.

lrandersson (dismissed):
I think this looks good. I had some "concerns" initially about user-supplied scripts and use of `%TMP%` but I think that won't be an issue.

constructor/nsis/main.nsi.tmpl
```diff
@@ -1796,6 +1808,14 @@ Section "Uninstall"
         goto loop_py
     endloop_py:
 
+    # Create the temporary directory inside $INSTDIR so that
+    # Windows Defender does not get triggered by conda-standalone
+    # being extracted into LOCALAPPDATA
+    # See https://github.com/conda/constructor/issues/1243
+    StrCpy $TmpDir "$INSTDIR\tmp"
+    System::Call 'kernel32::SetEnvironmentVariable(t,t)i("TMP", "${TmpDir}")'
```
```suggestion
    System::Call 'kernel32::SetEnvironmentVariable(t,t)i("TMP", "$TmpDir")'
```

lrandersson (approved):

PR #1248, opened 2026-05-27
MSI: Update documentation

lrandersson:
### Checklist - did you ...

### DIFF
--- a/docs/source/cli-options.md
+++ b/docs/source/cli-options.md
@@ -227,7 +227,7 @@ set of command-line options passed as public properties.
 | Property | Description | Default |
 |----------|-------------|---------|
 | `ALLUSERS` | `1` for all users, `0` (or empty) for current user | Per-user |
-| `TARGETDIR` | Installation directory | `%LOCALAPPDATA%\<name>` or `%PROGRAMFILES%\<name>` |
+| `TARGETDIR` | Installation directory | `%LOCALAPPDATA%\<name>` (per-user) or `%PROGRAMFILES%\<name>` (all users) |
 | `OPTION_REGISTER_PYTHON` | Register as system Python (`1` or `0`) | `0` |
 | `OPTION_INITIALIZE_CONDA` | Add to PATH (`1` or `0`) | `0` |
 | `OPTION_CLEAR_PACKAGE_CACHE` | Clear package cache after install (`1` or `0`) | `1` |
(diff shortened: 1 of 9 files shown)


docs/source/cli-options.md
```diff
@@ -200,6 +200,100 @@ Combine this with `CONDA_VERBOSITY=3` at install time for maximum details:
...
+| `/qb` | Basic UI (progress bar only) |
+| `/qr` | Reduced UI |
+| `/l*v <logfile>` | Verbose logging to file |
+
+**Installer properties (passed as `PROPERTY=value`):**
+
+| Property | Description | Default |
+|----------|-------------|---------|
+| `ALLUSERS` | `1` for all users, `0` (or empty) for current user | Per-user |
+| `TARGETDIR` | Installation directory | `%LOCALAPPDATA%\<name>` or `%PROGRAMFILES%\<name>` |
```
marcoesters: Should mention that the former is for current user and the latter for all users.

docs/source/cli-options.md
```diff
@@ -200,6 +200,100 @@ Combine this with `CONDA_VERBOSITY=3` at install time for maximum details:
...
+| `OPTION_ENABLE_SHORTCUTS` | Create shortcuts (`1` or `0`) | `1` |
+| `OPTION_PRE_INSTALL_SCRIPT` | Run pre-install script if configured (`1` or `0`) | `0` |
+| `OPTION_POST_INSTALL_SCRIPT` | Run post-install script if configured (`1` or `0`) | `0` |
+
+### Uninstallation
+
+MSI installers can be uninstalled via:
+
+1. **Add/Remove Programs**: Right-click and select "Uninstall"
+2. **Double-clicking the MSI file**: Shows a dialog with "Repair" and "Remove" options. **This is the only way to access uninstall options** like removing caches or config files.
```
marcoesters: I assume that limitation is due to the MSI format, not `briefcase`? Is there a way we could manipulate the registry entry somehow to make it work with Option 1?

It's also not quite true that this is the only way since 3. provides those options, too.

docs/source/howto.md
```diff
@@ -257,3 +259,44 @@ it to your `specs` definition and it will be available in your installations as
...
+## MSI Installers (Experimental)
+
+:::{warning}
+MSI installer support is experimental and may change in future releases.
+:::
+
+MSI installers provide an alternative to EXE installers on Windows, built using
+[Briefcase](https://beeware.org/project/projects/tools/briefcase/). They offer
+native Windows Installer integration, which is useful for enterprise deployment
+via Group Policy.
```
marcoesters: Has this been tested?

docs/source/cli-options.md
lrandersson: Fixed 13a5933

docs/source/cli-options.md
lrandersson: Good catch, I fixed it here 13a5933.

docs/source/howto.md
lrandersson: No it hasn't been tested, I removed it for now (13a5933) - it's better to not say anything than to sound fancy about possibly incorrect statements.

docs/source/cli-options.md
lrandersson: To answer the question about the registry manipulation: Yes it's possible, this is actually configured via the WiX template (right now hardcoded and not possible to configure from the installer builder POV).

I think the current setting is standard behavior and changing this could in theory break automated scripts and/or deployment tools that just expect it to work (to uninstall silently), but I would need to read up more on this in order to make an accurate statement.

marcoesters (approved):

PR #1250, opened 2026-05-27
 Serialize package records and versions as built-in types in info.json files

marcoesters:
### Description

The `info.json` file contains some objects that are not fully serialized. The different package records, for example, are serialized as `__repr__`, which makes them not parseable without using `eval`. They do support a `dump()` method, which allows serializing them into built-in types.

### Checklist - did you ...


tests/test_examples.py
```diff
@@ -1452,6 +1452,23 @@ def test_output_files(tmp_path):
         files_exist = [file for file in files_not_expected if (root_path / file).exists()]
         assert files_exist == []
 
+        # Test that info.json contains serialized objects
+        info_json = json.loads((root_path / "info.json").read_text())
+        assert isinstance(info_json.get("_conda_exe_version"), str)
+        _all_pkg_records = info_json.get("_all_pkg_records")
+        assert isinstance(_all_pkg_records, list), "Package record is not a list."
+        assert len(_all_pkg_records) > 0, "Package record is empty."
+        assert isinstance(_all_pkg_records[0], dict), "Package record no serialized."
```
lrandersson:
```suggestion
        assert isinstance(_all_pkg_records[0], dict), "Package record not serialized."
```

constructor/build_outputs.py
```diff
@@ -76,9 +77,17 @@ def dump_hash(info, algorithm=None):
 
 
 def dump_info(info):
+    def _serialize(obj):
+        if hasattr(obj, "dump"):
+            return obj.dump()
+        elif isinstance(obj, VersionOrder):
+            return obj.norm_version
+        else:
+            return repr(obj)
```
lrandersson, line 86: Is there any need to do any logging in case we end-up here with `repr` which can still generate non-parseable strings? I'm fine leaving it as is but just for consideration.

marcoesters: Since the majority of entries will use `repr`, I don't think logging would be particularly useful.

lrandersson (approved):

PR #1251, opened 2026-05-28
Add MSI experimental support

lrandersson:
### Description

This branch is a copy of the branch `briefcase-integration` containing the entire feature branch for MSI installers.

### Checklist - did you ...


news/1235-msi-branding
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* MSI: Add branding image support (`welcome_image`, `header_image`, `icon_image`). (#1235)
```
lrandersson: TODO

constructor/osx/run_installation.sh
```diff
@@ -115,10 +115,6 @@ done
 # Cleanup!
 find "$PREFIX/pkgs" -type d -empty -exec rmdir {} \; 2>/dev/null || :
 
-{%- for condarc in write_condarc %}
-{{ condarc }}
-{%- endfor %}
```
lrandersson, line 120: TODO; dont remember the origin of this change

lrandersson: Now I recall, this is done now by `preconda.py` (see `def write_condarc(info: dict, dst_dir: str)`)

lrandersson:
All commits have been reviewed on this branch **except**:
1. Enable more tests 18065e31842649bf20039a1fa96fefb95e2844f1
2. Update test to account for MSI layout 153e707604b1cd2e36a7e9323275aaebf4629c18 + 7a6ad19e34a51a069d5f73e1f7a33fa07570b1b3 (pre-commit fix)
3. Add news 50c11998722204e6e4eefd47baa0630922e2f4d0

pyproject.toml
```diff
@@ -16,7 +16,9 @@ dependencies = [
     "ruamel.yaml >=0.11.14,<0.19",
     "pillow >=3.1 ; platform_system=='Windows' or platform_system=='Darwin'",
     "jinja2",
-    "jsonschema >=4"
+    "jsonschema >=4",
+    "briefcase >=0.3.26 ; platform_system=='Windows'",
+    "tomli-w >=1.2.0 ; platform_system=='Windows'",
```
marcoesters, line 21: Should they become optional dependencies instead while support is still experimental?

recipe/meta.yaml
```diff
@@ -25,11 +25,13 @@ requirements:
...
     - python  # >=3.10
     - ruamel.yaml >=0.11.14,<0.19
-    - conda-standalone >=24.1.2
+    - conda-standalone >=24.11.0
     - jinja2
     - jsonschema >=4
     - pillow >=3.1     # [win or osx]
     - nsis >=3.08      # [win]
+    - briefcase >=0.3.26  # [win]
+    - tomli-w >=1.2.0  # [win]
```
marcoesters, line 34: Same here with optional dependencies: if they become optional dependencies, they need to go into `run_constrained`.

news/1251-msi-experimental-support
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Add experimental support for building Windows MSI installers (via BeeWare's Briefcase). (#1251)
```
marcoesters: I'm sure we can close at least one issue with this.

constructor/signing.py
```diff
@@ -75,21 +79,49 @@ def __init__(self, certificate_file=None):
...
+            str(self.certificate_file),
+            "/tr",
+            params["timestamp_server"],
+            "/td",
+            params["timestamp_digest"],
+            "/fd",
+            params["file_digest"],
+        ]
+        if params["password"]:
+            command.extend(["/p", params["password"]])
```
marcoesters: This looks like it would add the password of the certificate in plain text to the command. We should use the literal environment variable here like we did before.

constructor/signing.py
```diff
@@ -75,21 +79,49 @@ def __init__(self, certificate_file=None):
...
 
+    def _get_signing_params(self):
+        """Get signing parameters from environment."""
+        return {
+            "timestamp_server": os.environ.get(
+                "CONSTRUCTOR_SIGNTOOL_TIMESTAMP_SERVER_URL", "http://timestamp.sectigo.com"
+            ),
+            "timestamp_digest": os.environ.get("CONSTRUCTOR_SIGNTOOL_TIMESTAMP_DIGEST", "sha256"),
+            "file_digest": os.environ.get("CONSTRUCTOR_SIGNTOOL_FILE_DIGEST", "sha256"),
+            "password": os.environ.get("CONSTRUCTOR_PFX_CERTIFICATE_PASSWORD"),
```
marcoesters: The password should probably not be stored like that. See below on the signing command.

constructor/signing.py
```diff
@@ -120,6 +152,75 @@ class AzureSignTool(SigningTool):
...
+            "timestamp_server": os.environ.get(
+                "AZURE_SIGNTOOL_TIMESTAMP_SERVER_URL", "http://timestamp.sectigo.com"
+            ),
+            "timestamp_digest": os.environ.get("AZURE_SIGNTOOL_TIMESTAMP_DIGEST", "sha256"),
+            "file_digest": os.environ.get("AZURE_SIGNTOOL_FILE_DIGEST", "sha256"),
+            "access_token": os.environ.get("AZURE_SIGNTOOL_KEY_VAULT_ACCESSTOKEN"),
+            "secret": os.environ.get("AZURE_SIGNTOOL_KEY_VAULT_SECRET"),
+            "client_id": os.environ.get("AZURE_SIGNTOOL_KEY_VAULT_CLIENT_ID"),
+            "tenant_id": os.environ.get("AZURE_SIGNTOOL_KEY_VAULT_TENANT_ID"),
+        }
```
marcoesters: This contains a lot of sensitive data that is later added as plain text into the command. The environment variables should be used so that a traceback on a failed command doesn't expose the secrets.

marcoesters (commented):
Since I reviewed most of this already, I just gave it another once-over. I do have a security concern around signing though.

constructor/signing.py
lrandersson: Good call, see 23c040d

news/1251-msi-experimental-support
lrandersson: What about https://github.com/conda/constructor/issues/646? And https://github.com/conda/constructor/issues/967 of course

Let me know if you agree and I'll add those

pyproject.toml
lrandersson: I can see your argument for this but I like the idea of having this work out-of-the-box for users who want to test it out. To me experimental only means that the interface can change and/or is less tested compared to our other options, and just from convenience having everything installed at once seems like a reasonable expectation.

constructor/signing.py
marcoesters: `os.environ` would still add the values into the command. The issue isn't (just) that we store them, but that the values are added to the executed command in plain text when you use `check_call`. The subprocess object will contain `cmd`, which may be added to an error traceback and thus leak credentials.

lrandersson: I looked into this a bit further, for `AzureSignTool` can use `-kvm` flag with environment variables (which would solve this), but for `signtool` I cannot find any documentation with similar flags. Because of this I changed the approach to use `shell=True` which will not expose evaluated environment variables, instead (if there is a traceback) it will only show the name of the environment variable.
Doing this also made the two sign tool implementations identical so a lot of code could be removed:
See d5d0977

marcoesters: Why are we changing it compared to how it was handled before? For example: https://github.com/conda/constructor/blob/f80d676e442355e9fbe12eb9fba1aedf5746ec20/constructor/signing.py#L135-L142

lrandersson: Not sure I follow, with the last commit we changed it back to use the existing `get_signing_command` (i.e. the one you referenced).

marcoesters: Oops, yes, you're right.

news/1251-msi-experimental-support
marcoesters: Those look good. We probably need to also close some other issues as "not planned" afterwards.

lrandersson: See 09eda39

marcoesters (approved):

PR #1260, opened 2026-06-10
Add installer attributes for name, version, platform, and type to .installer.info

stacynoland:
### Description

This PR adds attributes to a `.installer.info` file stored in the `workspace` directory for persistent storage.

The installer attributes are stored in JSON format using the following keys:

- `name`
- `version`
- `platform`
- `type`

This information is being stored so other services, such as telemetry, can find details about the installer that was used to setup `conda`.

### Checklist - did you ...

### DIFF
--- a/constructor/preconda.py
+++ b/constructor/preconda.py
@@ -49,6 +49,7 @@ except ImportError:
     import ruamel_json as json
 
 files = (
+    ".installer.info",
     "pkgs/.constructor-build.info",
     "pkgs/urls",
     "pkgs/urls.txt",
@@ -164,6 +165,15 @@ def write_files(info: dict, workspace: str):
     os.makedirs(pkgs_dir, exist_ok=True)
     with open(join(pkgs_dir, ".constructor-build.info"), "w") as fo:
         json.dump(system_info(), fo)
+
+    out = {
+        "name": info.get("name"),
+        "version": info.get("version"),
+        "platform": info.get("platform"),
+        "type": info.get("installer_type"),
+    }
+    with open(join(workspace, ".installer.info"), "w") as fo:
+        json.dump(out, fo)
 
     all_urls = info["_urls"].copy()
     for env_info in info.get("_extra_envs_info", {}).values():
--- a/constructor/preconda.py
+++ b/constructor/preconda.py
@@ -49,7 +49,7 @@ except ImportError:
     import ruamel_json as json
 
 files = (
-    "pkgs/.constructor-build.info",
+    ".constructor-build.info",
     "pkgs/urls",
     "pkgs/urls.txt",
     "conda-meta/initial-state.explicit.txt",
@@ -113,7 +113,7 @@ def write_index_cache(info: dict, dst_dir: str, used_packages: list[str]):
             os.unlink(join(cache_dir, cache_file))
 
 
-def system_info():
+def system_info(info: dict):
     out = {
         "constructor": CONSTRUCTOR_VERSION,
         "conda": CONDA_INTERFACE_VERSION,
@@ -123,6 +123,10 @@ def system_info():
         "machine": platform.machine(),
         "platform_full": platform.version(),
     }
+    out["installer_name"] = info.get("installer_name", "Unknown")
+    out["installer_version"] = info.get("installer_version", "Unknown")
+    out["installer_platform"] = info.get("installer_platform", "Unknown")
+    out["installer_type"] = info.get("installer_type", "Unknown")
     if sys.platform == "darwin":
         out["extra"] = platform.mac_ver()
     elif sys.platform.startswith("linux"):
@@ -162,8 +166,8 @@ def write_files(info: dict, workspace: str):
     os.makedirs(join(workspace, "conda-meta"), exist_ok=True)
     pkgs_dir = join(workspace, "pkgs")
     os.makedirs(pkgs_dir, exist_ok=True)
-    with open(join(pkgs_dir, ".constructor-build.info"), "w") as fo:
-        json.dump(system_info(), fo)
+    with open(join(workspace, ".constructor-build.info"), "w") as fo:
+        json.dump(system_info(info), fo)
 
     all_urls = info["_urls"].copy()
     for env_info in info.get("_extra_envs_info", {}).values():
--- a/constructor/preconda.py
+++ b/constructor/preconda.py
@@ -169,7 +169,7 @@ def write_files(info: dict, workspace: str):
     out = {
         "name": info.get("name"),
         "version": info.get("version"),
-        "platform": info.get("platform"),
+        "platform": info.get("_platform"),
         "type": info.get("installer_type"),
     }
     with open(join(workspace, ".installer.info"), "w") as fo:
--- a/constructor/preconda.py
+++ b/constructor/preconda.py
@@ -49,7 +49,7 @@ except ImportError:
     import ruamel_json as json
 
 files = (
-    ".constructor-build.info",
+    "pkgs/.constructor-build.info",
     "pkgs/urls",
     "pkgs/urls.txt",
     "conda-meta/initial-state.explicit.txt",
@@ -113,7 +113,7 @@ def write_index_cache(info: dict, dst_dir: str, used_packages: list[str]):
             os.unlink(join(cache_dir, cache_file))
 
 
-def system_info(info: dict):
+def system_info():
     out = {
         "constructor": CONSTRUCTOR_VERSION,
         "conda": CONDA_INTERFACE_VERSION,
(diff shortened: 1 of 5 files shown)


stacynoland:
pre-commit.ci autofix

constructor/preconda.py
```diff
@@ -123,6 +123,10 @@ def system_info():
         "machine": platform.machine(),
         "platform_full": platform.version(),
     }
+    out["installer_name"] = info.get("installer_name", "Unknown")
+    out["installer_version"] = info.get("installer_version", "Unknown")
+    out["installer_platform"] = info.get("installer_platform", "Unknown")
```
lrandersson:
```suggestion
    out["installer_name"] = info.get("name", "Unknown")
    out["installer_version"] = info.get("version", "Unknown")
    out["installer_platform"] = info.get("_platform", "Unknown")
```
I looked at the keys again that I mentioned yesterday and I realized that was incorrect. I added this suggestion which should be accurate:
1. `name` from `construct.yaml`
2. `version` from `construct.yaml`
3. `_platform` is set via `main.py`
4. `installer_type` also set via `main.py`

I also wonder if we should simply have empty strings as default but no strong opinion, I guess the consumer of this data might have an opinion of this.
Perhaps also `installer_name` and `installer_version` should not even have a default since something must be fundamentally wrong if it not set.

constructor/preconda.py
```diff
@@ -49,7 +49,7 @@
     import ruamel_json as json
 
 files = (
-    "pkgs/.constructor-build.info",
+    ".constructor-build.info",
```
marcoesters: I'm not sure about whether to re-use that file. The file contains information about the build process and contains information on `conda`, `python`, etc. from the build environment. Adding information of the installer could imply that these are the version of `conda`, `python`, etc. that are packaged in the installer.

I think a separate file makes more sense.

constructor/preconda.py
marcoesters: They're all guaranteed to be set. I think a simple `get()` without default values is good here - in the unlikely case that a value isn't set, we'd get `null` instead of a special string in the JSON file. Empty string is a good alternative, which I'd prefer over `"Unknown"`.

marcoesters (commented):
I think we should use a separate file to not conflate installer and build environment information. I would also like to see the changes be covered by a test - I recommend just modifying an existing one.

stacynoland:
> I think we should use a separate file to not conflate installer and build environment information. I would also like to see the changes be covered by a test - I recommend just modifying an existing one.

The installer attributes are now saved to a `.installer.info` file in the installer/workspace directory and I reverted the changes we had made to the `.constructor-build.info` file.

I'm still working on getting it covered by a test, but I will have that for review by tomorrow.

constructor/preconda.py
stacynoland: @lrandersson Thanks for catching the `info` dict keys issue. I made sure this was updated.

@marcoesters I also removed the default value as suggested on the `get()` method.

lrandersson (commented):
I think this looks good. As you mentioned maybe a simple unit test would be good to add and I think we should not hardcode `.installer.info` in both `preconda.py` and `winexe.py`, instead define a variable somewhere that can be utilized.

marcoesters:
> I think this looks good. As you mentioned maybe a simple unit test would be good to add and I think we should not hardcode `.installer.info` in both `preconda.py` and `winexe.py`, instead define a variable somewhere that can be utilized.

I think changing an existing test like `test_example_miniforge` is sufficient. The current hardcoding is consistent with existing code, actually. A general refactor to replace the hard-coding with variables is a good issue though.

lrandersson:
> > I think this looks good. As you mentioned maybe a simple unit test would be good to add and I think we should not hardcode `.installer.info` in both `preconda.py` and `winexe.py`, instead define a variable somewhere that can be utilized.
> 
> I think changing an existing test like `test_example_miniforge` is sufficient. The current hardcoding is consistent with existing code, actually. A general refactor to replace the hard-coding with variables is a good issue though.

Created https://github.com/conda/constructor/issues/1261

stacynoland:
pre-commit.ci autofix

constructor/preconda.py
```diff
@@ -165,6 +166,15 @@ def write_files(info: dict, workspace: str):
     with open(join(pkgs_dir, ".constructor-build.info"), "w") as fo:
         json.dump(system_info(), fo)
 
+    out = {
+        "name": info.get("name"),
+        "version": info.get("version"),
+        "platform": info.get("_platform"),
+        "type": info.get("installer_type"),
+    }
+    with open(join(workspace, ".installer.info"), "w") as fo:
```
marcoesters, line 175: Should this get the `JSON` extension since it's a JSON file? I know the build info is `.info`, but `.json` is more transparent.

tests/test_examples.py
```diff
@@ -596,6 +597,18 @@ def test_example_miniforge(tmp_path, request, example):
             # Check that key metadata files are in place
             assert install_dir.glob("conda-meta/*.json")
             assert install_dir.glob("pkgs/cache/*.json")  # enables offline installs
+            # Check that the installer info file is in place
+            assert (install_dir / ".installer.info").is_file()
+            with open(install_dir / ".installer.info") as f:
+                installer_info = json.load(f)
```
```suggestion
            info_file = install_dir / ".installer.info"
            assert info_file.is_file()
            installer_info = json.loads(info_file.read_text())
```

tests/test_examples.py
```diff
@@ -596,6 +597,18 @@ def test_example_miniforge(tmp_path, request, example):
...
+            assert (
+                installer_info["name"] == "Miniforge3"
+                if example == "miniforge"
+                else "Miniforge3-mamba2"
+            )
+            assert installer_info["version"] == "25.0.0-1" if example == "miniforge" else "25.1.1-0"
+            assert installer_info["platform"] in SUPPORTED_PLATFORMS
```
marcoesters: That doesn't test that the platform is correct, just that it's a supported platform.

tests/test_examples.py
```diff
@@ -596,6 +597,18 @@ def test_example_miniforge(tmp_path, request, example):
...
+            # Check that the installer info file is in place
+            assert (install_dir / ".installer.info").is_file()
+            with open(install_dir / ".installer.info") as f:
+                installer_info = json.load(f)
+            assert (
+                installer_info["name"] == "Miniforge3"
+                if example == "miniforge"
+                else "Miniforge3-mamba2"
+            )
+            assert installer_info["version"] == "25.0.0-1" if example == "miniforge" else "25.1.1-0"
```
marcoesters: I'm wondering if all these `if example == "miniforge" else` instances should just be parametrized. I don't know if that makes it more legible though.

stacynoland: `example` is the paramaterized value for this test, isn't it? `@pytest.mark.parametrize("example", ("miniforge", "miniforge-mamba2"))`

marcoesters: Right, I'm wondering if it makes it more or less legible if we add the expected results (as a dictionary?) as a parameter, too. We have at least three of these `if-else` ternary operators in the test.

stacynoland: Ah okay I understand what you mean. That's a fair point. I am always a fan of less repetitive code.

stacynoland:
pre-commit.ci autofix

tests/test_examples.py
stacynoland: @marcoesters Yeah I tried a single `if-elif` vs parameterized values, and the latter was the way to go. That's good to go now.

tests/test_examples.py
stacynoland: We can use the `platform` module and piece together each "platform" value that way to check against it, replicating the format `SUPPORTED_PLATFORMS` expects. That's the only option I see here, because this value isn't provided by the `construct.yaml` files.

constructor/preconda.py
stacynoland: `.constructor-build.info` is also a `JSON` file. I was following that as an example, so I guess that question would apply to both of these, correct?

marcoesters: I don't like that either. 😅 I don't have a strong opinion either way.

tests/test_examples.py
marcoesters: Since you don't cross-build in this test, you could use `cc_platform`: https://github.com/conda/constructor/blob/f80d676e442355e9fbe12eb9fba1aedf5746ec20/constructor/conda_interface.py#L98

That's the default value for the platform.

stacynoland: I kept seeing that in places and wasn't 100% sure if that was it. I'll make that change and send it along in the next push.

constructor/preconda.py
stacynoland: Yeah, I don't have a strong opinion either, but if we do make a choice I just want to stay consistent 🤓

stacynoland:
pre-commit.ci autofix

lrandersson (approved):

marcoesters (approved):

PR #1263, opened 2026-06-15
Add CLI option to render `construct.yaml` files

marcoesters:
### Description

Add a CLI option to render `construct.yaml` files. Rendering these files can be useful for debugging or to generate a snapshot for installer builds.

### Checklist - did you ...

### DIFF
--- a/constructor/construct.py
+++ b/constructor/construct.py
@@ -16,6 +16,7 @@ import sys
 from functools import partial
 from os.path import dirname
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


lrandersson (approved):
I think the changes look good. My only thought was if we want to add parts of the error message before exiting in case we can not open `path` in the function `render` (at line 104) in `construct.py`). However, considering this is likely not going to happen that often I am OK leaving it as it.

PR #1264, opened 2026-06-16
Fix MSI test for Windows Canary build

lrandersson:
### Description
This PR updates a test that was failing in the Windows canary build https://github.com/conda/constructor/actions/runs/27581405719/job/81677012823.
The issue was that the test was referencing a file (an image in this case) which isn't available because, Canary builds do:
- First builds a conda package from the source
- Then tests the installed package in an isolated environment (example) `D:\a\constructor\constructor\pkgs\constructor_1781618608424\test_tmp\`
- Only files listed in `recipe/meta.yaml` `source_files` are copied into this test environment
  - This is where it breaks because the file is not part of the environment.

Therefore, dynamically create a temporary image as part of the test and dont reference the example file. Originally when implementing this I only referenced the image because I just needed _any_ image in order to do the test.

### Checklist - did you ...

### DIFF
--- a/tests/test_briefcase.py
+++ b/tests/test_briefcase.py
@@ -8,6 +8,7 @@ except ModuleNotFoundError:
     import tomli as tomllib
 
 import pytest
+from PIL import Image
 
 from constructor.briefcase import (
     Payload,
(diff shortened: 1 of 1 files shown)


marcoesters:
> Only files listed in recipe/meta.yaml source_files are copied into this test environment 

Why is this solution preferred over just adding the missing file into `source_files`?

lrandersson:
> > Only files listed in recipe/meta.yaml source_files are copied into this test environment
> 
> Why is this solution preferred over just adding the missing file into `source_files`?

That's a good question and definitely one of the options. I chose this over the other because I didn't want to couple contents of `source_files` to this specific test. The test doesn't need the example - just any BMP.

Let me know if you've any preference.

marcoesters (approved):

Issue #1146, opened 2025-12-19
Improve testing of examples


lrandersson:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

Currently when we run `tests_examples.py` we test each installer via a for-loop, in most cases it's written as:
```
def test_something():
    for installer, install_dir in create_installer:
       ...
```
This means that for platforms that support multiple installer types will in some cases not run all the tests upon a test failure. As an example, considering `macOS` where we have `pkg` and `sh`, and in the future with Windows where we might have both `exe` and `msi`; if the for-loop fails in the first iteration, the second type of installer will never be tested. Additionally, if it fails in the second iteration (or later), all the test output from the previous installers is written to the log which clutters the console output with unnecessary information. For example, for Windows-installers I'd say this makes test debugging extremely difficult because you will have all of the NSIS output written even if it was the MSI installer that failed.
Currently on the feature branch where we have both `msi` and `exe`, the test job console output has to be opened in a separate tab with raw text because the browser is not able to display all of the text.

### Why is this needed?

Improve testing infrastructure and simplify debugging in case of test failure. In detail:
1. Tests are not blocked by early failures (i.e. if one installer type fails while there are more remaining)
2. Test output can be read more easily which will improve debugging capabilities
3. Depending on the design choice, it can make it possible to run tests for specific installer types locally.

### What should happen?

Installer types should be tested without the use of a loop, since today the design is for each test-function, loop over each installer type, which leads to the issues mentioned above upon test failures.

PR #1266, opened 2026-06-18
Refactor the test examples to parametrize per installer type

lrandersson:
### Description
Solves https://github.com/conda/constructor/issues/1146. The file `test_examples.py` runs every applicable installer type for an example in a single constructor call and loops over them, which interleaves build/run output and gives no test isolation. When one type fails, the others are masked or skipped, making CI failures painful to debug.

This PR adds the infrastructure to run each installer type as its own pytest item:
- A flag `--installer-type` to allow testing a specific installer type via the CLI
- Removes the for-loop for iterating over installer types - now uses pytest parametrization (which is based on the `installer_types` field in each example recipe).
- For demonstrating the changes, this PR migrates only `test_example_miniforge` (split into `miniforge` and `miniforge-mamba2`, each parametrized over its own recipe's types).

The remaining example tests are intentionally left unchanged here to keep this PR small and easy to review. A follow-up PR is expected once this is approved and merged.

### Checklist - did you ...

### DIFF
--- a/constructor/main.py
+++ b/constructor/main.py
@@ -211,6 +211,7 @@ def main_build(
     conda_exe: str = "conda.exe",
     config_filename: str = "construct.yaml",
     debug: bool = False,
+    installer_type: str | None = None,
 ):
     logger.info("platform: %s", platform)
     if not os.path.isfile(conda_exe):
@@ -245,6 +246,13 @@ def main_build(
                 "Install Docker Buildx to proceed, or remove `docker_image_format` to "
                 "generate the Dockerfile without building the portable image."
             )
+    if installer_type:
+        if installer_type not in itypes:
+            sys.exit(
+                f"Error: installer type '{installer_type}' not available for this "
+                f"config/platform; allowed: {', '.join(itypes)}"
+            )
+        itypes = (installer_type,)
 
     if platform != cc_platform and "pkg" in itypes and not cc_platform.startswith("osx-"):
         sys.exit("Error: cannot construct a macOS 'pkg' installer on '%s'" % cc_platform)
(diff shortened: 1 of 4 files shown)


lrandersson:
We can now see in the macOS CI output (as an example):
```
tests/test_examples.py::test_example_miniforge[sh] PASSED                [ 18%]
tests/test_examples.py::test_example_miniforge[pkg] PASSED               [ 20%]
tests/test_examples.py::test_example_miniforge_mamba2[sh] PASSED         [ 22%]
tests/test_examples.py::test_example_miniforge_mamba2[pkg] PASSED        [ 24%]
```

constructor/main.py
```diff
@@ -245,6 +246,13 @@ def main_build(
                 "Install Docker Buildx to proceed, or remove `docker_image_format` to "
                 "generate the Dockerfile without building the portable image."
             )
+    if installer_type:
+        if installer_type not in itypes:
```
marcoesters: `itypes` are the installer types that are read from the `construct.yaml` file. What this does is ensure that `--installer-type` is consistent with what you have in the `construct.yaml` file.

I think what you actually want is to do something like this:

```
if installer_type:
    info["installer_type"] installer_type
itypes = get_installer_type(info)
```

lrandersson: Great catch, see 6cff5e3.

constructor/main.py
```diff
@@ -618,6 +620,15 @@ def main(argv=None):
         default="construct.yaml",
     )
 
+    p.add_argument(
+        "--installer-type",
+        help="Build only this installer type (sh, pkg, exe, msi). "
+        "Primarily for testing; overrides 'installer_type' from the config.",
```
marcoesters, line 626: I think this has more uses than testing. Since this is a new feature, we should put this into a news file.

lrandersson: See 8c32a57

marcoesters (approved):

PR #1276, opened 2026-06-30
MSI: Add support for INSTALLER_UNATTENDED

lrandersson:
### Description

Add support for the environment variable INSTALLER_UNATTENDED for MSI installers (this is fixed upstream https://github.com/beeware/briefcase-windows-app-template/pull/94).

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/constructor/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [x] Add / update outdated documentation?

PR #1277, opened 2026-06-30
Fix installer option inconsistency between EXE/MSI

lrandersson:
### Description
I noticed during local testing that two of the installer options default values were hardcoded `False` instead of using the same approach as the EXE installers. This PR fixes that and updates a test to account for this (since these were default `False` the test example pre/post uninstall/install scripts were never run by the MSI installer).

I also did another check to see if any other option was inconsistently set but these two were the only ones I found.

### Checklist - did you ...

### DIFF
--- a/examples/scripts/post_install.bat
+++ b/examples/scripts/post_install.bat
@@ -2,19 +2,23 @@ echo Added by post-install script > "%PREFIX%\post_install_sentinel.txt"
 if not "%INSTALLER_NAME%" == "Scripts" exit 1
 if not "%INSTALLER_VER%" == "1.0.0" exit 1
 if not "%INSTALLER_PLAT%" == "win-64" exit 1
-if not "%INSTALLER_TYPE%" == "EXE" exit 1
-if not "%INSTALLER_UNATTENDED%" == "1" exit 1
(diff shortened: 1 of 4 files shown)


examples/scripts/post_install.bat
```diff
@@ -2,19 +2,23 @@ echo Added by post-install script > "%PREFIX%\post_install_sentinel.txt"
 if not "%INSTALLER_NAME%" == "Scripts" exit 1
 if not "%INSTALLER_VER%" == "1.0.0" exit 1
 if not "%INSTALLER_PLAT%" == "win-64" exit 1
-if not "%INSTALLER_TYPE%" == "EXE" exit 1
-if not "%INSTALLER_UNATTENDED%" == "1" exit 1
+if not "%INSTALLER_TYPE%" == "EXE" if not "%INSTALLER_TYPE%" == "MSI" exit 1
+rem INSTALLER_UNATTENDED is only set by EXE installers.
```
lrandersson: This comment (and in `pre_install.bat`) is `True` only for this branch but something that needs to be accounted for once https://github.com/conda/constructor/pull/1276 has been merged.

marcoesters (approved):

PR #1278, opened 2026-07-01
Migrate more test examples to use pytest parametrization

lrandersson:
### Description
This is a follow-up of https://github.com/conda/constructor/pull/1266, since that PR added the foundation, this PR migrates about half of the tests to the new layout (to simplify review). I am planning to a third (last) PR with the remaining tests.

### Checklist - did you ...

### DIFF
--- a/examples/azure_signtool/construct.yaml
+++ b/examples/azure_signtool/construct.yaml
@@ -3,7 +3,9 @@
 
 name: Signed_AzureSignTool
 version: 1.0.0
-installer_type: [exe, msi]  # [win]
+installer_type:
+  - exe
+  - msi
 channels:
   - https://repo.anaconda.com/pkgs/main/
 specs:
--- a/examples/azure_signtool/construct.yaml
+++ b/examples/azure_signtool/construct.yaml
@@ -3,7 +3,7 @@
 
 name: Signed_AzureSignTool
 version: 1.0.0
-installer_type: [exe, msi]
+installer_type: [exe, msi]  # [win]
 channels:
   - https://repo.anaconda.com/pkgs/main/
 specs:
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -829,31 +829,35 @@ def test_example_customized_welcome_conclusion(tmp_path, request, installer_type
 
 
 @pytest.mark.parametrize("extra_pages", ("str", "list"))
+@pytest.mark.parametrize(
+    "installer_type", installer_types_for_example(_example_path("exe_extra_pages"))
+)
 @pytest.mark.skipif(sys.platform != "win32", reason="Windows only")
-def test_example_extra_pages_win(tmp_path, request, extra_pages, monkeypatch):
+def test_example_extra_pages_win(tmp_path, request, extra_pages, installer_type, monkeypatch):
     if extra_pages == "list":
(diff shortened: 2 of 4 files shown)


tests/test_examples.py
```diff
@@ -664,12 +664,18 @@ def _install_dir_for(
...
     config_filename: str,
     with_spaces: bool,
+    short: bool = False,
 ) -> Path:
     install_dir_prefix = "i n s t a l l" if with_spaces else "install"
     if installer.suffix == ".pkg" and ON_CI:
         return Path("~").expanduser() / calculate_install_dir(input_dir / config_filename)
     elif installer.suffix == ".msi":
         return calculate_msi_install_path(input_dir / config_filename)
+    elif short:
```
lrandersson, line 675: One of the examples reached maximum path length with the new pytest parametrization. pytest automatically creates unique tmp directories for each test parameter run (like `/tmp/pytest-of-<username>/pytest-<run_number>/<test_module_name>/<test_function_name><[parameter_id]>/`), because of this, one of the examples reached the max path limit (it was very close to the limit already). Therefore this new arg `short` simply forces it into a shorter name.

examples/azure_signtool/construct.yaml
```diff
@@ -3,7 +3,7 @@
 
 name: Signed_AzureSignTool
 version: 1.0.0
-installer_type: [exe, msi]
+installer_type: [exe, msi]  # [win]
```
marcoesters: The entire installer should only run on Windows, so I don't think we need to add that selector.

lrandersson: Originally I did not add the selector but I found a bug which resulted in me adding it. Without `# [win]` the regex matches `[exe, msi]` as a selector because the regex looks for any [...] at the end of a line. It tries to evaluate `exe, msi` as a Python expression, which fails because `exe` and `msi` aren't defined variables.
The `# [win]` selector is needed to prevent the YAML list value from being misinterpreted as a platform selector.

I am planning to fix it but I propose to do it in a separate PR to minimize the changes in this PR since it's not a new bug.

marcoesters: Then why not format it as YAML?

```
installer_type:
  - exe
  - msi
```

lrandersson: Good call, I was so locked in on the formats already in our code-base and forgot this was an option. See 3e648b4.

lrandersson: Actually I think the selector is needed:
```
INTERNALERROR> SystemExit: Error: invalid installer type 'exe' for linux; allowed: sh
```

marcoesters: This is happening during the unit tests. I think there is something going on during the import step when the tests are collected that the selector may be masking.

lrandersson: I did a little more changes, with some minor refactoring to be able to filter out this scenario fc7df2e. The refactoring allows `installer_types_for_example()` to handle platform-incompatible examples by catching `InvalidInstallerTypeError` instead of `SystemExit`, so tests are properly skipped without needing platform specific selectors in the YAML file or manual `skipif` decorators.

marcoesters (approved):

PR #1289, opened 2026-07-16
Use `InstallerTypes` for installer type handling

### LINKED ISSUE #1285: Make `InstallerTypes` the single source of truth for installer type handling
### Checklist

### What is the idea?

Make `InstallerTypes` (already defined as a `StrEnum` in `constructor/_schema.py`, covering `ALL`, `EXE`, `MSI`, `PKG`, `SH`, `DOCKER`) the single source of truth for installer type handling throughout the codebase, instead of raw string literals.

Today the enum is only used as a type annotation on the `installer_type` field. Everywhere else — `main.py`'s `os_allowed`/`all_allowed`, and comparisons in `main.py`, `osxpkg.py`, `imaging.py`, `briefcase.py` — installer types are handled as plain strings (`"sh"`, `"pkg"`, `"exe"`, `"msi"`, `"docker"`, `"all"`), duplicated independently of the schema enum.

### Why is this needed?

1. `main.py`'s allowed-type tables are hardcoded string tuples, fully disconnected from `InstallerTypes`. Adding a new installer type means updating both places by hand, with nothing enforcing they stay in sync.
2. String comparisons against installer type literals are scattered across several files, are typo-prone, and aren't statically checked.
3. Error messages listing valid installer type values are also hand-maintained strings that can drift from the actual supported set.
4. The `--installer-type` CLI flag has no `choices=` constraint, so invalid values are only caught later through manual validation, giving a less helpful CLI error than argparse could provide for free.

### What should happen?

1. `InstallerTypes` becomes the canonical definition of valid installer types, including `DOCKER` and `ALL` (already present as members — docker does not need its own enum).
2. Allowed-type tables in `main.py` are derived from `InstallerTypes` rather than duplicating string tuples.
3. Comparison sites across `main.py`, `osxpkg.py`, `imaging.py`, `briefcase.py` switch to comparing against `InstallerTypes` members.
4. Optionally, the `--installer-type` CLI flag gains `choices=` derived from `InstallerTypes` for earlier validation and clearer errors.

Since `InstallerTypes` is a `StrEnum`, its members compare equal to and serialize as plain strings, so `construct.yaml` parsing, `info.json` metadata, and existing string-based test parametrization should continue to work without changes.

### Additional Context

**Open questions for discussion:**

- **Test literals:** Should `tests/test_examples.py` / `tests/test_main.py` also be migrated to use `InstallerTypes`, or is leaving them as raw strings acceptable given `StrEnum` compatibility?
- **CLI behavior change:** Adding `choices=` to `--installer-type` changes the argparse error message for invalid values — minor user-facing change worth flagging in the PR.
- **Docs generation:** `CONSTRUCT.md` / `docs/source/construct-yaml.md` are generated from hand-written docstring prose on the `installer_type` field, not from iterating `InstallerTypes` members. Keeping that docstring in sync with the enum is a separate concern and not required for this change.

There are likely more but just for the sake of bookkeeping.

lrandersson:
### Description

This PR implements https://github.com/conda/constructor/issues/1285.

### Checklist - did you ...

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -23,6 +23,7 @@ from conda.core.prefix_data import PrefixData
 from conda.models.version import VersionOrder as Version
 from ruamel.yaml import YAML
 
+from constructor._schema import InstallerTypes
 from constructor.conda_interface import cc_platform
 from constructor.construct import parse as parse_construct
 from constructor.exceptions import InvalidInstallerTypeError
@@ -53,6 +54,15 @@ try:
 except ImportError:
     COV_CMD = ()
 
+
+# Installer types that produce a single file named after the type (e.g. "foo.sh"),
+# in the order shell installers should run before GUI ones.
+FILE_INSTALLER_TYPES = (
+    InstallerTypes.SH,
+    InstallerTypes.PKG,
+    InstallerTypes.EXE,
+    InstallerTypes.MSI,
+)
 
 pytestmark = pytest.mark.examples
 REPO_DIR = Path(__file__).parent.parent
@@ -616,7 +626,7 @@ def _run_installer(
 def _build_installers(
     input_dir: Path,
     workspace: Path,
-    installer_type: str | None = None,
+    installer_type: InstallerTypes | None = None,
     conda_exe=CONSTRUCTOR_CONDA_EXE,
     debug=CONSTRUCTOR_DEBUG,
     timeout=420,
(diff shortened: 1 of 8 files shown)


tests/test_examples.py
```diff
@@ -54,6 +55,15 @@
     COV_CMD = ()
 
 
+# Installer types that produce a single file named after the type (e.g. "foo.sh"),
+# in the order shell installers should run before GUI ones.
```
marcoesters, line 59: Can you elaborate on why they should run first?

lrandersson: I only added that because of this already existing comment in `test_examples.py`, it wasn't completely clear to me either why but I didn't want to change this:
```
    def _sort_by_extension(path):
        "Return shell installers first so they are run before the GUI ones"
```

marcoesters (approved):
Assuming the Intel macOS tests pass eventually.

PR #1297, opened 2026-07-20
Fix Windows Canary Build collection crash

lrandersson:
### Description

Windows Canary Build has been failing (https://github.com/conda/constructor/actions/runs/29743641137/job/88408197352) with:
```
INTERNALERROR> FileNotFoundError: ...\test_tmp\examples\customize_controls\construct.yaml
```
Cause: conda-build's recipe test step only copies a few `examples/` dirs into its test tree, but `test_examples.py` now evaluates `installer_types_for_example()` for every example at module/collection time, so missing files crash collection entirely.

Fix: `installer_types_for_example()` skips (with reason) if `construct.yaml` is missing, instead of raising.

### Checklist - did you ...

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -239,7 +239,7 @@ jobs:
         uses: conda/actions/canary-release@7f6830b1428a9bd47f0b068892c77eae95207037 # v26.1.0
         # TODO: remove 'dev-ra-fix-test-issue' from this condition once the
         # Canary Build fix on that branch is verified green and merged.
-        if: github.ref_name == 'main' || startsWith(github.ref_name, 'feature/') || github.ref_name == 'dev-ra-fix-test-issue'
+        # if: github.ref_name == 'main' || startsWith(github.ref_name, 'feature/')
         with:
           package-name: ${{ github.event.repository.name }}
           subdir: ${{ matrix.subdir }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -236,8 +236,10 @@ jobs:
           python -m build --sdist --wheel . --outdir dist
 
       - name: Create and upload canary build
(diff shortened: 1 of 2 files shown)


tests/test_examples.py
```diff
@@ -702,7 +702,21 @@ def installer_types_for_example(
     Returns a tuple with a single skip marker if the example's installer types
     are not valid for the current platform (e.g., Windows-only examples on Linux).
     """
-    info = parse_construct(str(example_path / config_filename), platform=cc_platform)
+    config_path = example_path / config_filename
```
lrandersson, line 705: An alternative approach is to simply add `examples/*` to the package source files but not sure if we want to bloat the entire package.

.github/workflows/main.yml
```diff
@@ -237,11 +237,15 @@ jobs:
 
       - name: Create and upload canary build
         uses: conda/actions/canary-release@7f6830b1428a9bd47f0b068892c77eae95207037 # v26.1.0
-        if: github.ref_name == 'main' || startsWith(github.ref_name, 'feature/')
+        # TODO: remove 'dev-ra-fix-test-issue' from this condition once the
+        # Canary Build fix on that branch is verified green and merged.
+        if: github.ref_name == 'main' || startsWith(github.ref_name, 'feature/') || github.ref_name == 'dev-ra-fix-test-issue'
```
jaimergp: Why don't we enable this for PRs to test it here? As long as we don't upload it's ok.

```suggestion
        # if: github.ref_name == 'main' || startsWith(github.ref_name, 'feature/')
```

lrandersson: I made some changes to that file now while accounting for your input. I never knew if there was any historical reason for not running canary tests on PRs like this but I'm more than happy to include it.

marcoesters (approved):

PR #1300, opened 2026-07-24
Record conda-ship runtime ownership during installation

### LINKED ISSUE #1299: Add a constructor step to record conda-ship installation metadata
## Use case

A constructor-based distribution may include a prebuilt executable produced by conda-ship, such as a standalone `conda` executable.

The executable already knows how to bootstrap its managed prefix, including from an embedded bundle for offline installation. This issue is not about air-gapped installers or choosing a conda-ship artifact layout.

The missing integration is to record who may replace the installed executable later. For a standalone `conda` runtime, this is the outer executable installed by constructor, not the `conda` package inside its managed prefix. A later `conda self update` needs to know whether the runtime may replace that executable itself or whether an external installer owns it and should provide the user with an installer-specific update instruction.

## Existing conda-ship interface

conda-ship provides the local `v1/record-installation` helper. An installer invokes the installed executable once to record these values in the existing `.RUNTIME_NAME.json` file:

- installation label, such as `constructor`
- ownership, either `direct` or `external`
- stable installed executable path
- optional update instruction for external ownership

This does not require a constructor-specific metadata format or different executable bytes.

## Possible constructor integration

When a constructor-based distribution needs this, add a construct option that invokes `v1/record-installation` after installing the executable. If the metadata is required for correct update behavior, the invocation must be part of the installer rather than a user post-install script because user scripts can be disabled.

The first consumer should determine which constructor backends and fields are actually needed. The earlier draft in #1300 attempted SH, PKG, EXE, and MSI support without a downstream consumer and was closed.

## Status

There is no concrete constructor-based consumer yet, so implementation is deferred until one has a construct file and a defined executable update model. Homebrew and Python installer detection are separate work tracked in jezdez/conda-runtime#15.

jezdez:
### Description

Add the optional `conda_ship_runtime` construct setting for SH, PKG, EXE, and MSI installers.

The mandatory installer path invokes `v1/record-installation` on the stamped executable after package installation. It records ownership, installation kind, stable executable path, and an optional external instruction in the existing runtime metadata file.

This is not a user post-install script and cannot be skipped through user-script options.

Depends on jezdez/conda-ship#94.

Closes #1299.

### Checklist - did you ...

- [x] Add a file to the `news` directory ([using the template](https://github.com/conda/constructor/blob/main/news/TEMPLATE)) for inclusion in the next release notes?

### DIFF
--- a/CONSTRUCT.md
+++ b/CONSTRUCT.md
@@ -312,6 +312,32 @@ a dictionary, which will be converted to a YAML string for writing. _Note:_ if t
 option is used, then all other options related to the construction of a `.condarc`
 file (`write_condarc`, `conda_default_channels`, etc.) are ignored.
 
+### `conda_ship_runtime`
+
+Record a conda-ship stamped executable as part of the mandatory installer
+transaction. The executable must already be included in the installed
+prefix, either through a package or `extra_files`.
+
+Constructor records `direct` ownership and the `constructor` installation
+kind by default. Set `managed_prefix` to a separate subdirectory of the
+constructor prefix. The runtime executable must live outside that
+subdirectory so it can bootstrap the managed prefix. SH installers may
+omit `managed_prefix` to initialize the runtime's stamped per-user default.
+PKG, EXE, and MSI installers require it because their mandatory
+installation steps may run with elevated credentials. An installer with
+its own ongoing updater may instead set `ownership` to `external` and
+provide an `instruction`. EXE runtime values cannot contain single or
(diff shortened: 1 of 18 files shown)


jezdez:
Closing this draft because no downstream construct currently consumes `conda_ship_runtime`. The direct and external ownership work can proceed in conda-ship and conda-runtime without adding mandatory integration to four constructor backends. Issue #1299 remains open for a concrete constructor-based distribution.

PR #1310, opened 2026-08-03
Run test examples in more parallel instances

lrandersson:
### Checklist - did you ...

- [ ] Add a file to the `news` directory ([using the template](https://github.com/conda/constructor/blob/main/news/TEMPLATE)) for the next release's release notes?
     
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

Issue #1306, opened 2026-07-28
Investigate parallelizing `test_examples.py` across CI runners


lrandersson:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

`test_examples.py` currently runs as a single serial `pytest` step per OS/Python/conda-standalone matrix entry in `main.yml`. Investigate splitting it across multiple parallel jobs/runners per OS, and decide on an approach.

Options to look into:
- **`pytest-split`** (available on conda-forge) — balances test groups by recorded duration; needs a `--durations`-based baseline, or falls back to even count-splitting.
- **Manual `-k`/marker-based grouping** — no new dependency; tests grouped by hand into buckets and run separately per shard via `-k`.

Note: some tests touch OS-global state (Windows registry, Start Menu, macOS keychain, a hardcoded `C:/ProgramData/conda/.condarc`, fixed Docker image tags). Running shards as separate CI jobs/runners avoids collisions from this; running them concurrently on one machine (e.g. `pytest-xdist -n auto` in a single job) would not, without further changes.

### Why is this needed?

Reduce test times, Illustrative arithmetic (current full-run time ÷ N, no overhead accounted for):

| OS | Full run | ÷2 | ÷3 | ÷4 |
|---|---|---|---|---|
| Windows | ~100 min | ~50 min | ~33 min | ~25 min |
| macOS | ~85 min | ~43 min | ~28 min | ~21 min |
| Linux | ~30 min | ~15 min | ~10 min | ~8 min |

### What should happen?

Examples split into multiple runners, reducing overall testing time.

### Additional Context

**Decisions needed**

1. Which mechanism to use (`pytest-split` vs. manual grouping).
2. Number of shards per OS.
3. Whether shard boundaries should be based on measured per-test timing (`pytest --durations=0`) rather than a manual guess.
4. How to handle steps that assume one test run per matrix entry: coverage upload (Codecov `integration` flag), the `installers-*` artifact upload, and the `check-docs-schema` step.

Notes:
- `pytest-xdist`-style same-machine parallelization should not be attempted without first auditing tests that mutate shared OS-global state (e.g. `test_initialization` registry writes, `conftest.py` keychain mutation, `test_uninstallation_standalone`'s hardcoded `.condarc` path, `test_docker_image_build`'s fixed image tag).
- This issue is for deciding shard count and split mechanism; implementation should follow in a separate PR.

lrandersson:
[...]

PR #1312, opened 2026-08-03
Split test examples using pytest-split and improve test clean-up

lrandersson:
### Description
This PR tests a different approach to solve https://github.com/conda/constructor/issues/1306, related PR with another approach in https://github.com/conda/constructor/pull/1310.
In this PR we use `pytest-split` and simply separates the Windows test examples into 3 groups and macOS into 2 groups. 
https://github.com/jerry-git/pytest-split/tree/master

I also expanded the scope to include some test cleanup to minimize the chances of more intermittent issues with memory/disk-space related CI issues. Also removed the artifact archiving since it seems not to be used.

Benchmarks:
https://github.com/conda/constructor/issues/1306#issuecomment-5196606752

### Checklist - did you ...


pyproject.toml
```diff
@@ -90,3 +90,7 @@ select = [
 markers = [
     "examples",
 ]
+addopts = [
+  "--store-durations",  # pytest-split, update duration cache
```
lrandersson, line 94: Inspired by https://github.com/conda/conda/blob/a58f588f1853f51706d34e1197b94289fe759904/pyproject.toml#L161

examples/miniforge-mamba2/test_install.sh
```diff
@@ -41,24 +41,35 @@ if [ "$LIBMAMBAPY_VERSION" != "$EXPECTED_MAMBA_VERSION" ]; then
     exit 1
 fi
 
+# mamba occasionally fails to load libmamba.2.dylib on macOS CI right after
```
lrandersson, line 44: All the changes in this file is to see if it resolves the intermittent issue:
```
INFO:constructor.main:Successfully created '/private/var/folders/df/djsxfhc17x95674wsm_g8s980000gn/T/pytest-of-runner/pytest-1/test_example_miniforge_mamba2_0/installer/Miniforge3-mamba2-25.1.1-0-MacOSX-arm64.sh'.

dyld[31994]: Library not loaded: @rpath/libmamba.2.dylib
  Referenced from: <A07BA708-81EE-3434-9A5E-50BBA7FCC449> /private/var/folders/df/djsxfhc17x95674wsm_g8s980000gn/T/pytest-of-runner/pytest-1/test_example_miniforge_mamba2_0/install/batch/pkgs/mamba-2.0.8-h105ca85_2/bin/mamba
  Reason: tried: '/private/var/folders/df/djsxfhc17x95674wsm_g8s980000gn/T/pytest-of-runner/pytest-1/test_example_miniforge_mamba2_0/install/batch/pkgs/mamba-2.0.8-h105ca85_2/bin/../lib/libmamba.2.dylib' (no such file), '/private/var/folders/df/djsxfhc17x95674wsm_g8s980000gn/T/pytest-of-runner/pytest-1/test_example_miniforge_mamba2_0/install/batch/pkgs/mamba-2.0.8-h105ca85_2/bin/../lib/libmamba.2.dylib' (no such file), '/usr/local/lib/libmamba.2.dylib' (no such file), '/usr/lib/libmamba.2.dylib' (no such file, not in dyld cache)
/private/var/folders/df/djsxfhc17x95674wsm_g8s980000gn/T/pytest-of-runner/pytest-1/test_example_miniforge_mamba2_0/install/batch/pkgs/post_install.sh: line 48: 31994 Abort trap: 6           mamba info
ERROR: executing post_install.sh failed
```
When debugging it seems like `mamba --version` passes, but `mamba info` directly after fails. Therefore, perhaps this will help mitigate the flakiness of this test.

.github/workflows/main.yml
```diff
@@ -32,7 +32,7 @@ defaults:
     shell: bash -el {0}
 jobs:
   tests:
-    name: ${{ matrix.os }}, Python ${{ matrix.python-version }}, ${{ matrix.conda-standalone }}
+    name: ${{ matrix.os }}, Python ${{ matrix.python-version }}, ${{ matrix.conda-standalone }}${{ matrix.split-group && format(', group {0}/3', matrix.split-group) || '' }}
```
marcoesters: Since you are only deploying two groups for macOS, hard-coding `3` makes it looks like tests are missing.

.github/workflows/main.yml
```diff
@@ -52,34 +52,94 @@ jobs:
             python-version: "3.13"
             conda-standalone: conda-standalone-onedir
             check-docs-schema: true
-          # MACOS
+          # MACOS (split into 2 groups per Python version)
```
marcoesters: Why split macOS in two groups? Because of the queue? This matrix is hard to manage with every scenario hard-coded.

.github/workflows/main.yml
```diff
@@ -154,12 +214,17 @@ jobs:
...
-          CONSTRUCTOR_EXAMPLES_KEEP_ARTIFACTS: "${{ runner.temp }}/examples_artifacts"
           CONSTRUCTOR_SIGNTOOL_PATH: "C:/Program Files (x86)/Windows Kits/10/bin/10.0.26100.0/x86/signtool.exe"
           CONSTRUCTOR_VERBOSE: 0
         run: |
           rm -rf coverage.json
-          pytest -ra -vvv --cov=constructor --cov-branch tests/test_examples.py
+          # Use pytest-split for Windows runners with split-group, run all tests otherwise
+          if [[ -n "${{ matrix.split-group }}" ]]; then
+            pytest -ra -vvv --cov=constructor --cov-branch tests/test_examples.py \
+              --splits 3 --group ${{ matrix.split-group }}
```
marcoesters: We only split macOS in 2, so this would introduce a coverage gap.

.github/workflows/main.yml
lrandersson: Good catch (see below)

.github/workflows/main.yml
lrandersson: Originally I wasn't planning on modifying macOS, but then I did anyway for testing purposes, and then I did 2 groups instead of 3 because we always have to sit and wait for the macOS runner queue.
I've now split it into 3 groups: 592f6cf
I'm also fine just leaving macOS as is (no split at all) since the main bottle neck is Windows, but I still think doing _something_ about macOS would be good for us since sometimes the test duration reach almost 90 minutes.

marcoesters: I agree that we should split macOS. I don't mind it being split in two necessarily as long as we don't introduce coverage gaps. Runner availability is a real concern for sure.

lrandersson: Let me know if I should change it again or not. On the other hand with more runners they also finish faster so I guess there is a trade-off. If we ever feel like changing it again in the future it's done very easily.

marcoesters (approved):
Lets give this a try

PR #1314, opened 2026-08-06
Add build environment to info.json

lrandersson:
### Description
This PR adds a new key to the file `info.json` containing a list of packages that are installed in the environment used to build the installer.
Small sample image from the file `info.json` that is produced when I build the test example `outputs` using my local `conda` environment:
```
  "_build_environment_packages": [
    {
      "name": "zstandard",
      "version": "0.24.0",
      "build": "py313h3d778a8_0",
      "build_number": 0,
      "channel": "https://repo.anaconda.com/pkgs/main/linux-64",
      "subdir": "linux-64",
    ...
```

### Checklist - did you ...

### DIFF
--- a/constructor/main.py
+++ b/constructor/main.py
@@ -25,7 +25,7 @@ from textwrap import dedent
 from . import __version__
 from ._schema import InstallerTypes
 from .build_outputs import process_build_outputs
-from .conda_interface import SUPPORTED_PLATFORMS, cc_platform
+from .conda_interface import SUPPORTED_PLATFORMS, cc_platform, get_build_env_records
 from .conda_interface import VersionOrder as Version
 from .construct import SCHEMA_PATH, ns_platform
 from .construct import parse as construct_parse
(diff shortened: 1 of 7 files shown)


constructor/main.py
```diff
@@ -279,6 +279,8 @@ def main_build(
         exe_version = Version(exe_version)
     info["_conda_exe_type"] = exe_type
     info["_conda_exe_version"] = exe_version
+    # Packages installed in the environment running constructor.
+    info["_build_environment_packages"] = get_build_env_records()
```
marcoesters: Since it's only needed for the `info.json` output, maybe this should move to the output section. If we don't output that file, we don't need to gather all this data.

lrandersson: Thats a good point. See a86c4db.

lrandersson: Just to follow-up, I moved the function definition due to https://github.com/conda/constructor/issues/1319.

marcoesters (approved):

PR #1316, opened 2026-08-06
Fix deprecation warning since conda 26.3

### LINKED ISSUE #1315: Fix deprecated conda._vendor.distro import in `conda_interface.py
The file `constructor/conda_interface.py:` imports `distro` from `conda._vendor`, which is deprecated and slated for removal in conda 26.3 the existing `try/except ImportError` silently swallows the failure, leaving `distro = None`). This means `preconda.py`'s `linux_distribution()` info silently stops being collected once conda drops the vendored module, with no visible error.

This is probably resolved by adding the package `distro` and making necessary changes to account for the additional package.

lrandersson:
### Description
This PR fixes https://github.com/conda/constructor/issues/1315.

As of right now I only added `distro` to Linux to mimic the old behavior. However, let me know if we rather add it for all the platforms so I can remove the conditional imports which I'm personally not a big fan of.

### Checklist - did you ...

### DIFF
--- a/constructor/conda_interface.py
+++ b/constructor/conda_interface.py
@@ -102,15 +102,9 @@ if conda_interface_type == "conda":
     distro = None
     if sys.platform.startswith("linux"):
         try:
-            import distro
+            import distro  # noqa: F401  # re-exported for use in preconda.py
         except ImportError:
             pass
-
-        if distro is None:
-            try:
-                from conda._vendor import distro
-            except ImportError:
-                pass
 
     def get_repodata(url):
         if CONDA_MAJOR_MINOR >= (23, 5):
--- a/constructor/conda_interface.py
+++ b/constructor/conda_interface.py
@@ -102,9 +102,15 @@ if conda_interface_type == "conda":
     distro = None
     if sys.platform.startswith("linux"):
         try:
-            import distro  # noqa
+            import distro
         except ImportError:
             pass
+
+        if distro is None:
+            try:
+                from conda._vendor import distro
+            except ImportError:
+                pass
 
     def get_repodata(url):
         if CONDA_MAJOR_MINOR >= (23, 5):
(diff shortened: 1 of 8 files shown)


constructor/conda_interface.py
```diff
@@ -102,7 +102,7 @@
     distro = None
     if sys.platform.startswith("linux"):
         try:
-            from conda._vendor import distro  # noqa
+            import distro  # noqa
```
marcoesters: `conda` has been depending on `distro` for some time now. Did you verify that this applies to all supported version for `constructor`? If so, you can remove the `try` block and remove `distro` as a conditional dependency.

If not, I suggest

```
try:
    import distro
except ImportError:
    pass

if not distro:
    try:
        from conda._vendor import distro
    except ImportError:
        pass
```

lrandersson: Good point, I saw this https://github.com/conda/conda/commit/18084382ecd54ad878148066cbc18813ab365ad1.
But since the constructor spec allows for much older versions of conda it's probably better to with your suggestion, see 042b5f6.

marcoesters: Are the changes in `preconda.py` still valid for the vendored `distro`? I'm also okay with bumping the minimum `conda` version to `24.9`, which is almost two years old.

lrandersson: So it looks like `distro` became a dependency with `conda` since `conda` version `24.1`. Since `conda 4.x` is very old (and likely not supported with the Python versions we use today) I would suggest that we bump the `minimum` depenency of `conda` to `24.1`. We will still list the explicit `distro` dependency even though conda pulls it transitively, but we can simplify how to handle `distro` in `conda_interface.py` by dropping the fallback entirely.

How does this sound?

marcoesters: That works for me.

marcoesters (approved):

PR #1317, opened 2026-08-06
Ensure installer option is gated behind install scope

lrandersson:
### Description
I noticed the option to add to path is not gated behind the install scope, and it is not possible to hide an installer option for MSI installers thus this option is displayed even when installing for "All Users". 

### Checklist - did you ...

### DIFF
--- a/constructor/briefcase.py
+++ b/constructor/briefcase.py
@@ -308,7 +308,7 @@ def create_install_options_list(info: dict) -> list[dict]:
         options.append(
             {
                 "name": "initialize_conda",
-                "title": "Add installation to my PATH environment variable",
+                "title": "Add installation to my PATH environment variable (single-user installs only)",
                 "description": description,
                 "default": info.get("initialize_by_default", False),
             }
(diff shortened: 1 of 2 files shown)


marcoesters (commented):
This should be disclosed in the option description, too.

Even though I wonder if this is even necessary. [CVE-2022-26526](https://nvd.nist.gov/vuln/detail/CVE-2022-26526), which this restriction is based on, has two components: the directory is added to `PATH` *and* is world-writable. The latter is the actual attack vector, not adding to `PATH`.

We can add this to be consistent, but I don't think this is 100% necessary.

lrandersson:
> This should be disclosed in the option description, too.
> 
> Even though I wonder if this is even necessary. [CVE-2022-26526](https://nvd.nist.gov/vuln/detail/CVE-2022-26526), which this restriction is based on, has two components: the directory is added to `PATH` _and_ is world-writable. The latter is the actual attack vector, not adding to `PATH`.
> 
> We can add this to be consistent, but I don't think this is 100% necessary.

@marcoesters 
That's fair. I actually only created this PR because from internal testing the question came up "what happens if I do this?" and I just expect there will be some unexpected failure. This ticket has also been created recently to see if it can be handled in a better way https://github.com/beeware/briefcase/issues/2986.

marcoesters (approved):

PR #1323, opened 2026-08-11
Extend support to include win-arm64

lrandersson:
### Description
This PR adds source-level support for `win-arm64` (arch parsing, platform registration, etc.) such that `constructor` can build ARM64 installers when invoked with `--platform win-arm64`. Does not add `win-arm64` to CI — that's tracked separately.

New error check - how it looks like running an `.exe` built on my amd64 pc while running it on arm64 via parallels:
image

### Checklist - did you ...

### DIFF
--- a/constructor/imaging.py
+++ b/constructor/imaging.py
@@ -14,6 +14,8 @@ from os.path import dirname, join
 from pathlib import Path
 from random import randint
 
+from PIL import Image, ImageDraw, ImageFont
+
 from ._schema import InstallerTypes
 
 ttf_path = join(dirname(__file__), "ttf", "Vera.ttf")
@@ -35,8 +37,6 @@ header_size_msi = (493, 58)
 
 
 def new_background(size, color, bs=20, boxes=50):
-    from PIL import Image, ImageDraw
-
     im = Image.new("RGB", size, color=color)
     d = ImageDraw.Draw(im)
     for unused in range(boxes):
@@ -48,8 +48,6 @@ def new_background(size, color, bs=20, boxes=50):
 
 
 def add_text(im, xy, text, min_lines, line_height, font, color):
-    from PIL import ImageDraw
-
     x, y = xy
     d = ImageDraw.Draw(im)
     lines = text.splitlines()
@@ -63,8 +61,6 @@ def add_text(im, xy, text, min_lines, line_height, font, color):
 
 
 def mk_welcome_image(info):
-    from PIL import ImageFont
-
     font = ImageFont.truetype(BytesIO(ttf_bytes), 20)
     im = new_background(welcome_size, info["_color"])
     text = "\n".join([info["welcome_image_text"], info["version"]])
@@ -73,8 +69,6 @@ def mk_welcome_image(info):
 
 
 def mk_welcome_image_osx(info):
-    from PIL import Image, ImageFont
-
     font = ImageFont.truetype(BytesIO(ttf_bytes), 40)
     # Transparent background
     im = Image.new("RGBA", welcome_size_osx, color=(0, 0, 0, 0))
@@ -84,8 +78,6 @@ def mk_welcome_image_osx(info):
 
 
 def mk_header_image(info):
-    from PIL import Image, ImageFont
-
     font = ImageFont.truetype(BytesIO(ttf_bytes), 20)
     im = Image.new("RGB", header_size, color=white)
     text = info["header_image_text"]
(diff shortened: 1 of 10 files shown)


constructor/nsis/main.nsi.tmpl
```diff
@@ -657,8 +657,9 @@ Function .onInit
     # Select the correct registry to look at, depending
     # on whether it's a 32-bit or 64-bit installer
     SetRegView {{ BITS }}
-{%- if win64 %}
+{%- if win64 or win_arm64 %}
     # If we're a 64-bit installer, make sure it's 64-bit Windows
+    # Not verified yet on ARM64 Windows.
```
lrandersson: TODO: To test

constructor/nsis/main.nsi.tmpl
```diff
@@ -658,12 +658,23 @@ Function .onInit
     # on whether it's a 32-bit or 64-bit installer
     SetRegView {{ BITS }}
 {%- if win64 %}
-    # If we're a 64-bit installer, make sure it's 64-bit Windows
-    ${IfNot} ${RunningX64}
+    # Make sure we're not on 32-bit or native ARM64 Windows (the latter would
```
lrandersson, line 661: Image

How it looks like running an `.exe` built on my x64 PC while running it on arm64.

tests/test_construct.py
```diff
@@ -37,6 +38,10 @@
 """
 
 
+def test_supported_platforms_includes_win_arm64():
+    assert "win-arm64" in SUPPORTED_PLATFORMS
```
marcoesters: Aren't you covering that in `test_ns_platform`? If not, I think this should be a more general test.

constructor/imaging.py
```diff
@@ -37,6 +35,8 @@
 
 
 def new_background(size, color, bs=20, boxes=50):
+    from PIL import Image, ImageDraw
```
marcoesters: Is there a way we can do this test on a module level?

constructor/nsis/main.nsi.tmpl
```diff
@@ -658,12 +658,23 @@ Function .onInit
...
-            ${NAME} installer." \
+            "This installer is for the 64-bit (x86_64) version of ${NAME}$\n\
+            but your system is not x86_64. Please use the installer that$\n\
+            matches your system's architecture." \
+            /SD IDOK
+        Abort
+    ${EndIf}
+{%- elif win_arm64 %}
+    # Make sure we're actually on native ARM64 Windows.
+    ${IfNot} ${IsNativeARM64}
```
marcoesters, line 673: Can we have a similar check for the MSI installers?

marcoesters (commented):
Nice! Just a few details to check.

tests/test_construct.py
lrandersson: Good catch, removed 5fce452

constructor/imaging.py
lrandersson: Yeah so there are many ways and honestly I dont think any solution looks pretty, but I think the changes I made here 5fce452 now looks "the least cluttered".

constructor/nsis/main.nsi.tmpl
lrandersson: I did some investigation and couldnt find anything that exists today unfortunately. @freakboy3742 do you know if there is anything to check if an `.msi` installer is running on the wrong architecture? If not, let me know I'm happy to create a ticket.

marcoesters (approved):

constructor/nsis/main.nsi.tmpl
freakboy3742: WiX (and therefore Briefcase-generated) installers do some CPU detection by default - if you try to run an ARM64 installer on x86-64, you get an error:

Screenshot 2026-08-17 at 6 32 43 am

The inverse isn't true, however - ARM64 machines can run x86-64 installers (using CPU emulation mode). It might be possible to prevent this case as well, but it's not currently prevented by default.

lrandersson: I see, perhaps also due to emulation it currently makes sense not to prevent it by default but perhaps exposing an option in pyproject.toml to enable/disable it would do the job? If you think so, let me know and Im happy to open a new ticket.

freakboy3742: Exposing it as an option seems plausible as an option; my main concern would be whether there's actually demand for it. From what I've seen, it won't be a *trivial* addition (as in - it's not a simple "if" statement or pre-existing MSI config option - we'd need to build it from low-level primitives). If this is a case of "theoretical" interest rather than a use case, then I'd rather avoid the complexity. 

That said, I guess a ticket couldn't hurt from at least an information gathering capacity...

PR #1325, opened 2026-08-11
Switch from PrefixRecord to PackageRecord

lrandersson:
### Description
This PR improves the changes added in https://github.com/conda/constructor/pull/1314. It was noticed that `PrefixRecord` added much more metadata than necessary which bloated the `info.json` file. Comparing the two `info.json` (before/after this implementation) shows about a 70x reduction in file size when using `PackageRecord`.

### Checklist - did you ...

### DIFF
--- a/constructor/build_outputs.py
+++ b/constructor/build_outputs.py
@@ -15,7 +15,6 @@ from conda.base.constants import UNKNOWN_CHANNEL
 from conda.common.url import remove_auth, split_anaconda_token
 from conda.core.prefix_data import PrefixData, PrefixGraph
 from conda.exports import default_prefix
-from conda.models.records import PackageRecord
 
 from . import __version__
 from .conda_interface import VersionOrder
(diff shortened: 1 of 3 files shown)


marcoesters (commented):
This PR should be added to the news item.

Technically, this is a hack. `PackageRecord` and `PrefixRecord` describe two separate items (an archive file vs an installed package in an environment). Instead of coercing this into `PackageRecords`, could we just remove the entries we don't need in the dictionary representation/the data dump before we add it into `info.json`? That way, we stay correct while also removing unnecessary data.

lrandersson:
Good call @marcoesters, see e9d8bce. However regarding the news item, we only modified the existing implementation so I would say it is already covered by https://github.com/conda/constructor/blob/main/news/1314-build-environment-info, or do you disagree?

marcoesters:
> However regarding the news item, we only modified the existing implementation so I would say it is already covered by https://github.com/conda/constructor/blob/main/news/1314-build-environment-info, or do you disagree?

Correct. So, the new item should be modified to list both PR numbers, i.e., `(#1314, #1325)`.

marcoesters (approved):

PR #1327, opened 2026-08-12
Add installer hashes to info.json

Jrice1317:
### Description
Adds installer hashes to info.json so downstream tooling can use it as a single source of truth for installer checksums.

## Changes:
* Calculate requested installer hashes once and reuse them across build outputs.
* Always include SHA256 in info.json; include additional algorithms requested by hash.
* Refactor hash_files() to calculate multiple algorithms in one pass.
* Preserve existing checksum file output.
* Update schema, docs, and tests

## Test expectations:
* If only info.json is requested → expect SHA256 in _installer_hashes.
* If only MD5 is requested → expect MD5 and the .md5 checksum file.
* If both info.json and MD5 requested → expect SHA256 and MD5 to be calculated once and reused.
* If neither is requested → expect no installer hashes to be calculated.
* If multiple algorithms are requested → expect all hashes to be calculated in one pass.
* If info.json and checksum files are generated → expect their hash values to match.
* Existing SH installer MD5 behavior remains unchanged.

### Checklist - did you ...


constructor/data/construct.schema.json
```diff
@@ -432,7 +432,7 @@
     },
     "build_outputs": {
       "default": [],
-      "description": "Additional artifacts to be produced after building the installer. It expects either a list of strings or single-key dictionaries.\nAllowed strings / keys: `hash`, `info.json`, `licenses`, `lockfile`, `pkgs_list`.",
+      "description": "Additional artifacts to be produced after building the installer. It expects either a list of strings or single-key dictionaries.\nRequesting `info.json` adds an `_installer_hashes` property containing the SHA256 digest of the generated installer. Hash algorithms requested through a `hash` build output are also included in this property.\nFor example:\n```json\n\"_installer_hashes\": {\n  \"sha256\": \"...\",\n  \"md5\": \"...\"\n}\n```\nThe `hash` output continues to create separate checksum files.\nAllowed strings / keys: `hash`, `info.json`, `licenses`, `lockfile`, `pkgs_list`.",
```
lrandersson: I wonder if we should not mention this here. I do think the information per say is good but as far as I can see we dont have any place right now explaining the contents of `info.json` - because of this it seems out of place.

constructor/build_outputs.py
```diff
@@ -46,48 +46,73 @@ def _validate_output(output):
...
+    """Return hash algorithms required by the requested build outputs."""
+    algorithms = set()
+
     for output in info.get("build_outputs", ()):
         output = _validate_output(output)
         name, config = output.popitem()
+
+        if name == "info.json":
+            algorithms.add("sha256")
+        elif name == "hash":
```
lrandersson:
```suggestion
        if name == BuildOutputs.INFO_JSON:
            algorithms.add("sha256")
        elif name == BuildOutputs.HASH:
```
You might have to change some imports to be able to apply this suggestion.

constructor/utils.py
```diff
@@ -76,16 +76,45 @@ def replace(match):
...
+    dict[str, str]
+        A mapping of algorithm names to digest values.
+    """
+    if isinstance(algorithms, str):
+        algorithms = [algorithms]
+
+    algorithms = set(algorithms)
+    invalid = algorithms.difference(hashlib.algorithms_available)
+    if invalid:
+        raise ValueError(f"Invalid algorithm: {', '.join(sorted(invalid))}")
```
lrandersson: These needs to be singular/plural based on `len(invalid)`.

constructor/utils.py
```diff
@@ -76,16 +76,45 @@ def replace(match):
...
+
+    BUFFER_SIZE = 65536
```
lrandersson, line 104: Whats the rationale here for changing to `64KB` instead of `256KB` (262144 below)?

news/1327-installer-hashes-to-info-json
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Add installer hashes to `info.json` making it the single source of truth for installer checksums used in SBOM generation. (#1327)
```
lrandersson:
```suggestion
* Add installer hashes to `info.json` making it the single source of truth for installer checksums. (#1327)
```
I propose that we skip the last part since in theory users don't need to know the connection to SBOM generation.

constructor/data/construct.schema.json
marcoesters: I concur

Jrice1317: Should I remove the whole thing or just the hardcoded example? I thought this was a good place because we do list `info.json` as an option in the schema [here](https://github.com/conda/constructor/blob/main/constructor/_schema.py#L124). However, I could see how my addition could be a little too detailed.

constructor/utils.py
Jrice1317: I didn't change it. I only copied over what was already in `build_outputs.py` [here](https://github.com/conda/constructor/blob/main/constructor/build_outputs.py#L87). With that being said, if it should be changed to 262144 instead, then just let me know. 😃

constructor/build_outputs.py
```diff
@@ -46,48 +46,73 @@ def _validate_output(output):
...
+            algorithm = config.get("algorithm")
+            if isinstance(algorithm, str):
+                algorithms.add(algorithm)
+            elif algorithm:
+                algorithms.update(algorithm)
```
marcoesters, line 78: We need to validate that the hash algorithms are valid. It looks like that part of the code got removed.

constructor/build_outputs.py
```diff
@@ -46,48 +46,73 @@ def _validate_output(output):
...
-            outpath = Path(f"{installer}.{algo}")
-            with open(outpath, "w", newline="\n") as f:
-                f.write(f"{filehash.hexdigest()}  {installer.name}\n")
-            outpaths.append(str(outpath.absolute()))
+
+    for algo in algorithms:
+        outpath = Path(f"{installer}.{algo}")
+
+        with open(outpath, "w", newline="\n") as f:
+            f.write(f"{info['_installer_hashes'][algo]}  {installer.name}\n")
```
marcoesters: This should have a guard to make sure we don't have an uncaught exception here. That way, we can have a more meaningful error message.

constructor/data/construct.schema.json
marcoesters: I'd say remove the whole thing.

constructor/build_outputs.py
Jrice1317: It didn't get removed completely. It got moved [here](https://github.com/conda/constructor/pull/1327/changes/BASE..8b90819db09daf2fff1ba3a410baffe676000b18#diff-12957650a93c55573868d4e88403ef85d08ec75609d343fe46bd7cbd75624af6R100-R101).

marcoesters: Ah, yes, thanks for pointing that out! It looks like the test catching that `ValueError` got removed though, so we are missing some test coverage.

Jrice1317: The `ValueError` is in `utils.py` now.

marcoesters (approved):

lrandersson (approved):

PR #1331, opened 2026-08-20
Size estimate updates using construct.yaml options

### LINKED ISSUE #1330: Size estimate low, based on constructor.yaml options
### Checklist

### What happened?

Creating a new installer estimates the total size of the install lower than the actual size on disk.

Image

Image

### Additional Context

The install estimate is off because of selected `construct.yaml` options.

```yaml
keep_pkgs: true
condarc:
  allow_softlinks: false
  always_copy: true
```

This set of options ends up with more disk space needed because `always_copy: true` creates a copy.

amaloney:
### Description

This commit modifies `constructor/winexe.py` based on the following options found in `construct.yaml`

```yaml
keep_pkgs: true
condarc:
  allow_softlinks: false
  always_copy: true
```

These sets of options within `construct.yaml` creates copies of packages because hard links are not used. Not using hard links prevents users from copy/pasting environments or packages across drives, and getting errors due to broken links.

Resolves: #1330 

previous-size-estimate
actual-size
updated-size-estimate

### Checklist - did you ...

- [ ] ~Add / update necessary tests?~
- [ ] ~Add / update outdated documentation?~

### DIFF
--- a/constructor/winexe.py
+++ b/constructor/winexe.py
@@ -278,7 +278,21 @@ def make_nsi(
     variables["uninstall_with_conda_exe"] = bool(info.get("uninstall_with_conda_exe"))
     variables["needs_python_exe"] = info.get("_win_install_needs_python_exe", True)
 
-    approx_pkgs_size_kb = approx_size_kb(info, "pkgs")
+    # - `keep_pkgs: true` + `always_copy: true` = tarballs + 2 * extracted
+    # - `keep_pkgs: true` + `always_copy: false` = tarballs + extracted
+    # - `keep_pkgs: false` = extracted only
+    # `keep_pkgs: true`
+    if info.get("keep_pkgs", False):
+        # tarballs in pkgs/
+        size_tarballs = info.get("_approx_tarballs_size", 0)
+        # extracted tarballs in pkgs/<name>/
+        size_extracted = info.get("_approx_pkgs_size", 0)
+        # `always_copy: true` in condarc means installed files are copies, not hard links
+        if info.get("condarc", {}).get("always_copy", False):
+            size_extracted = 2 * size_extracted
+        approx_pkgs_size_kb = int((size_tarballs + size_extracted) / 1000)
+    else:
+        approx_pkgs_size_kb = approx_size_kb(info, "pkgs")
 
     # UPPERCASE variables are unescaped (and unquoted)
     variables["CONDA_LOG_ARG"] = (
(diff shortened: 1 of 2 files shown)


constructor/winexe.py
```diff
@@ -278,7 +278,21 @@ def make_nsi(
     variables["uninstall_with_conda_exe"] = bool(info.get("uninstall_with_conda_exe"))
     variables["needs_python_exe"] = info.get("_win_install_needs_python_exe", True)
 
-    approx_pkgs_size_kb = approx_size_kb(info, "pkgs")
+    # - `keep_pkgs: true` + `always_copy: true` = tarballs + 2 * extracted
+    # - `keep_pkgs: true` + `always_copy: false` = tarballs + extracted
+    # - `keep_pkgs: false` = extracted only
+    # `keep_pkgs: true`
+    if info.get("keep_pkgs", False):
```
marcoesters, line 285: I think we need to keep the archives in the size calculation no matter what the value of `keep_pkgs` is since these files are removed after package installation.

The purpose of the file size calculation is not just to inform the user, but to prevent running out of disk space. So, the value should be the amount of disk space the installer needs for the installation, not what's left in the end.

constructor/winexe.py
```diff
@@ -278,7 +278,21 @@ def make_nsi(
...
+        # tarballs in pkgs/
+        size_tarballs = info.get("_approx_tarballs_size", 0)
+        # extracted tarballs in pkgs/<name>/
+        size_extracted = info.get("_approx_pkgs_size", 0)
+        # `always_copy: true` in condarc means installed files are copies, not hard links
+        if info.get("condarc", {}).get("always_copy", False):
+            size_extracted = 2 * size_extracted
```
marcoesters, line 292: This is correct for an installer that only has a `base` environment, but not for installers with multiple environments.

news/1330-disk-size-estimate-updates
```diff
@@ -0,0 +1,19 @@
+### Enhancements
+
+* Adjusts the way that the estimated disk usage is calculated, based on more `construct.yaml` variables.
```
marcoesters:
```suggestion
* Adjusts the way that the estimated disk usage is calculated, based on more `construct.yaml` variables. (#1330 via #1331)
```

marcoesters (commented):
Thanks! This is an interesting problem and your code is a good start. I don't think we should take `keep_pkgs` into account though and the estimate is going to get complicated when extra environments are involved.

constructor/winexe.py
amaloney: understood, let me think harder about it. I can see that this logic worked for my current use-case, but there are others that I have not thought about

travishathaway:
@conda-bot check

PR #1335, opened 2026-08-24
Improve how ACLs are set to improve runtime performance

lrandersson:
### Description

The `Setting installation directory permissions...` step runs `icacls /T` over all of `$INSTDIR`, which costs ~170 µs per file (~25 s for 150k files) even though only a few files actually needed fixing. This PR scopes it to one call per environment against that environment's `Scripts` directory.

#### Investigation into `conda`
Only `Scripts` is affected because exactly one conda code path hard links from a shared template: `LinkPathAction.create_python_entry_point_windows_exe_action`, whose target is hard coded to `Scripts/{command}.exe`. That template is the launcher stub inside `conda-standalone`'s `_MEI*` directory, which is created with an explicit user-only ACE and no inheritance; since hard links share one security descriptor, every launcher inherits that broken ACL. All other candidate directories are ruled out in code — `condabin` and `Library\bin` only ever get fresh writes from `conda init`, `Library\usr\bin` never gets generated files at all, and `Scripts\conda.exe` is a `copy()`.

`test_allusers_exe` is our guard for a future regression if conda ever creates launchers elsewhere. Its guard is now extracted and also called from `test_example_extra_envs` to cover the per-environment loop, and it additionally asserts that `windows_python_entry_point_exe` paths exist and all live in a `Scripts` directory, this way the test cannot silently stop testing this.

### Checklist - did you ...

### DIFF
--- a/constructor/winexe.py
+++ b/constructor/winexe.py
@@ -144,46 +144,6 @@ def setup_envs_commands(info, dir_path):
     return environments
 
 
-FIX_LAUNCHER_ACLS_BAT = """\
-SET ERR=0
-SET ICACLS=%1
-SHIFT
-:loop
-IF "%~1"=="" GOTO :done
-IF EXIST "%~1" (
-    %ICACLS% "%~1" /inheritance:e /T /C /Q
-    IF ERRORLEVEL 1 (
-        ECHO Failed to enable inheritance for entry point launchers in "%~1"
-        SET ERR=1
-    )
-)
-SHIFT
-GOTO :loop
-:done
-EXIT /B %ERR%
-"""
-
-
-def write_fix_launcher_acls_bat(dir_path: str) -> None:
-    """Write a batch script that re-enables ACL inheritance on entry point launchers.
-
-    Entry point launchers are hard links to a stub exe with inheritance disabled (see
-    the .nsi template's "Setting installation directory permissions" install step),
-    and hard links share one ACL. This script takes icacls.exe's path as its first
-    argument, followed by one `Scripts` glob per environment, and reports which
-    glob(s), if any, it failed to fix.
-
-    This is a standalone file (its args, including $INSTDIR-based paths only known at
-    install time, are filled in by the .nsi template) rather than an inlined nsExec
-    one-liner, because cmd.exe's quote-stripping when invoked via `cmd /C "..."`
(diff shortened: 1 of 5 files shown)


tests/test_examples.py
```diff
@@ -904,6 +954,7 @@ def test_example_extra_envs(tmp_path, request, installer_type):
         if installer_type == InstallerTypes.MSI:
             _run_uninstaller_msi(installer, install_dir)
         else:
+            _check_permission_inheritance(base)
```
lrandersson, line 944: Added the ACL verification to `extra_envs` to ensure it's working as intended with additional environments as well (since we dont recurse the entire installation directory anymore).

constructor/winexe.py
```diff
@@ -144,6 +144,46 @@ def setup_envs_commands(info, dir_path):
...
+    )
+)
+SHIFT
+GOTO :loop
+:done
+EXIT /B %ERR%
+"""
+
+
+def write_fix_launcher_acls_bat(dir_path: str) -> None:
```
lrandersson: I tried to inline everything in the NSIS template but never got it to work with multiple environments.
To avoid making the NSIS template harder to read I thought the best is simply to combine the approach with Python.

tests/test_examples.py
```diff
@@ -822,6 +822,47 @@ def _is_micromamba(path) -> bool:
...
+
+    Also verify the installation actually contains entry point launchers, so this
+    check cannot silently stop testing anything if a future fixture change drops
+    them. If conda ever moves them outside `Scripts` (where the installer's fix is
+    scoped to), the tree-wide scan below will still catch the resulting broken
+    inheritance regardless of location.
+    """
+    prefixes = [install_dir, *install_dir.glob("envs/*")]
+    launchers = [launcher for prefix in prefixes for launcher in _entry_point_launchers(prefix)]
+    assert launchers, "Installation must contain entry point launchers"
```
marcoesters: I'm not sure if we must assert that. This is the source of a bug, but I'm not sure we need to assert that. As you say in the doc string, the full scan makes sure that inheritance is correct for all files.

lrandersson: I agree, it's unnecessary. Removed in 6c6ea15

marcoesters (approved):

PR #1341, opened 2026-08-31
Fix CWE-426 (Untrusted Search Path) in the main NSIS template

michalwyszynski93:
### Description

Mitigating CWE-426 vulnerability in the example NSIS template by using a fully qualified path instead of naked `cmd.exe`.

Many dev teams build up on this example for their custom installers. This propagates the vulnerability downstream to other projects.

### Checklist - did you ...

### DIFF
--- a/examples/custom_nsis_template/custom.nsi.tmpl
+++ b/examples/custom_nsis_template/custom.nsi.tmpl
@@ -675,7 +675,7 @@ Section "Uninstall"
     !insertmacro AbortRetryNSExecWaitLibNsisCmd "rmreg"
 
     DetailPrint "Removing files and folders..."
-    nsExec::Exec 'cmd.exe /D /C RMDIR /Q /S "$INSTDIR"'
+    nsExec::Exec '"$SYSDIR\cmd.exe" /D /C RMDIR /Q /S "$INSTDIR"'
 
     # In case the last command fails, run the slow method to remove leftover
     RMDir /r /REBOOTOK "$INSTDIR"
--- a/examples/custom_nsis_template/custom.nsi.tmpl
+++ b/examples/custom_nsis_template/custom.nsi.tmpl
@@ -676,7 +676,7 @@ Section "Uninstall"
     !insertmacro AbortRetryNSExecWaitLibNsisCmd "rmreg"
 
     DetailPrint "Removing files and folders..."
-    nsExec::Exec '"$SYSDIR\cmd.exe" /D /C RMDIR /Q /S "$INSTDIR"'
+    nsExec::Exec 'cmd.exe /D /C RMDIR /Q /S "$INSTDIR"'
 
     # In case the last command fails, run the slow method to remove leftover
     RMDir /r /REBOOTOK "$INSTDIR"
--- a/examples/custom_nsis_template/custom.nsi.tmpl
+++ b/examples/custom_nsis_template/custom.nsi.tmpl
@@ -1,4 +1,5 @@
 # Installer template file for creating a Windows installer using NSIS.
+# This example is not actively maintained, use the main template as a blueprint for custom templates instead.
 
 # Dependencies:
 #   NSIS >=3.08      conda install "nsis>=3.08"  (includes extra unicode plugins)
--- a/news/TEMPLATE
+++ b/news/TEMPLATE
@@ -4,7 +4,7 @@
 
 ### Bug fixes
 
-* <news item>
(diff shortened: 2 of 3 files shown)


travishathaway:
@conda-bot check

examples/custom_nsis_template/custom.nsi.tmpl
```diff
@@ -675,7 +675,7 @@ Section "Uninstall"
     !insertmacro AbortRetryNSExecWaitLibNsisCmd "rmreg"
 
     DetailPrint "Removing files and folders..."
-    nsExec::Exec 'cmd.exe /D /C RMDIR /Q /S "$INSTDIR"'
+    nsExec::Exec '"$SYSDIR\cmd.exe" /D /C RMDIR /Q /S "$INSTDIR"'
```
marcoesters:
```suggestion
    nsExec::Exec '"$CMD_EXE" /D /C RMDIR /Q /S "$INSTDIR"'
```

news/TEMPLATE
```diff
@@ -4,7 +4,7 @@
 
 ### Bug fixes
 
-* <news item>
+* Mitigated a CWE-426 untrusted search path vulnerability in the example `custom.nsi.tmpl` by using a fully qualified path to `$SYSDIR\cmd.exe`.
```
marcoesters:
```suggestion
* Mitigated a CWE-426 untrusted search path vulnerability in the example `custom.nsi.tmpl` by using a fully qualified path to `$CMD_EXE`. (#1341)
```

marcoesters (commented):
Good catch, thanks!

We do already try to figure out where `cmd.exe` might be located, but this instance has fallen through the cracks: https://github.com/conda/constructor/blob/29afc8098d6bfd741db32eb68eb7885f55464e6a/constructor/nsis/main.nsi.tmpl#L182-L197

Do you suggest we should rather error out than use `PATH` as the fallback?

michalwyszynski93:
> Good catch, thanks!
> 
> We do already try to figure out where `cmd.exe` might be located, but this instance has fallen through the cracks:
> 
> https://github.com/conda/constructor/blob/29afc8098d6bfd741db32eb68eb7885f55464e6a/constructor/nsis/main.nsi.tmpl#L182-L197
> 
> Do you suggest we should rather error out than use `PATH` as the fallback?

Hi @marcoesters, thanks for your feedback.

I'm more inclined to error it out than to fall back on unqualified path, otherwise the vulnerability is still there. Realistically speaking, the fallback is only needed when the host system has been tampered with badly, so it's OK to wash our hands?

Regarding your suggestion: as far as I understand, the `CMD_EXE` variable is declared/defined in the default template (constructor/nsis/main.nsi.tmpl) but not in the custom one (examples/custom_nsis_template/custom.nsi.tmpl), so it won't work out of the box and we'll need to add some equivalent of `FindWindowsBinaries` macro there?

marcoesters:
> I'm more inclined to error it out than to fall back on unqualified path, otherwise the vulnerability is still there. Realistically speaking, the fallback is only needed when the host system has been tampered with badly, so it's OK to wash our hands?

I think that sounds fair enough as long as we exhaust all possible locations for a legitimate `cmd.exe` binary (and `icacls.exe` while we're at it).

> Regarding your suggestion: as far as I understand, the `CMD_EXE` variable is declared/defined in the default template (constructor/nsis/main.nsi.tmpl) but not in the custom one (examples/custom_nsis_template/custom.nsi.tmpl), so it won't work out of the box and we'll need to add some equivalent of `FindWindowsBinaries` macro there?

My apologies, I completely missed the file name because the main template has the same line: https://github.com/conda/constructor/blob/29afc8098d6bfd741db32eb68eb7885f55464e6a/constructor/nsis/main.nsi.tmpl#L1966-L1967

To answer your question: yes, you would have to add the macro. I want to emphasize though that this example was primarily designed for integration testing and have not been kept up-to-date unless there was a bug in the template that breaks the NSIS compilation. The main template is the better blueprint for custom installers. My suggestion is to:

* Fix CWE-426 in the main template.
* Add a comment to the top of the example template to use the main template as a blueprint for custom templates, not the example template.
* Optional: use the macro to the example template.

marcoesters:
The failing Windows test is likely not due to this change - they tend to be a little flaky.

michalwyszynski93:
Took the liberty to rename the PR to reflect the changes accurately

marcoesters (approved):
Thanks! I will merge this once the CI runs are finished.

